PR #7, opened 2023-12-22
Preparing code for pypi

jdgratz10:
### Summary

This PR is a stepping stone on the way to preparing Aviary for packaging on pypi. The changes here are both necessary in order to ensure proper installation on pypi. Changes are:
- in setup.py the package name is changed to match what the package name with be on pypi, and a long description that loads the README.md file is added.
- In spec tests a test skip is added if the `.json` files do not exist. The reason for this is that on pypi the user will not be able to build these files and the tests will fail.

### Related Issues

- Resolves None

None

### New Dependencies

None

### DIFF
--- a/aviary/mission/gasp_based/ode/constraints/test/test_flight_constraints.py
+++ b/aviary/mission/gasp_based/ode/constraints/test/test_flight_constraints.py
@@ -50,7 +50,7 @@ class FlightConstraintTestCase(unittest.TestCase):
         partial_data = self.prob.check_partials(out_stream=None, method="cs")
         assert_check_partials(partial_data, atol=3e-11, rtol=1e-12)
 
-    @unittest.skipIf(not os.path.isfile(os.path.join('climb_specs/constraints.json', XDSM_PATH)), "`climb_specs/constraints.json` does not exist")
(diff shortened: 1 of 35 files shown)


aviary/mission/gasp_based/ode/constraints/test/test_flight_constraints.py
```diff
@@ -49,9 +50,9 @@ def test_case1(self):
         partial_data = self.prob.check_partials(out_stream=None, method="cs")
         assert_check_partials(partial_data, atol=3e-11, rtol=1e-12)
 
+    @unittest.skipIf(not os.path.isfile(os.path.join('climb_specs/constraints.json', XDSM_PATH)), "`climb_specs/constraints.json` does not exist")
```
crecine, line 53: I think the order of this needs to be flipped

crecine (approved):
approved, just fix the one path

aviary/mission/gasp_based/ode/constraints/test/test_flight_constraints.py
crecine, line 53:
```suggestion
    @unittest.skipIf(not os.path.isfile(os.path.join(XDSM_PATH, 'climb_specs/constraints.json')), "`climb_specs/constraints.json` does not exist")
```

PR #16, opened 2023-12-27
Fixed issues with conda environments not building successfully

crecine:
### Summary

The default environment name is now read from the yml file if it is not entered as an argument
pyoptsparse isn't moved from pip to conda dependencies (prevents failure)
moves all pip dependencies to a separate requirements.txt that is used after the conda environment is built
Modified workflow file to name environment based on test case, instead of a generic name

### Related Issues

- Resolves #

None

### New Dependencies

None


.github/install_env_from_github.sh
```diff
@@ -60,25 +61,45 @@ fi
...
 
 # Create a new conda environment from the modified YAML file
 $pkg_manager env create -f $output_yaml -n $env_name
 
-source activate base
+# Attempting to resolve "CommandNotFoundError: Your shell has not been properly configured to use 'conda activate'."
+# tmp=${SHELL##*/}
+# MYSHELL=${tmp##*\\}
+# $pkg_manager init $MYSHELL
+# source activate base
```
crecine, line 93: this seems to be unnecessary with the other fixes.

jkirk5 (approved):

johnjasa (approved):
I made a minor change to get `conda activate` to work correctly on my machine

PR #20, opened 2023-12-28
Limit Python version temporarily

crecine:
Changes to unittest are incompatible with testflo
The addSkip method in testflo causes test to fail in Python 3.12.1

testflo 1.4.14 is incompatible with Python 3.12.1


crecine:
unnecessary now that testflo has been updated

PR #27, opened 2024-01-03
Properly package Aviary and add a workflow for testing non-dev pip install

johnjasa:
### Summary

Some changes needed and helpful for pypi upload.

- added to the `setup.py` package setup so Aviary works as a non-dev pip install
- broke out the docs creation to a different workflow than the tests
- added a workflow for Aviary that just does `pip install .` instead of `pip install -e .[all]` to ensure all tests pass

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):

jdgratz10 (approved):

.github/workflows/test_workflow_no_dev_install.yml
```diff
@@ -0,0 +1,90 @@
...
+          path: ${{ matrix.NAME }}_environment.yml
+          retention-days: 5
+
+      - name: Run tests
+        shell: bash -l {0}
+        run: |
+          echo "============================================================="
+          echo "Run Tests"
+          echo "============================================================="
+          testflo . -n 1 --show_skipped
```
swryan, line 90: Note that it appears you are running the tests in the root directory of the repo, which means issues like missing `package_data` maynot be caught (since those files can be found relative to the working directory).

I would suggest changing to some neutral working directory (e.g. $HOME) for testing.

johnjasa: Great point, thanks Steve!

PR #36, opened 2024-01-05
Split Aircraft.Design.RESERVES to two variables

xjjiang:
### Summary

- Aviary variable Aircraft.Design.RESERVES is split into two variables in order to be able to set units for fuel reserves:
  - Aircraft.Design.RESERVES (with units: lbm, range: > 10)
  - Aircraft.Design.RESERVES_FRACTION (with units: unitless, range: [-1.0, 0])
-There is an additional new variable Aircraft.Design.RESERVES_OPTION so that users can choose to use either Aircraft.Design.RESERVES (option = 1) or Aircraft.Design.RESERVES_FRACTION (option = 2).
- All three variables are optional.
- When initial_guess['reserves'] is given, it will overwrite the data in .csv model file.
- The description is written in docs/getting_started/input_csv_phase_info.md.
- In GASP, the historic name is INGASP.FRESF. It has another range [0.0, 1.0] for time fraction where 1.0 for 45 minutes. This is not implemented in Aviary and should be.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,45 @@ def _add_gasp_landing_systems(self):
...
-                    reserve_fuel={"val": reserves_val, "units": "lbm"}
-                ),
-                promotes_outputs=[("reserve_fuel", reserves_name)],
-            )
+        reserves_val = self.aviary_inputs.get_val(Aircraft.Design.RESERVES, units='lbm')
+        reserves_fac = self.aviary_inputs.get_val(
+            Aircraft.Design.RESERVES_FRACTION, units='unitless')
+        reserves_opt = self.aviary_inputs.get_val(
+            Aircraft.Design.RESERVES_OPTION, units='unitless')
+        if reserves_opt == 1:
```
jdgratz10, line 2444: It seems like it might be more helpful here to have the reserves option be an enum or a string-valued variable instead of an integer, that would probably be more clear for the users.

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,45 @@ def _add_gasp_landing_systems(self):
...
+                self.model.add_subsystem(
+                    "reserves_calc",
+                    om.ExecComp(
+                        f"reserve_fuel = {reserves_val}",
+                        reserve_fuel={"val": reserves_val, "units": "lbm"}
+                    ),
+                    promotes_outputs=[("reserve_fuel", reserves_name)],
+                )
+            else:
+                raise ValueError('"aircraft:design:reserves" is not valid below 10.')
```
jdgratz10, line 2455: Just to be safe in case the variable name changes, I would use the actual variable name of Aircraft.Design.RESERVES here in the message instead of typing out its string name

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,45 @@ def _add_gasp_landing_systems(self):
...
+                    ),
+                    promotes_inputs=[
+                        ("takeoff_mass", Mission.Summary.GROSS_MASS),
+                        ("final_mass", Mission.Landing.TOUCHDOWN_MASS),
+                    ],
+                    promotes_outputs=[("reserve_fuel", reserves_name)],
+                )
+            elif reserves_fac > 0 and reserves_fac <= 1:
+                raise ValueError(
+                    '"aircraft:design:reserves_fraction" between 0 and 1 is not implemented.')
```
jdgratz10, line 2475: Just to be safe in case the variable name changes, I would use the actual variable name of Aircraft.Design.RESERVES_FRACTION here in the message instead of typing out its string name

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,45 @@ def _add_gasp_landing_systems(self):
...
+            else:
+                raise ValueError(
+                    '"aircraft:design:reserves_fraction" is not valid if less than -1 or larger than 1.')
```
jdgratz10, line 2478: Just to be safe in case the variable name changes, I would use the actual variable name of Aircraft.Design.RESERVES_FRACTION here in the message instead of typing out its string name

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,45 @@ def _add_gasp_landing_systems(self):
...
         else:
-            raise ValueError('"aircraft:design:reserves" is not valid between 0 and 10.')
+            raise ValueError('"aircraft:design:reserves_option" must be 1 or 2.')
```
jdgratz10, line 2480: Love all the detailed error checking :)

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.dat
```diff
@@ -195,6 +195,8 @@ aircraft:engine:data_file=models/engines/turbofan_23k_1.deck
  ROSCAB=9999.,	! cabin rate of sink during descent (500.) fpm
  RSMX=900.,	! maximum allowable rate of sink during landing approach ft per min (1000)
  FRESF=4998,   ! Reserve Fuel fraction of mission fuel (minus sign for code trigger)
+ FRESF2=0,   ! Reserve fuel fraction of mission fuel (negative sign trigger)
+ FRESF3=1,  ! Use Reserve Fuel
```
jdgratz10, line 199: Why the addition of these two lines?

aviary/models/small_single_aisle/small_single_aisle_GwGm.dat
```diff
@@ -205,7 +205,9 @@
  OFEM=.78,	! off design specified mission Mach number
  ROSCAB=9999.,	! cabin rate of sink during descent (500.) fpm
  RSMX=900.,	! maximum allowable rate of sink during landing approach ft per min (1000)
- FRESF=-.125,     ! Reserve Fuel fraction of mission fuel (minus sign for code trigger)
+ FRESF=0,   ! Reserve Fuel
+ FRESF2=-.125,     ! Reserve Fuel fraction of mission fuel (minus sign for code trigger)
+ FRESF3=2,  ! Use Reserve Fuel
```
jdgratz10, line 210: I'm not up on my GASP so I'm a little confused, why this change?

aviary/models/test_aircraft/converter_configuration_test_data_GwGm.dat
```diff
@@ -209,7 +209,9 @@ aircraft:engine:data_file=models/engines/turbofan_23k_1.deck
  OFEM=.78,	! off design specified mission Mach number
  ROSCAB=9999.,	! cabin rate of sink during descent (500.) fpm
  RSMX=900.,	! maximum allowable rate of sink during landing approach ft per min (1000)
- FRESF=-0.15,   ! Reserve fuel fraction of mission fuel (negative sign trigger)
+ FRESF=0,   ! Reserve Fuel
+ FRESF2=-0.15,   ! Reserve fuel fraction of mission fuel (negative sign trigger)
+ FRESF3=2,  ! Use Reserve Fuel fraction
```
jdgratz10, line 214: Same question about why the change

aviary/utils/legacy_code_data/gasp_default_values.dat
```diff
@@ -83,7 +83,9 @@ EYEW=1.5,deg
 FLAPN=1,unitless
 FN_REF=28690,lbf
 FPYL=0.7,unitless
-FRESF=4998,unitless
+FRESF=4998,lbm
+FRESF2=-0.1,unitless
+FRESF3=1,unitless
```
jdgratz10, line 88: Same question

aviary/utils/process_input_decks.py
```diff
@@ -45,7 +45,9 @@ def create_vehicle(vehicle_deck=''):
     aircraft_values.set_val('mass_defect', val=10000, units='lbm')
     aircraft_values.set_val('problem_type', val=ProblemType.SIZING)
     aircraft_values.set_val(Aircraft.Electrical.HAS_HYBRID_SYSTEM, val=False)
-    aircraft_values.set_val(Aircraft.Design.RESERVES, val=4998)
+    aircraft_values.set_val(Aircraft.Design.RESERVES, val=4998, units='lbm')
+    aircraft_values.set_val(Aircraft.Design.RESERVES_FRACTION, val=0, units='unitless')
+    aircraft_values.set_val(Aircraft.Design.RESERVES_OPTION, val=0, units='unitless')
```
jdgratz10, line 50: Why is the reserves option being set to an invalid value?

aviary/utils/process_input_decks.py
```diff
@@ -164,10 +166,56 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
 def initial_guessing(aircraft_values: AviaryValues()):
     problem_type = aircraft_values.get_val('problem_type')
+    reserves_option = aircraft_values.get_val(Aircraft.Design.RESERVES_OPTION)
     reserves = aircraft_values.get_val(
-        Aircraft.Design.RESERVES) if initial_guesses['reserves'] == 0 else initial_guesses['reserves']
+        Aircraft.Design.RESERVES, units='lbm') if initial_guesses['reserves'] == 0 else initial_guesses['reserves']
+    if reserves_option == 1:
+        aircraft_values.set_val(Aircraft.Design.RESERVES, reserves, units='lbm')
+        if reserves < 10:
+            raise ValueError('"aircraft:design:reserves" must be greater than 10 lbm.')
```
jdgratz10, line 175: Just to be safe in case the variable name changes, I would use the actual variable name of Aircraft.Design.RESERVES here in the message instead of typing out its string name

aviary/utils/process_input_decks.py
```diff
@@ -164,10 +166,56 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
+    elif reserves_option == 2:
+        if initial_guesses['reserves'] == 0:
+            reserves = aircraft_values.get_val(
+                Aircraft.Design.RESERVES_FRACTION, units='unitless')
+        else:
+            reserves = initial_guesses['reserves']
+            aircraft_values.set_val(Aircraft.Design.RESERVES_FRACTION,
+                                    reserves, units='unitless')
+        if reserves < -1 or reserves > 0:
+            raise ValueError('"aircraft:design:reserves_fraction" must be in [-1, 0].')
```
jdgratz10, line 185: Just to be safe in case the variable name changes, I would use the actual variable name of Aircraft.Design.RESERVES_FRACTION here in the message instead of typing out its string name

aviary/variable_info/variable_meta_data.py
```diff
@@ -1204,9 +1204,35 @@
...
     option=True,
+    units="lbm",
+    desc='required fuel reserves: directly in lbm',
+    default_value=0,
+)
+
+add_meta_data(
+    Aircraft.Design.RESERVES_FRACTION,
+    meta_data=_MetaData,
+    historical_name={"GASP": 'INGASP.FRESF2',  # the real name is INGASP.FRESF
```
jdgratz10, line 1215: If the value in the historical name is not the real value, how will that affect our Fortran-to-Aviary converter? @jkirk5 ?

aviary/variable_info/variable_meta_data.py
```diff
@@ -1204,9 +1204,35 @@
...
+    option=True,
+    units="unitless",
+    desc='required fuel reserves: given as a proportion of mission fuel',
+    default_value=0,
+)
+
+add_meta_data(
+    Aircraft.Design.RESERVES_OPTION,
+    meta_data=_MetaData,
+    historical_name={"GASP": 'INGASP.FRESF3',  # the real name is INGASP.FRESF
```
jdgratz10, line 1228: Same question as above

aviary/variable_info/variable_meta_data.py
```diff
@@ -1204,9 +1204,35 @@
...
+                     "FLOPS": None,
+                     "LEAPS1": None
+                     },
+    option=True,
+    types=int,
```
jdgratz10, line 1233: Again, I would like to see this be an enum or string instead of int

jdgratz10 (changes requested):
Good PR! I had a couple small change requests, but in general this looks solid.

aviary/interface/methods_for_level2.py
xjjiang: Agree. I will do.

aviary/interface/methods_for_level2.py
xjjiang: True. Will do.

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
```diff
@@ -14,7 +14,9 @@ aircraft:design:drag_increment,0.00175,unitless
 aircraft:design:equipment_mass_coefficients,928,0.0736,0.112,0.14,1959,1.65,551,11192,5,3,50,7.6,12,unitless
 aircraft:design:max_structural_speed,402.5,mi/h
 aircraft:design:part25_structural_category,3,unitless
-aircraft:design:reserves,4998,unitless
+aircraft:design:reserves,4998,lbm
+aircraft:design:reserves_fraction,0,unitless
+aircraft:design:reserves_option,1,unitless
 aircraft:design:static_margin,0.03,unitless
```
Kenneth-T-Moore, line 20: Thinking about the API here. There are 3 variables to specify now - "reserves", "reserves_fraction", and "reserves_option". Depending on how they set the option, one will be used and the other won't be used. I think this might cause some confusion.

What if we just had 2 variables to set?

reserves
reserves_fraction

The user can set one or the other. If they set both, then we raise an error in the pre-processor. I think this would be concise and unambiguous. We can process the logic based on which one they set.

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.dat
xjjiang: If I don't add these two lines, test_Fortran_to_Aviary fails.

aviary/models/small_single_aisle/small_single_aisle_GwGm.dat
xjjiang: In GASP, FRESF is used for actual total fuel or fraction of total fuel depending on its value. Now, we want FRESF to be read into the total fuel (Aircraft.Design.RESERVES), FRESF2 for fraction of fuel (Aircraft.Design.RESERVES_FRACTION). The third one (FRESF3) is to be read into Aircraft.Design.RESERVES_OPTION.

aviary/utils/process_input_decks.py
xjjiang: Aviary allows users to set the fuel in initial_guesses['reserves'] which overwrites the values in Aircraft.Design.RESERVES. So, I allow Aircraft.Design.RESERVES_OPTION to be 0 to indicate that we allow this behavior.

aviary/utils/process_input_decks.py
xjjiang: Done.

aviary/variable_info/variable_meta_data.py
xjjiang: My understanding is that historical_name creates a link between Aviary variable name and the one in gasp_default_values.dat. In the case of fuel reserves, 'INGASP.FRESF' is mapped to 'Aircraft.Design.RESERVES', 'INGASP.FRESF2' is mapped to 'Aircraft.Design.RESERVES_FRACTION'. I have to create 'INGASP.FRESF2' when I split 'Aircraft.Design.RESERVES'.

aviary/variable_info/variable_meta_data.py
xjjiang: Done.

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
xjjiang: I thought about this approach but decided not to go this way. There are two reasons:

1. GASP actually have three cases. In addition to the above two, there is another fraction for time. It maps (0, 1) to (0, 45 minutes). Jeff Bowles told me that it was important and shouldn't be missed. I think it is for reserve missions and we have not implemented. Your approach will be complicate when we add the third case.
2. I was thinking that users may want to set both  'Aircraft.Design.RESERVES' and  'Aircraft.Design.RESERVES_FRACTION' to some reasonable values and use  'Aircraft.Design.RESERVES_OPTION' to switch back and force.

Kenneth-T-Moore: That makes sense. I guess when that case is added, we would have another value for reserves_option.

aviary/interface/methods_for_level2.py
xjjiang: Done

aviary/interface/methods_for_level2.py
xjjiang: Done

aviary/variable_info/variable_meta_data.py
jkirk5: No, fortran_to_aviary directly checks variable_metadata for name matching, so this won't work. The defaults files aren't used for conversion in the same way.

Because we have split this variable in Aviary, we need to add logic to fortran_to_aviary to handle getting the legacy value to where it goes. Check out line 479 on Fortran_to_Aviary.py for an example of how I handled this for the mass scaling factors in FLOPS. Have the original historical name map to reserves_fraction, then check if the value is actually total mass - if so, update RESERVES and set RESERVES_FRACTION to 1.0.

Never "invent" a historical name, this just adds confusion. Leave historical_name for RESERVES empty (or None).

aviary/variable_info/variable_meta_data.py
jkirk5: This flag shouldn't be necessary - check out the FLOPS mass scalars for similar split variables that don't require a user-set flag. Also, please remove fake historical name, set as None

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
jkirk5: The third case definitely should not be set in Aviary using a (0,1) -> (0,45 min) variable, and should instead be handled the way we currently set up the design mission.

I disagree that Ken's suggestion makes things harder for the future, I have more thoughts in my overall review comments.

jkirk5 (changes requested):
These changes make sense for a literal translation of what GASP did, but we have a chance to rework things and make them so much better!

The three options users could use to specify reserve fuel (a fixed amount, % of design mission fuel, and fuel burned in a reserve mission) are all independent and can all be used at the same time - FLOPS actually allows this. It is very reasonable for a user to want to use all three to calculate total reserve fuel.

This means we should have the reserve fraction (from 0-1) default to zero and always get calculated, and fixed reserve fuel should likewise default to zero. This allows for so much more flexibility to the user while not loosing any capabilities.

aviary/utils/process_input_decks.py
jkirk5: Can you update the names to make it clear they refer to fuel? Such as FIXED_RESERVE_FUEL and RESERVE_FUEL_FRACTION

aviary/utils/process_input_decks.py
jkirk5: Also, can you update the names to make it clear they refer to fuel? Such as FIXED_RESERVE_FUEL and RESERVE_FUEL_FRACTION

aviary/variable_info/enums.py
```diff
@@ -181,3 +181,27 @@ class Flap_Type(Enum):
     """
     Double-slotted Fowler flaps
     """
+
+
+@unique
+class Reserves_Type(Enum):
```
jkirk5, line 187: This flag shouldn't be necessary - check out the FLOPS mass scalars for an example of how we've split variables in a way that doesn't require a user-set flag.

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,46 @@ def _add_gasp_landing_systems(self):
...
-            self.model.add_subsystem(
-                "reserves_calc",
-                om.ExecComp(
-                    f"reserve_fuel = {reserves_val}",
-                    reserve_fuel={"val": reserves_val, "units": "lbm"}
-                ),
-                promotes_outputs=[("reserve_fuel", reserves_name)],
-            )
+        reserves_val = self.aviary_inputs.get_val(Aircraft.Design.RESERVES, units='lbm')
+        reserves_fac = self.aviary_inputs.get_val(
```
crecine, line 2440: this should be "frac" or "fraction", "fac" sounds like it is a factor that will be applied to the reserves.

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,46 @@ def _add_gasp_landing_systems(self):
...
+            Aircraft.Design.RESERVES_FRACTION, units='unitless')
+        reserves_opt = self.aviary_inputs.get_val(
+            Aircraft.Design.RESERVES_OPTION, units='unitless')
+        if reserves_opt == Reserves_Type.Set_Direct:
```
crecine, line 2444: for Enums, `if reserves_opt is Reserves_Type.Set_Direct:` is preferred

aviary/interface/methods_for_level2.py
```diff
@@ -2436,31 +2436,46 @@ def _add_gasp_landing_systems(self):
...
+                        ("final_mass", Mission.Landing.TOUCHDOWN_MASS),
+                    ],
+                    promotes_outputs=[("reserve_fuel", reserves_name)],
+                )
+            elif reserves_fac > 0 and reserves_fac <= 1:
+                raise ValueError(
+                    f'"{Aircraft.Design.RESERVES_FRACTION}" between 0 and 1 is not implemented.')
+            else:
+                raise ValueError(
+                    f'"{Aircraft.Design.RESERVES_FRACTION}" is not valid if less than -1 or larger than 1.')
```
crecine, line 2478: although it doesn't necessarily make sense, a value less than -1 is allowed in GASP, and probably should be supported here

aviary/utils/process_input_decks.py
```diff
@@ -164,10 +167,58 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
 
 def initial_guessing(aircraft_values: AviaryValues()):
     problem_type = aircraft_values.get_val('problem_type')
+    reserves_option = aircraft_values.get_val(Aircraft.Design.RESERVES_OPTION)
     reserves = aircraft_values.get_val(
-        Aircraft.Design.RESERVES) if initial_guesses['reserves'] == 0 else initial_guesses['reserves']
+        Aircraft.Design.RESERVES, units='lbm') if initial_guesses['reserves'] == 0 else initial_guesses['reserves']
```
crecine, line 172: the initial guessing wasn't intended to be used to set/override the actual value used in the reserve calculations, it was just intended to provide reasonable guesses for masses along the trajectory to improve the convergence without requiring users to set every point. If we stick with that, we don't need all this detailed evaluation of the options and error checking, because all that will be handled when FRESF is read in.

crecine (changes requested):
I think it would be better to continue to have a single variable (FRESF) in the Fortran decks and add a simple check to the Fortran to Aviary converter that splits into reserves and reserves_fraction (something simple like the current reserve fuel check). This would allow us to reduce the number of places we need to do the error checking to a single file. We can keep reserves_option, but have that be set internally by the Fortran to Aviary converter (instead of by the user through an artificial legacy variable) which would allow us to simplify the reserve fuel component as well.
https://github.com/OpenMDAO/om-Aviary/blob/f9bcfffa0ca32bc861ce104e96250c93871e2c8d/aviary/interface/methods_for_level2.py#L2438-L2466

aviary/interface/methods_for_level2.py
xjjiang: Now, I removed `Aircraft.Design.RESERVES_OPTION`.

aviary/utils/process_input_decks.py
xjjiang: Removed the check.

aviary/utils/process_input_decks.py
xjjiang: - Now, reserves fraction goes from [0, 1].
- RESERVES is renamed to FIXED_RESERVE_FUEL.

aviary/variable_info/variable_meta_data.py
xjjiang: - I removed "invented" historical names.
- I use `Fortran_to_Aviary` to check and set `Aircraft.Design.FIXED_RESERVES_FUEL` and `Aircraft.Design.RESERVES_FRACTION`.

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
xjjiang: Now, I excluded the potential implementation of the third case here.

aviary/variable_info/enums.py
xjjiang: This class is removed.

aviary/interface/methods_for_level2.py
xjjiang: renamed "frac" to "frac".

aviary/interface/methods_for_level2.py
xjjiang: do you mean that GASP allow a fraction to be greater than 1?

aviary/utils/process_input_decks.py
xjjiang: Now, I keep the only GASP variable FRESF and let Fortran_to_Aviary to process the GASP .dat file.

xjjiang:
- I removed `Aircraft.Design.RESERVES_OPTION`.
- `Aircraft.Design.RESERVES` is renamed to `Aircraft.Design.FIXED_RESERVES_FUEL`.
- The description is in `docs/getting_started/input_csv_phase_info.md`.
- I removed `Aircraft.Design.IJEFF` because it is not used.

xjjiang:
After discussing with Carl, I made additional changes:
1. `Aircraft.Design.RESERVES` is renamed to `Aircraft.Design.FIXED_RESERVES_FUEL`.
2. In process_input_decks.py:initial_guessing(), allow fuel_fraction to be greater than 1. Do not update `Aircraft.Design.FIXED_RESERVES_FUEL` and `Aircraft.Design.RESERVES_FRACTION`
3. In `methods_for_level2.py:_add_fuel_reserve_component()`,  allow fuel_fraction to be greater than 1.
4. The behavior is described in `variable_meta_data.py` and `input_csv_phase_info.md`.

aviary/variable_info/variable_meta_data.py
```diff
@@ -1196,47 +1197,22 @@
...
-                     "FLOPS": None,
-                     "LEAPS1": None
-                     },
-    option=True,
-    types=Reserves_Type,
-    units="unitless",
-    desc=f'if {Reserves_Type.Set_Direct}, use Aircraft.Design.RESERVES; \
-           if {Reserves_Type.Set_Fraction}, use Aircraft.Design.RESERVES_FRACTION',
-    default_value=Reserves_Type.Set_None,
-)
```
crecine: Leave this in, but change the historical_name for GASP to `None`, the description to `"A flag that is set internally to switch between using a fixed reserve fuel mass or a fraction of mission fuel."` and default_value to `Set_Direct`

aviary/interface/methods_for_level2.py
```diff
@@ -2436,46 +2436,31 @@ def _add_gasp_landing_systems(self):
...
+                    takeoff_mass={"units": "lbm"},
+                    final_mass={"units": "lbm"},
+                    reserve_fuel={"units": "lbm"}
+                ),
+                promotes_inputs=[
+                    ("takeoff_mass", Mission.Summary.GROSS_MASS),
+                    ("final_mass", Mission.Landing.TOUCHDOWN_MASS),
+                ],
+                promotes_outputs=[("reserve_fuel", reserves_name)],
+            )
```
crecine, line 2466:
```suggestion
	reserves_opt =  self.aviary_inputs.get_val(Aircraft.Design.RESERVES_OPTION)
        if reserves_opt is Reserves_Type.Set_Direct:
            self.model.add_subsystem(
                "reserves_calc",
                om.ExecComp(
                    f"reserve_fuel = {reserves_val}",
                    reserve_fuel={"val": reserves_val, "units": "lbm"}
                ),
                promotes_outputs=[("reserve_fuel", reserves_name)],
            )
        elif reserves_opt is Reserves_Type.Set_Fraction:
            self.model.add_subsystem(
                "reserves_calc",
                om.ExecComp(
                    f"reserve_fuel = {reserves_frac}*(takeoff_mass - final_mass)",
                    takeoff_mass={"units": "lbm"},
                    final_mass={"units": "lbm"},
                    reserve_fuel={"units": "lbm"}
                ),
                promotes_inputs=[
                    ("takeoff_mass", Mission.Summary.GROSS_MASS),
                    ("final_mass", Mission.Landing.TOUCHDOWN_MASS),
                ],
                promotes_outputs=[("reserve_fuel", reserves_name)],
            )
```

aviary/utils/Fortran_to_Aviary.py
```diff
@@ -434,6 +434,13 @@ def update_gasp_options(vehicle_data):
         input_values.set_val(Aircraft.Wing.FLAP_DRAG_INCREMENT_OPTIMUM,
                              [[.12, .23, .13, .23, .23, .1, .15][flap_ind]])
 
+    res = input_values.get_val(Aircraft.Design.FIXED_RESERVES_FUEL, units='lbm')[0]
+    if res <= 1.0:
+        input_values.set_val(Aircraft.Design.FIXED_RESERVES_FUEL, [0], units='lbm')
+        input_values.set_val(Aircraft.Design.RESERVES_FRACTION, [res], units='unitless')
+    else:
+        input_values.set_val(Aircraft.Design.RESERVES_FRACTION, [0], units='unitless')
+
```
crecine, line 443:
```suggestion
    if res <= 0:
        input_values.set_val(Aircraft.Design.FIXED_RESERVES_FUEL, [0], units='lbm')
        input_values.set_val(Aircraft.Design.RESERVES_FRACTION, [-res], units='unitless')
    elif res > 10:
        input_values.set_val(Aircraft.Design.RESERVES_FRACTION, [0], units='unitless')
    else:
    	ValueError('"FRESF" is not valid between 0 and 10.')
```

aviary/utils/process_input_decks.py
crecine, line 1: It won't let me comment, because it is too far from any changes, but in update_options before initial_guessing is called:
image
add the following code (adjusting white space as necessary):
```
fixed_val = aircraft_values.get_val(Aircraft.Design.FIXED_RESERVES_FUEL, units='lbm')
fractional_val = aircraft_values.get_val(Aircraft.Design.RESERVES_FRACTION, units='unitless')
if fixed_val and fractional_val:
	raise RuntimeError(f'"{Aircraft.Design.FIXED_RESERVES_FUEL}" and "{Aircraft.Design.RESERVES_FRACTION}" cannot both be specified')
elif fractional_val:
	aircraft_values.set_val(Aircraft.Design.RESERVES_OPTION, Reserves_Type.Set_Fraction)
else:
	aircraft_values.set_val(Aircraft.Design.RESERVES_OPTION, Reserves_Type.Set_Direct)
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
             reserves = aircraft_values.get_val(
                 Aircraft.Design.RESERVES_FRACTION, units='unitless')
-        else:
-            reserves = initial_guesses['reserves']
-            aircraft_values.set_val(Aircraft.Design.RESERVES_FRACTION,
-                                    reserves, units='unitless')
-        if reserves < -1 or reserves > 0:
-            raise ValueError(
-                f'"{Aircraft.Design.RESERVES_FRACTION}" must be in [-1, 0].')
+            reserves_option = "fraction_val"
```
crecine, line 175:
```suggestion
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
-        if reserves < 10:
-            raise ValueError(
-                f'"{Aircraft.Design.RESERVES}" must be greater than 10 lbm.')
-    elif reserves_option == Reserves_Type.Set_Fraction:
-        if initial_guesses['reserves'] == 0:
+    if initial_guesses['reserves'] == 0:
+        reserves = aircraft_values.get_val(
+            Aircraft.Design.FIXED_RESERVES_FUEL, units='lbm')
+        reserves_option = "direct_val"
+        if reserves == 0:
```
crecine, line 172:
```suggestion
    	num_pax = aircraft_values.get_val(Aircraft.CrewPayload.NUM_PASSENGERS)
	reserve_option = aircraft_values.get_val(Aircraft.Design.RESERVE_OPTION)
	if reserve_option is Reserves_Type.Set_Direct:
	        reserves = aircraft_values.get_val(
	            Aircraft.Design.FIXED_RESERVES_FUEL, units='lbm')
        elif reserve_option is Reserves_Type.Set_Fraction:
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
-                                    reserves, units='unitless')
-        aircraft_values.set_val(Aircraft.Design.RESERVES_OPTION,
-                                reserves_option, units='unitless')
+        reserves = initial_guesses['reserves']
+        if reserves < 0.0:
+            raise ValueError('initial_guesses["reserves"] must be greater than 0.')
+        elif reserves > 10:
+            reserves_option = "direct_val"
+        else:
+            reserves_option = "fraction_val"
```
crecine, line 183:
```suggestion
        if reserves < 0.0:
            reserves *= -(num_pax * initial_guesses['fuel_burn_per_passenger_mile'] *
                     aircraft_values.get_val(Mission.Design.RANGE, units='NM'))
        elif reserves < 10:
            raise ValueError('initial_guesses["reserves"] is not valid between 0 and 10.')
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
-    if reserves_option == Reserves_Type.Set_Fraction:
-        if reserves <= 0:
-            reserves *= -(num_pax * initial_guesses['fuel_burn_per_passenger_mile'] *
-                          aircraft_values.get_val(Mission.Design.RANGE, units='NM'))
-        else:
-            raise ValueError(
-                f'The case that "{Aircraft.Design.RESERVES_FRACTION}" is between 0 and 1 is not implemented.')
+    if reserves_option == "fraction_val":
+        reserves *= (num_pax * initial_guesses['fuel_burn_per_passenger_mile'] *
+                     aircraft_values.get_val(Mission.Design.RANGE, units='NM'))
```
crecine, line 189:
```suggestion
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
-                f'"{Aircraft.Design.RESERVES}" must be greater than 10 lbm.')
-    elif reserves_option == Reserves_Type.Set_Fraction:
-        if initial_guesses['reserves'] == 0:
+    if initial_guesses['reserves'] == 0:
+        reserves = aircraft_values.get_val(
+            Aircraft.Design.FIXED_RESERVES_FUEL, units='lbm')
+        reserves_option = "direct_val"
+        if reserves == 0:
             reserves = aircraft_values.get_val(
                 Aircraft.Design.RESERVES_FRACTION, units='unitless')
```
crecine, line 174:
```suggestion
            reserves = aircraft_values.get_val(
                Aircraft.Design.RESERVES_FRACTION, units='unitless') *
              	(num_pax * initial_guesses['fuel_burn_per_passenger_mile'] *
                aircraft_values.get_val(Mission.Design.RANGE, units='NM'))
```

aviary/utils/process_input_decks.py
```diff
@@ -167,56 +165,28 @@ def update_dependent_options(aircraft_values: AviaryValues(), dependent_options)
...
-                                reserves_option, units='unitless')
+        reserves = initial_guesses['reserves']
+        if reserves < 0.0:
+            raise ValueError('initial_guesses["reserves"] must be greater than 0.')
+        elif reserves > 10:
+            reserves_option = "direct_val"
+        else:
+            reserves_option = "fraction_val"
 
     num_pax = aircraft_values.get_val(Aircraft.CrewPayload.NUM_PASSENGERS)
```
crecine, line 185:
```suggestion
```

crecine (changes requested):

aviary/interface/methods_for_level2.py
xjjiang: Now, allow this case.

aviary/docs/getting_started/input_csv_phase_info.md
```diff
@@ -18,14 +18,13 @@ This section is under development.
...
+The initial guess of `reserves` is used to define the reserve fuel. Initially, its value can be anything larger than or equal to 0. There are two Aviary variables to control the reserve fuel in the model file (`.csv`):
+- `Aircraft.Design.FIXED_RESERVES_FUEL`: the required fuel reserves: directly in lbm,
+- `Aircraft.Design.RESERVES_FRACTION`: the required fuel reserves: given as a proportion of mission fuel.
 
-If the value of parameter `reserves` is not 0 and is within the allowed range, the initial guess of reserve fuel is determined by the parameter `reserves`:
-- if `reserves > 10.0`, we set `Aircraft.Design.RESERVES = reserves` and `Aircraft.Design.RESERVES_OPTION = 1`,
-- if `-1.0 <= reserves <= 0`, we set `Aircraft.Design.RESERVES_FRACTION = reserves` and `Aircraft.Design.RESERVES_OPTION = 2`.
+Users should set one of them to zero. If the value of initial guess of `reserves` (also in the model file if any) is 0, the initial guess of reserve fuel comes from the above two Aviary variables. Otherwise, it is determined by the parameter `reserves`:
```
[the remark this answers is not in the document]
xjjiang: Yes, they are both initialized to 0. It is Fortran_to_Aviary that sets one of them to 0.

aviary/docs/getting_started/input_csv_phase_info.md
```diff
@@ -18,14 +18,13 @@ This section is under development.
...
+- if `reserves > 10`, we assume it is the actual fuel reserves.
+- if `0.0 <= reserves <= 10`, we assume it is the fraction of the mission fuel. The case when `reserves` is between 1 and 10 is non-sensical but technically valid.
```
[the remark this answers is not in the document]
xjjiang: After discussing with Carl, we decided to keep `reserves` on the positive side.

aviary/models/test_aircraft/converter_configuration_test_data_GwGm.dat
```diff
@@ -209,9 +209,7 @@ aircraft:engine:data_file=models/engines/turbofan_23k_1.deck
  OFEM=.78,	! off design specified mission Mach number
  ROSCAB=9999.,	! cabin rate of sink during descent (500.) fpm
  RSMX=900.,	! maximum allowable rate of sink during landing approach ft per min (1000)
- FRESF=0,   ! Reserve Fuel
- FRESF2=-0.15,   ! Reserve fuel fraction of mission fuel (negative sign trigger)
- FRESF3=2,  ! Use Reserve Fuel fraction
+ FRESF=0.15,   ! Reserve fuel fraction of mission fuel (negative sign trigger)
```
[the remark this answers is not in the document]
xjjiang: After discussing with Carl, we decided to keep reserves on the positive side.

aviary/variable_info/variables.py
```diff
@@ -131,8 +131,8 @@ class Design:
         EXTERNAL_SUBSYSTEMS_MASS = 'aircraft:design:external_subsystems_mass'
         FINENESS = 'aircraft:design:fineness'
         FIXED_EQUIPMENT_MASS = 'aircraft:design:fixed_equipment_mass'
+        FIXED_RESERVES_FUEL = 'aircraft:design:fixed_reserves_fuel'
         FIXED_USEFUL_LOAD = 'aircraft:design:fixed_useful_load'
-        IJEFF = 'aircraft:design:ijeff'
```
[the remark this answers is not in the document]
xjjiang: I put it back.

aviary/variable_info/variable_meta_data.py
```diff
@@ -1049,26 +1049,27 @@
...
+    historical_name={"GASP": 'INGASP.WFUL',
                      "FLOPS": None,
                      "LEAPS1": None
                      },
-    desc="A flag used by Jeff V. Bowles to debug GASP code during his 53 years supporting the development of GASP. \
-        This flag is planted here to thank him for his hard work and dedication, Aviary wouldn't be what it is today \
-        without his help.",
+    units='lbm',
+    desc='total mass of fixed useful load: crew, service items, trapped oil, etc',
 )
```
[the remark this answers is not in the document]
xjjiang: I put it back.

aviary/variable_info/variables.py
```diff
@@ -131,8 +131,8 @@ class Design:
         EXTERNAL_SUBSYSTEMS_MASS = 'aircraft:design:external_subsystems_mass'
         FINENESS = 'aircraft:design:fineness'
         FIXED_EQUIPMENT_MASS = 'aircraft:design:fixed_equipment_mass'
+        FIXED_RESERVES_FUEL = 'aircraft:design:fixed_reserves_fuel'
```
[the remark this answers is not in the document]
xjjiang: It was Jason's suggestion. @jkirk5 which one is better?

aviary/variable_info/variable_meta_data.py
xjjiang: Jason and Ken want to remove it. @jkirk5 and @Kenneth-T-Moore , what do you think? Should I keep it?

crecine:
I like RESERVE_FUEL_FIXED and RESERVE_FUEL_FRACTION

aviary/models/test_aircraft/converter_configuration_test_data_GwGm.dat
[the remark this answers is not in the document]
crecine: FRESF in the .dat should still be negative

aviary/docs/getting_started/input_csv_phase_info.md
```diff
@@ -18,14 +18,13 @@ This section is under development.
 - climb_range: 0,
 - reserves: 0
 
-The initial guess of `reserves` is used to define the reserve fuel. Initially, its value can be anything larger than 10 or anything between -1 and 0. There are three variables to control the reserve fuel in the model file (`.csv`):
-- `Aircraft.Design.RESERVES`: the required fuel reserves: directly in lbm,
-- `Aircraft.Design.RESERVES_FRACTION`: the required fuel reserves: given as a proportion of mission fuel, 
-- `Aircraft.Design.RESERVES_OPTION`: if `Set_Direct` (1), use `Aircraft.Design.RESERVES`; if `Set_Fraction` (2), use `Aircraft.Design.RESERVES_FRACTION`.
+The initial guess of `reserves` is used to define the reserve fuel. Initially, its value can be anything larger than or equal to 0. There are two Aviary variables to control the reserve fuel in the model file (`.csv`):
```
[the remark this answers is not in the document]
xjjiang: I discussed with Carl. He agreed that we will keep `reserves` consistent with `Aircraft.Design.FIXED_RESERVES_FUEL`. So, we will not allow `reserves` to go negative.

aviary/utils/process_input_decks.py
xjjiang: `num_pax` is used in the next blocks. I can not remove it.

aviary/variable_info/variables.py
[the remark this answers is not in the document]
xjjiang: I use `RESERVE_FUEL_FIXED` and `RESERVE_FUEL_FRACTION` now.

aviary/interface/methods_for_level2.py
xjjiang: I haven't decided if I will use `Aircraft.Design.RESERVES_OPTION`. So, I will hold on this suggestion.

aviary/utils/Fortran_to_Aviary.py
xjjiang: We have decided not to allow res to go negative.

crecine: In Fortran to Aviary, a negative value for FRESF still needs to be supported, but this stores RESERVES_FRACTION as a positive number

jdgratz10 (approved):

aviary/docs/getting_started/input_csv_phase_info.md
```diff
@@ -6,6 +6,28 @@
...
+- time_to_climb: 0,
+- climb_range: 0,
+- reserves: 0
+
+The initial guess of `reserves` is used to define the reserve fuel. Initially, its value can be anything larger than or equal to 0. There are two Aviary variables to control the reserve fuel in the model file (`.csv`):
+- `Aircraft.Design.RESERVE_FUEL_ADDITIONAL`: the required fuel reserves: directly in lbm,
+- `Aircraft.Design.RESERVE_FUEL_FRACTION`: the required fuel reserves: given as a proportion of mission fuel.
+
+If the value of initial guess of `reserves` (also in the model file if any) is 0, the initial guess of reserve fuel comes from the above two Aviary variables.  Otherwise, it is determined by the parameter `reserves`:
+- if `reserves > 10`, we assume it is the actual fuel reserves.
```
jkirk5, line 26: I'd to try and find a way to avoid this greater/less than 10 split for the initial guess too. @crecine does reserves initial guess need to exist at all anymore? Since the only way to define reserves right now don't involve "analysis" (beside a trivial equation) can we just use the user-provided variables and not make them enter that info twice?

jkirk5 (approved):
Some small questions but nothing critical, approving! Thanks for being flexible with our many change requests!

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
```diff
@@ -14,7 +14,8 @@ aircraft:design:drag_increment,0.00175,unitless
 aircraft:design:equipment_mass_coefficients,928,0.0736,0.112,0.14,1959,1.65,551,11192,5,3,50,7.6,12,unitless
 aircraft:design:max_structural_speed,402.5,mi/h
 aircraft:design:part25_structural_category,3,unitless
-aircraft:design:reserves,4998,unitless
+aircraft:design:reserve_fuel_additional,4998,lbm
+aircraft:design:reserve_fuel_fraction,0,unitless
```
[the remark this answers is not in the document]
crecine: If it is unspecified, it should default to 0, I'll double check. I believe this is just here because it was generated by the fortran to aviary converter

aviary/docs/getting_started/input_csv_phase_info.md
crecine: no, reserves is still an optional part of initial guessing

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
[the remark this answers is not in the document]
crecine: confirmed that removing it will default it to 0 and doesn't change the results. It is only needs to be here for testing the fortran to aviary converter.

crecine (approved):

PR #37, opened 2024-01-05
Check the interpolators and use more efficient versions where applicable.

### LINKED ISSUE #31: Interpolation Check (Legacy 54)

Many interpolants are used in Aviary. Check over all of them and make sure we use the fixed methods to improve speed.

Originally submitted by @jdgratz10 

_No response_

### Associated Bug Report

_No response_

Kenneth-T-Moore:
### Summary

1. Replaced a few interpolation methods with faster ones.
2. Adjusted the values on one of the `swap` benchmarks that was failing.
3. Added an upper bound on mass throughout the detailed takeoff phases so that IPOPT would perform better (and not hang.)

### Related Issues

- Resolves #31

This PR requires the following PR in OpenMDAO, which fixes a bug in the 2D-slinear interpolation: 

https://github.com/OpenMDAO/OpenMDAO/pull/3094

### New Dependencies

None


Kenneth-T-Moore:
Should I just switch all of our workflows over to use the OpenMDAO dev branch, once my fix there is merged? It will be a while before the next release.

johnjasa:
> Should I just switch all of our workflows over to use the OpenMDAO dev branch, once my fix there is merged? It will be a while before the next release.

I think this is worth discussing in the Aviary dev meeting.
I believe we don't want to require OpenMDAO dev because we want users to be able to `pip install openmdao` to use Aviary.
We may need to start to follow a policy of releasing Aviary soon after OM releases and allowing Aviary dev to depend on OM dev.

johnjasa (changes requested):
Putting a hold on this until the pypi release

.github/workflows/test_workflow_no_dev_install.yml
```diff
@@ -55,6 +55,16 @@ jobs:
           auto-update-conda: true
           python-version: ${{ matrix.PY }}
 
+      - name: Install OpenMDAO Dev
```
crecine, line 58: filename: test_workflow_**no_dev_install**.yml
test name: **Install** OpenMDAO **Dev**

Kenneth-T-Moore: yeah, with this PR, we can't run anything without dev.  My choices would be to remove this workflow, or just have it load "dev" until the next release, and then remove that task to go back to a no-dev run.  I went with the latter.

crecine: I was just pointing out the irony, not requesting changes

johnjasa (approved):

.github/workflows/test_workflow.yml
crecine, line 1: with the pypi release, can these be changed back from dev?

crecine (approved):

PR #44, opened 2024-01-08
Remove deepcopy of phase_info from the level 2 interface.

### LINKED ISSUE #40: Deepcopy of phase_info in AviaryProblem can cause problems.
### Description

In the AviaryProblem init, we do a deepcopy of the phase_info object. This was done to address a problem during testflo execution, where a test fails because the default phase_info was modified in some previous test. It is not a safe solution though, because the phase_info can contain an object that wraps an external code., and it is possible (maybe likely) for that to contain an attribute that can't (or shouldn't) be copied. 

The proper way to address the testing problem is, sadly, to copy the phase_info on a case-by-case basis in the tests or benchmarks where it is imported.

### Example

See MBSA&E VSP aero wrappers.

### Aviary Version

0.9.0-defv

_No response_

Kenneth-T-Moore:
### Summary

Removed the deepcopy of phase_info from the level 2 interface to prevent the potential problems when copying local attributes that aren't copyable.
- Update several tests to do a local deepcopy of phase_info so that the aviary default versions aren't modified during testing.

### Related Issues

- Resolves #40

None

### New Dependencies

None


Kenneth-T-Moore:
I made some changes to remove all of the intentional changes that level 2 makes to the phase_info object (essentially removed all the dictionary pops.)

crecine (approved):

PR #46, opened 2024-01-08
Add simple mission benchmarks

johnjasa:
### Summary

In preparation for combining the `simple` and `FLOPS` mission methods into one `height_energy` method, I've confirmed that the FLOPS swap benchmarks can be run in `simple` mission mode with these changes.

This PR adds two benchmarks that run the `simple` mission and checks against the same numbers as the `FLOPS` mission benchmarks.

I'll continue down this path and try to part out the changes into small PRs like this one.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/validation_cases/benchmark_tests/test_bench_GwFm.py
```diff
@@ -269,9 +266,108 @@ def bench_test_swap_1_GwFm(self):
...
+                          mission_method="simple", mass_method="GASP", max_iter=15)
+
+        compare_against_expected_values(prob, self.expected_dict, simple_flag=True)
 
 
 if __name__ == '__main__':
     z = ProblemPhaseTestCase()
-    z.bench_test_swap_1_GwFm()
+    z.setUp()
+    z.bench_test_swap_1_GwFm_simple()
```
crecine, line 373: I think this should be `test` (like in test_bench_FwFm.py) instead of `z`

aviary/validation_cases/benchmark_utils.py
```diff
@@ -5,7 +5,7 @@
 from aviary.utils.test_utils.assert_utils import warn_timeseries_near_equal
 
 
-def compare_against_expected_values(prob, expected_dict):
+def compare_against_expected_values(prob, expected_dict, simple_flag=False):
```
crecine, line 8: If this is temporary until the distinction between simple and FLOPS is removed, this is fine, but if it is more permanent I think it should be something more descriptive like use_simplified_mission

crecine (approved):

aviary/validation_cases/benchmark_utils.py
johnjasa: it is temporary

README.md
```diff
@@ -11,11 +11,10 @@ The user can select which type of mission analysis to use, as well as whether to
...
 ## Installation
 
-The simplest installation method for development is an "editable mode" install with ``pip`` in your terminal:
+The simplest installation method for users is to install via pip.
+From within the base folder of Aviary, perform this command in your terminal:
 
-    pip install -e .
-
-This installs the package in the current environment such that changes to the Python code don't require re-installation.This command should be performed while in the folder containing ``setup.py``.
+    pip install .
```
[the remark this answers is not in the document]
johnjasa: unless people do a dev install, yes indeed it'll be irrelevant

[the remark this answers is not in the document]
johnjasa: added words to clarify top-level folder not the `aviary` folder

aviary/interface/default_phase_info/simple.py
```diff
@@ -36,6 +36,7 @@
             "fix_duration": False,
             "initial_bounds": ((0.0, 0.0), "min"),
             "duration_bounds": ((64.0, 192.0), "min"),
+            "add_initial_mass_constraint": False,
```
jdgratz10, line 39: Why are we making this false? If I remember correctly, this is the constraint that ensures the GTOW used in the pre-mission is the same as the initial mass? Seems important...

aviary/interface/methods_for_level2.py
```diff
@@ -2361,16 +2361,30 @@ def _add_flops_landing_systems(self):
...
+                                   'traj.climb.initial_states:altitude')
+            else:
+                pass
+                # TODO: connect this correctly
+                # mass is the most important to connect but these others should
+                # be connected as well
+                # self.model.connect(Mission.Takeoff.FINAL_VELOCITY,
+                #                 'traj.climb.initial_states:mach')
+                # self.model.connect(Mission.Takeoff.FINAL_ALTITUDE,
+                #                 'traj.climb.controls:altitude')
```
jdgratz10, line 2377: Looks like for this case we are losing those initial connections, why did we do that?

aviary/interface/methods_for_level2.py
```diff
@@ -2361,16 +2361,30 @@ def _add_flops_landing_systems(self):
...
         self.model.connect('traj.descent.states:mass',
                            Mission.Landing.TOUCHDOWN_MASS, src_indices=[-1])
         # TODO: approach velocity should likely be connected
-        self.model.connect('traj.descent.states:altitude', Mission.Landing.INITIAL_ALTITUDE,
-                           src_indices=[-1])
+        if self.mission_method == "FLOPS":
+            self.model.connect('traj.descent.states:altitude', Mission.Landing.INITIAL_ALTITUDE,
+                               src_indices=[-1])
+        else:
+            self.model.connect('traj.descent.control_values:altitude', Mission.Landing.INITIAL_ALTITUDE,
```
jdgratz10, line 2386: if the mission method is GASP, wouldn't this throw an error because altitude is a state?

jdgratz10 (commented):
I had a few questions so I didn't approve or request changes yet, whether I have changes to request is TBD based on the answers :)

aviary/interface/methods_for_level2.py
johnjasa: These connections remain for the FLOPS based mission. I've added commented out code here for the simple mission to show that we need to connect it later.

We cannot do these connections now because the simplified takeoff outputs velocity whereas the simple mission deals with Mach. We cannot do a direct connection but could do a constraint. Instead, I will make the simplified takeoff output Mach as well and connect that directly in a follow-on PR.

We do not lose any capability the changes in this PR; this is more of a note for the future.

aviary/interface/methods_for_level2.py
johnjasa: This logic branch is never reached in the case of GASP missions because the method is only called for FLOPS-based missions.

aviary/interface/utils/check_phase_info.py
```diff
@@ -227,7 +227,7 @@ def check_phase_info(phase_info, mission_method):
         return
     else:
         raise ValueError(
-            "Invalid mission_method. Please choose either 'FLOPS', 'GASP', 'simple', or 'solved'.")
+            "Invalid mission_method. Please choose from 'HEIGHT_ENERGY', 'TWO_DEGREES_OF_FREEDOM', 'SIMPLE', or 'SOLVED'.")
```
crecine, line 230: so we don't have to keep updating enums as they change, could we do something like:
```
possible_values = ["'"+e.value+"'" for e in Equations_of_Motion]
possible_values[-1] = "or " + possible_values[-1]
raise ValueError("Invalid mission_method. Please choose from " + ", ".join(possible_values) + ".")
```

johnjasa: Thanks Carl, fun idea, added it and a test for it.

aviary/interface/default_phase_info/simple.py
jdgratz10: Ahh gotcha, that's what I get for remembering :)

aviary/interface/methods_for_level2.py
jdgratz10: Understood, thanks for the explanation!

jdgratz10 (approved):

PR #72, opened 2024-01-16
 Passenger Count Mismatch #394 

gawrenn:
Changed the passenger count for the large single aisle models to 169 from 180. If appears the 180 was inadvertently taken from the passenger mass value. Reinstated the preprocessor.py check for consistency between total passengers and individual passenger class data.

### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/models/test_aircraft/aircraft_for_bench_GwFm.csv
+++ b/aviary/models/test_aircraft/aircraft_for_bench_GwFm.csv
@@ -21,7 +21,7 @@ aircraft:crew_and_payload:num_flight_attendants,3,unitless
 aircraft:crew_and_payload:num_flight_crew,2,unitless
 aircraft:crew_and_payload:num_galley_crew,0,unitless
 aircraft:crew_and_payload:num_non_flight_crew,3,unitless
-aircraft:crew_and_payload:num_passengers,180,unitless
+aircraft:crew_and_payload:num_passengers,169,unitless
 aircraft:crew_and_payload:num_tourist_class,158,unitless
 aircraft:crew_and_payload:passenger_mass_with_bags,200,lbm
 aircraft:crew_and_payload:passenger_service_mass_scaler,1.0,unitless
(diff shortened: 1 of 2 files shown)


gawrenn:
This PR addresses issue #394.  It appears that in some data files the total passenger count for the large single aisle transport was set at 180 when it should have been 169.  This number may have accidentally been taken from the mass per passenger input of 180 lbm.
I reinstated the check for consistency between total passengers and passengers by class.  The test had been commented out. 
It now checks only if passengers by class and total passengers are both entered.

crecine (changes requested):
For the GASP based large_single_aisle cases, the passenger count should be 180 (the weight per passenger is 200).
For the swap case (GwFm) we can use 169.

Kenneth-T-Moore:
Have you run the benchmarks? I expect that some of the values will change because of the change in the number of passengers.

gawrenn:
I've now only changed the one data file that had both the passenger class data and the total passenger count present and mismatched.  The test for this mismatch is reinstated in preprocessors.py.

crecine (approved):

gawrenn:
Benchmark test 'test_bench_GwFm.py ' fails with a mass value mismatch.  The new test with 169 passengers instead of 180 passengers yields:
ValueError: actual [51029.61545852], desired [52955.84313972], rel error 0.036374223636145744, tolerance 0.02

I also need one more reviewer.

crecine:
> Benchmark test 'test_bench_GwFm.py ' fails with a mass value mismatch. The new test with 169 passengers instead of 180 passengers yields: ValueError: actual [51029.61545852], desired [52955.84313972], rel error 0.036374223636145744, tolerance 0.02

Since you reduced the passenger count by 11, it makes sense that the weight decreased ~2000 pounds. Update the value in the test with the new one.

johnjasa (approved):

PR #74, opened 2024-01-17
Issue 11 Resolution - ALTITUDE_RATE_MAX add to GASP ODEs

### LINKED ISSUE #11: Standardize required altitude rate naming
### Summary of Issue

In some places in Aviary, we have multiple names for effectively the same thing: `required_altitude_rate` and `roc_at_toc`.
Pick one name -- probably the former as verbosity is helpful -- and use it everywhere.

### Issue Type

- [ ] Bug
- [ ] Enhancement
- [x] Code Cleanup
- [ ] Docs
- [ ] Miscellaneous

Thinking more about this now, this is slightly more complex than just renaming.
I'll put my thoughts here.

GASP- and FLOPS-based missions handle rate of climb at top of climb differently. Aviary currently has no constraint on the GASP side but it exists in the input decks. FLOPS-based has `required_altitude_rate` come in at the `phase_info` level. This is valid for any phase and is enforced as a path constraint across the whole phase, not just climb.

We might as well rework this so that:
- the required altitude rate is available to be enforced anywhere, not just a phase called `climb`
- it's handled in the same way for all types of missions with a constraint at the `add_design_variables()` method level from `methods_for_level2.py`
- to do this, we'd need to compute `ALTITUDE_RATE_MAX` as part of the GASP equations of motion, which is not currently the case

### Associated Bug Report

_No response_

ehariton:
### Summary

Adding `ALTITUDE_RATE_MAX` as output from GASP ODEs

Status - simple implementation just on Climb_ODE failing tests because of no access to `THRUST_MAX_TOTAL`

### Related Issues

Works towards #11 but we need #88 in first.

None

### New Dependencies

None


aviary/xdsm/climb_xdsm.py
```diff
@@ -49,8 +51,10 @@
     Aircraft.Engine.SCALE_FACTOR,
 ])
 x.add_input("eom", [Dynamic.Mission.MASS])
+x.add_input("specific_energy", [Dynamic.Mission.MASS])
+x.add_input("alt_rate", ["TAS_rate"])
 
-# make connections
+# make connections6
```
crecine, line 57:
```suggestion
# make connections
```

johnjasa (approved):
I dig this! I'm going to continue this work once #88 is in. That'll make it easier to expose this in a cohesive way.

crecine (approved):

PR #90, opened 2024-01-22
Bugfix for SGM

crecine:
### Summary

references to engine.nox vs engine_deck.nox were causing problems
minor improvements to debugging and declaring inputs in preparation for HeightEnergy SGM

### Related Issues

- Resolves #

None

### New Dependencies

None


johnjasa:
Thanks for this quick fix, Carl! Is there a test or something we could add that would catch something like the `engine_deck.nox` issue that Joerg encountered?

crecine:
> Thanks for this quick fix, Carl! Is there a test or something we could add that would catch something like the `engine_deck.nox` issue that Joerg encountered?

Done

johnjasa (approved):

jkirk5 (approved):

PR #93, opened 2024-01-22
Updated GwGm truth values and added workflow for benchmarks

### LINKED ISSUE #63: Run benchmarks as part of CI

We need to run benchmarks as part of CI.
To do all of them we'd have to install SNOPT.
I don't want to do this.
Let's make sure they run with IPOPT and add a workflow for it.

_No response_

### Associated Bug Report

_No response_

johnjasa:
### Summary

Benchmark for GwGm fails on main.
I updated the values here and also introduced a workflow that runs the benches so regressions will be caught by CI in the future.

### Related Issues

- Resolves #63

None

### New Dependencies

None

### DIFF
--- a/aviary/validation_cases/benchmark_tests/test_FLOPS_based_sizing_N3CC.py
+++ b/aviary/validation_cases/benchmark_tests/test_FLOPS_based_sizing_N3CC.py
@@ -13,6 +13,7 @@ import openmdao.api as om
 import scipy.constants as _units
 from openmdao.utils.assert_utils import assert_near_equal
 from openmdao.utils.testing_utils import use_tempdirs
+from openmdao.utils.testing_utils import require_pyoptsparse
 
 from aviary.mission.flops_based.phases.build_landing import Landing
 from aviary.mission.flops_based.phases.build_takeoff import Takeoff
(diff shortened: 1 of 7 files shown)


aviary/validation_cases/benchmark_tests/test_FLOPS_based_sizing_N3CC.py
```diff
@@ -274,7 +275,7 @@ def run_trajectory(sim=True):
         promotes_outputs=['mission:*'])
 
     traj.link_phases(["climb", "cruise", "descent"], [
-                     "time", Dynamic.Mission.ALTITUDE, Dynamic.Mission.MASS, Dynamic.Mission.DISTANCE], connected=strong_couple)
```
crecine, line 277: why was altitude removed?

aviary/validation_cases/benchmark_tests/test_bench_GwGm.py
```diff
@@ -13,26 +12,26 @@
...
         assert_near_equal(prob.get_val(Aircraft.Design.OPERATING_MASS),
-                          95509, tolerance=rtol)
+                          101555., tolerance=rtol)
 
         assert_near_equal(prob.get_val(Mission.Summary.TOTAL_FUEL_MASS),
-                          42529., tolerance=rtol)
+                          44098., tolerance=rtol)
 
         assert_near_equal(prob.get_val('landing.' + Mission.Landing.GROUND_DISTANCE),
-                          2634.8, tolerance=rtol)
```
crecine, line 35: where did these updated values come from?

aviary/validation_cases/benchmark_tests/test_FLOPS_based_sizing_N3CC.py
johnjasa: altitude is not a design variable here but is a constraint. prior versions of OM warned about this, recent ones throw an error.

aviary/validation_cases/benchmark_tests/test_bench_GwGm.py
johnjasa: running the benchmark on main

johnjasa:
Closing until we determine what caused the GwGm values to change

PR #103, opened 2024-01-24
Notify teams

crecine:
### Summary

This creates a workflow that will send notifications to the Aviary pull request channel.
If reviewers are assigned when the pull request is created, the first two will be tagged in the tasks.
Currently, reviewers added after the PR is opened will not be tagged.

### Related Issues

- Resolves #

None

### New Dependencies

None


robfalck (approved):
Interesting, I hadn't tried to do anything like this in the other repos yet. I don't see any problems with the secrets, and I'm interested to see how well this works.

crecine:
> Interesting, I hadn't tried to do anything like this in the other repos yet. I don't see any problems with the secrets, and I'm interested to see how well this works.

I got the issue with the secrets worked out. The issue was using `pull_request` vs `pull_request_target`, the first runs in the repo the changes are coming from, the latter runs in the repo the changes are being merged into.
I can walk you through the Teams side, if you'd like.

johnjasa (approved):

PR #94, opened 2024-01-22
Improvements to height energy mission based on Ken's feedback

johnjasa:
### Summary

This is a minor PR to resolve some capability needed for MBSA&E work.

- Resolves #85
- Resolves #86

### Backwards incompatibilities

None

### New Dependencies

None

PR #109, opened 2024-01-25
Cleaning up many warnings and increasing speed of API import

### LINKED ISSUE #70: Multiple (unused) example aircraft models loaded every Aviary run

Whenever the Aviary interface is used (at any level), the api is imported - this results in every single data-file based example case being loaded. This adds additional time (anecdotally ~1 sec extra on my laptop) and can generate warnings and errors that aren't actually related to the model being ran, as well as making debugging some files significantly more difficult.

_No response_

### Associated Bug Report

_No response_

johnjasa:
### Summary

I've got a bevy of smallish changes.
- changed many small causes of warnings
- avoided importing test data in the API
- changed Dymos prefix usage so things in the `timeseries` are no longer `timeseries:states` (for example) and instead they are just `timeseries` following Dymos v1.8 changes

This PR includes #94 so please review and merge that first if not already done

### Related Issues

- Resolves #70 

None

### New Dependencies

None


aviary/interface/default_phase_info/two_dof.py
crecine, line 1: why these changes?

aviary/interface/methods_for_level2.py
```diff
@@ -604,11 +604,9 @@ def _get_2dof_phase(self, phase_name):
         if 'cruise' not in phase_name:
             phase.add_control(
                 Dynamic.Mission.THROTTLE, targets=Dynamic.Mission.THROTTLE, units='unitless',
-                opt=False, lower=0.0, upper=1.0
```
crecine, line 607: why are we removing these limits?

aviary/mission/gasp_based/phases/rotation_phase.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+from dymos.utils.misc import _unspecified
```
crecine, line 2: unused?

aviary/interface/methods_for_level2.py
johnjasa: not optimizer-controlled so limits don't mean nothing

aviary/interface/default_phase_info/two_dof.py
johnjasa: removing the initial bounds? because all time is connected so the initial bounds do not matter as `input_initial` is true for each phase

aviary/mission/gasp_based/phases/rotation_phase.py
johnjasa: thanks, removed

jkirk5 (approved):

aviary/interface/default_phase_info/two_dof.py
crecine: ahh, thanks

crecine (approved):

PR #124, opened 2024-01-31
Convert from debug_mode to verbosity

### LINKED ISSUE #68: Fully implement new VERBOSITY settings

With the introduction of the `Settings` category to the variable hierarchy, we now have a official way to control output via the `VERBOSITY` flag. This variable now needs to be propagated through the code. A few immediate fixes:
- Replace the `debug_mode` flag everywhere it appears, and replace with the appropriate `VERBOSITY` level. This includes modifying fortran_to_aviary to no longer hardcode adding in `debug_mode`, and all of the special handling needed to deal with this add-on variable
- Wrap all print statements in the code under checks for a minimum threshold `VERBOSITY` level

`VERBOSITY` enum values are ints, so greater than/less than comparators should be used for logic statements whenever possible (e.g. only print specific info if `VERBOSITY` is greater than 2, etc.)

_No response_

### Associated Bug Report

_No response_

crecine:
### Summary

converting from ad hoc debug mode implementations with a unified verbosity

### Related Issues

- Resolves #68 

None

### New Dependencies

None


crecine:
I initially converted everything to Verbosity.DEBUG or Verbosity.QUIET, but would appreciate feedback on what should be BRIEF or VERBOSE

aviary/utils/Fortran_to_Aviary.py
```diff
@@ -39,7 +38,7 @@
 
 
 def create_aviary_deck(fortran_deck: str, legacy_code=None, defaults_deck=None,
-                       out_file=None, force=False):
+                       out_file=None, force=False, verbosity=Verbosity.QUIET):
```
Kenneth-T-Moore, line 41: I feel like the default ought to be Verbosity.Brief.  There is a certain amount of information that we probably should print out for the user (I think Overrides fall into this category.)  Then, if they want to clear everything out, they can go quiet (and we might even suppress openmado warnings for that one.)

I could also be persuaded that Verbose should be default. I think granularizing our messages between Brief and Verbose will take some thought.  Of course, that leaves Debug for the kitchen sink.

crecine: hmm, maybe:
debug - everything
verbose - default
brief - only warnings
quiet - nothing

crecine: or:
debug - everything
verbose - optimizer output
brief - default
quite - nothing

aviary/mission/gasp_based/ode/time_integration_base_classes.py
```diff
@@ -186,7 +187,7 @@ def __init__(
         # TODO: add defensive checks to make sure dimensions match in both setup and
         # calls
 
-        if DEBUG or True:
+        if verbosity is Verbosity.DEBUG or True:
```
jkirk5, line 190: The "or True" should get removed

aviary/utils/Fortran_to_Aviary.py
jkirk5: I lean towards that second description. However I think we need to let the user know if the optimization was successful or not in that case, even if we don't show them the whole optimizer output

jkirk5: I agree with @Kenneth-T-Moore about quiet operation suppressing non-critical warnings. That was kind of the intent for that mode, to suppress literally everything that wasn't a show-stopping bug. That way big parametric studies won't generate pages of terminal output, which really isn't helpful to anyone.

aviary/utils/Fortran_to_Aviary.py
```diff
@@ -613,5 +612,10 @@ def _exec_F2A(args, user_args):
         args.input_deck = args.input_deck[0]
     filepath = args.input_deck
 
+    if args.verbose is True:
```
jkirk5, line 615: I like the simplified "on-off" verbose switch for the command args here

aviary/utils/process_input_decks.py
```diff
@@ -143,7 +138,7 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
                 initial_guesses[var_name] = float(var_values[0])
                 continue
 
-            if 'debug_mode' in aircraft_values and aircraft_values.get_val('debug_mode'):
+            if aircraft_values.get_val('verbosity') is Verbosity.DEBUG:
```
jkirk5, line 141: I think it is really important for users to know if their inputs are getting skipped. But, we don't need to share the unconverted FORTRAN variables either... maybe we keep it debug for now. It would be neat if we could filter out all FLOPS and GASP variables.

aviary/mission/gasp_based/phases/time_integration_traj.py
```diff
@@ -235,46 +233,46 @@ def compute(self, inputs, outputs):
 
 #             # trajectory-specific phase switching
 #             if current_problem is groundroll:
-#                 if DEBUG:
+#                 if sel.verbosity is Verbosity.DEBUG:
```
jkirk5, line 236: "self", just in case we need this code again

aviary/mission/gasp_based/test/test_idle_descent_estimation.py
```diff
@@ -17,7 +18,7 @@ def test_case1(self):
         tol = 1e-5
         input_deck = 'models/large_single_aisle_1/large_single_aisle_1_GwGm.csv'
         aviary_inputs, _ = create_vehicle(input_deck)
-        aviary_inputs.set_val('debug_mode', False)
+        aviary_inputs.set_val('verbosity', Verbosity.QUIET)
```
jkirk5, line 21: Maybe we should make an issue to add either QUIET or BRIEF to all tests to reduce unnecessary outputs. How does everyone feel about test outputs?

crecine: I'm not sure if it is possible, but it would be nice if running testflo for the entire repository was QUIET and then failed tests were rerun with verbose.

aviary/utils/process_input_decks.py
crecine: changing to not QUIET for now, we can look into more detailed filtering as a follow on.

Kenneth-T-Moore (approved):

aviary/interface/methods_for_level2.py
```diff
@@ -1585,9 +1585,9 @@ def add_driver(self, optimizer=None, use_coloring=None, max_iter=50, debug_print
             driver.options["maxiter"] = max_iter
             driver.options["disp"] = True
 
-        if debug_print:
-            if isinstance(debug_print, list):
-                driver.options['debug_print'] = debug_print
+        if verbosity != Verbosity.QUIET:
+            if isinstance(verbosity, list):
+                driver.options['debug_print'] = verbosity
```
johnjasa, line 1590: Because `verbosity` is only ever an enum value, not a list, this logic doesn't hold. There wouldn't be a way to have the user set a custom amountn of output. Here the default verbosity results in `debug_print` being turned on; I'd suggest  it only being on for `VERBOSE` and `DEBUG` if you keep the default as `BRIEF`. Also you can just remove the list logic.

johnjasa (approved):
Minor note and change request; I don't want it to hold up approval. Thanks for this!

PR #73, opened 2024-01-16
Issue 32 Resolution - Range to Distance switch

ehariton:
### Summary

`Dynamic.Mission.RANGE` was removed and replaced with `Dynamic.Mission.DISTANCE`.

To implement this, some additional names were replaced:

- `initial_range` -> `initial_distance`
- `states:range` -> `states:distance`
- `_Dynamic.Mission` was removed from all files and longer but easier to find names were introduced.
- Imported `Dynamic` to test_FLOPS_based_sizing_N3CC.py to enable it to pass tests
- `"distance_rate"`, `Dynamic.Mission.RANGE_RATE` -> `Dynamic.Mission.DISTANCE_RATE`
- `.RANGE_RATE` - > `.DISTANCE_RATE` 
- (note there is still a file called range_rate which has a RangeRate() function which should eventually be renamed to DistanceRate()
- `"range"`, `'range'`, `"distance"`, `'distance'`, `Dynamic.Mission.RANGE` -> `Dynamic.Mission.DISTANCE`
- `Dynamic.Mission.RANGE`, `Dynamic.Mission.RANGE_RATE`, `RANGE`, & `RANGE_RATE` were removed from variables.py
- Old names of range_rate and range were retained in the comments in variable_meta_data.py

### Related Issues

- Resolves #32 

### Backwards incompatibilities

None

### New Dependencies

None

PR #126, opened 2024-02-01
Mission Report

jkirk5:
### Summary
Addition of mission summary report as well as minor reworks to subsystem reports
- Adds a mission reporting framework that includes a simplified table of mission parameters, broken up by phase.
- Reworks propulsion report to query each engine to add to the report, as well as provide system-level metrics.
- Cleans up formatting for some existing subsystem reports
- Fixes some bugs related to Aviary reports not always being generated.

### Related Issues

None

None

### New Dependencies

None


aviary/interface/methods_for_level2.py
```diff
@@ -2,6 +2,7 @@
 import warnings
 import inspect
 from pathlib import Path
+from copy import deepcopy
```
crecine, line 5: unused?

aviary/interface/reports.py
```diff
@@ -42,3 +52,118 @@ def subsystem_report(prob, **kwargs):
...
+                        vals = prob.get_val(f"{traj}.{phase}.{var_name}",
+                                            units=units,
+                                            indices=indices,)
+                    # 2DOF breguet range cruise uses time integration to track mass
+                    except TypeError:
+                        vals = prob.get_val(f"{traj}.{phase}.timeseries.time",
+                                            units=units,
+                                            indices=indices,)
+                    except KeyError:
+                        vals = None
```
johnjasa, line 93: Do all of these try/except blocks strictly need to be here? I'm thinking that most of the values we care about from the mission should be added to the timeseries (so they'd be accessible in the first try block).

crecine, line 93: could this be changed to something like:
```
key_paths = [f"timeseries.{var_name}", f"states:{var_name}", f"timeseries.states:{var_name}", f"{var_name}", "timeseries.time"]
vals = None
for key_path in key_paths:
	try:
		vals = prob.get_val(f"{traj}.{phase}.{key_path}", units=units, indices=indices,)
		return vals
	except KeyError:
		pass
```

aviary/interface/reports.py
```diff
@@ -42,3 +52,118 @@ def subsystem_report(prob, **kwargs):
...
+    for idx, phase in enumerate(prob.phase_info):
+        # TODO for traj in trajectories, currently assuming single one named "traj"
+        # TODO delta mass and fuel consumption need to be tracked separately
+        fuel_burn = _get_phase_diff('traj', phase, 'mass', 'lbm', [-1, 0])
+        time = _get_phase_diff('traj', phase, 't', 'min')
+        range = _get_phase_diff('traj', phase, 'range', 'nmi')
+        # both "distance" and "range" currently exist to track ground distance, logic
+        # here accounts for that
+        if range is None:
+            range = _get_phase_diff('traj', phase, 'distance', 'nmi')
```
johnjasa, line 122: Do both distance and range still exist? I thought this PR resolved that by having just distance: https://github.com/OpenMDAO/Aviary/pull/73

johnjasa (changes requested):
Looks slick! Thank you. I added some minor questions.

aviary/subsystems/propulsion/propulsion_builder.py
```diff
@@ -61,14 +61,12 @@ def report(self, prob, reports_folder, **kwargs):
...
             write_markdown_variable_table(f, prob, propulsion_outputs, self.meta_data)
-            f.write('\n## ENGINES')
-            for engine in prob.aviary_inputs.get_val('engine_models'):
-                f.write(f'\n### {engine.name}')
-                write_markdown_variable_table(f, engine, engine_outputs, self.meta_data)
+            f.write('\n## Engines')
+
+        # each engine can append to this file
+        kwargs['meta_data'] = self.meta_data
+        for engine in prob.aviary_inputs.get_val('engine_models'):
```
crecine, line 71: This won't put in duplicates, right?
(a conventional twin engine aircraft would only have one entry)

crecine (approved):
One suggestion that may or may not be possible and a couple of minor comments

aviary/subsystems/propulsion/propulsion_builder.py
jkirk5: Yeah, this is looping through each unique _type_ of engine

aviary/interface/reports.py
jkirk5: That will work for everything except catching the TypeError that only shows up for the timeseries.time check. I'll play around a little with this idea

jkirk5: Good point. I did some tests and some of these aren't required. I've removed the unneeded try/except blocks

johnjasa (approved):

PR #132, opened 2024-02-07
Refactor level2 to eliminate the input_sink and set all input defaults in the top group's configure method.

Kenneth-T-Moore:
### Summary

Prior to this change, the level 2 interface handled setting the model's input defaults by a call to a utility function right before setup. Since at this point, it isn't possible to figure out what inputs are in the model, we called set_input_defaults on everything in the aircraft metadata. Unfortunately, if you promote an input that isn't in the model, OpenMDAO raises an error. Our original workaround was to create the "input_sink" component, which contained almost every variable in the metadata dictionary. This approach caused other problems that we have gradually discovered...

This PR is a re-implementation, where we delay calling set_input_defaults until the top group's configure method. At this point, we know all of our subsystem's inputs and can query them, so we no longer need the input_sink.

1. Removed the "input_sink" component from the AviaryProblem interface.
2. Implemented a custom top level group (AviaryGroup) with a configure method that handles setting the input defaults.
3. Fixed a test that added an objective that didn't exist in the model.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/docs/getting_started/onboarding_level2.ipynb
+++ b/aviary/docs/getting_started/onboarding_level2.ipynb
@@ -550,8 +550,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "print(\"Zero Fuel Mass (lbm)\",\n",
-    "      prob.get_val(Aircraft.Design.ZERO_FUEL_MASS, units='lbm'))\n",
     "print(\"Mission.Objectives.FUEL\",\n",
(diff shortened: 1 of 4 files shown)


jkirk5 (approved):
Looks nice! Should help clean up the n2 as well

johnjasa (approved):
Absolutely slick stuff Ken, I love this!

PR #149, opened 2024-02-20
Reserve Missions for Height-Energy

ehariton:
### Summary

Reserve missions for Height Energy Missions, plus updates to reserve_fuel computations.

- Fix-time & fixed-range reserve missions for Height-Energy phases. 
- Some background work was added for SOLVED_2DOF to have access to reserve missions.
- Mass and Time automatically linked between the regular phases and the reserve phase.
- updated `_link_phases_helper_with_options()` to only link states between regular phases, or reserve phases. It won't link from regular to reserve.
- Single-phase and multi-phase reserve mission examples with arbitrary ordering of either time-based ore range-based reserves.
- Constraints can only be applied per-phase (no multi-phase range or time constraints)
- user must instantiate regular phases first, and afterward, all reserve phases (i.e. climb, cruise, descent, reserve_climb, reserve_cruise, reserve descent)
- Added check for above and error messages in `check_and_preprocess_inputs()` and `phase_separator()`

Reserve fuel calcs for height-energy and SOLVED_2DOF now sum: reserve_mission_burned + fractional + additional.
- Mission.Summary.FUEL_BURNED and Mission.Summary.RESERVE_FUEL_BURNED added to variable hierarchy
- `reserve_fuel` calcs updated to add Mission.Summary.RESERVE_FUEL_BURNED
- `Mission.Summary.FUEL_BURNED` calcs updated to only include fuel burned during regular phases
- `Mission.Summary.RESERVE_FUEL_BURNED` added and is the fuel burned during reserve phases
- fractional reserve fuel is now added only based on mission fuel, not reserve fuel

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/mission/flops_based/phases/energy_phase.py
```diff
@@ -146,6 +148,7 @@ def build_phase(self, aviary_options: AviaryValues = None):
 
         user_options: AviaryValues = self.user_options
 
+        reserve = user_options.get_val('reserve')  # do we need this here?
```
Kenneth-T-Moore, line 151: You probably don't need it, since the variable isn't used in build_phase. 

In fact, it doesn't even look like it is used at all in the builder or in what it builds.

aviary/examples/run_aviary_example.py
```diff
@@ -87,7 +87,6 @@
     },
     "post_mission": {
         "include_landing": False,
-        "constrain_range": True,
```
Kenneth-T-Moore, line 90: I see.  You got rid of "constrain_range" because it is redundant to have it and target_range. If you specify a target range, then you obviously are constraining it.

aviary/examples/run_reserve_mission_cruise_fixedrange.py
```diff
@@ -0,0 +1,155 @@
...
+We then call the correct methods in order to set up and run an Aviary optimization problem.
+This performs a coupled design-mission optimization and outputs the results from Aviary into the `reports` folder.
+"""
+import aviary.api as av
+
+
+phase_info = {
+    "pre_mission": {"include_takeoff": False, "optimize_mass": True},
+    "climb": {
+        "subsystem_options": {"core_aerodynamics": {"method": "computed"}},
```
Kenneth-T-Moore, line 17: I wonder if it would be better to load in a base phase_info from our aircraft models instead of repeating all of the regular phases here. It is redundant, and doesn't highlight the actual changes that are needed to activate the reserve missions.

So, instead of this big dictionary, just import phase_info, add the 'cruise_reserve" phase, and edit the keys in post_mission.  That might be a bit clearer for the user about what they need to do to run a setup a reserve run. Also, it is more maintainable, since all of the normal flight phases are coming from 1 place.

aviary/interface/methods_for_level2.py
```diff
@@ -1394,13 +1486,24 @@ def add_linkage_constraint(self, phase_a, phase_b, var_b):
             self.traj.link_phases(phases[:7], vars=["TAS"], units='kn', ref=200.)
 
         elif self.mission_method is HEIGHT_ENERGY:
+
+            # connect mass and distance between all phases regardless of reserve / non-reserve status
             self.traj.link_phases(
                 phases, ["time", Dynamic.Mission.MASS, Dynamic.Mission.DISTANCE], connected=True)
 
+            regular_phases, reserve_phases = self.phase_separator()
```
Kenneth-T-Moore, line 1494: This is done twice, but I think we ought to be able to just call it once.

aviary/interface/methods_for_level2.py
```diff
@@ -433,6 +432,44 @@ def _update_metadata_from_subsystems(self):
             if variable in self.meta_data:
                 self.meta_data.pop(variable)
 
+    def phase_separator(self):
+        """
+        This method checks for reserve=True & False
+        Returns errors if reserve = False no reserve no specified after reserve = True 
+        return two dictionaries of phases, regular_phases and reserve_phaes
+        """
```
Kenneth-T-Moore, line 440: What about a case where we have actual landing phases defined. In such a case, we probably want to "branch" the reserve mission phases so that they connect to the final descent phase before the landing phases.

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
```diff
@@ -0,0 +1,204 @@
...
+            "initial_bounds": ((120.5, 361.5), "min"),
+            "duration_bounds": ((29.0, 87.0), "min"),
+        },
+        "initial_guesses": {"times": ([241, 58], "min")},
+    },
+    "post_mission": {
+        "include_landing": False,
+        "target_range": (1906, "nmi"),
+    },
+    "climb_reserve": {
```
Kenneth-T-Moore, line 95: Just a note that there are a lot of values to fill in to define a reserve mission. How feasible would it be to let the user define a minimal set, say (distance, mach) and have aviary build the phases for it? Note: I am not sure how we define these in FLOPS, so I suspect it is simpler.

aviary/interface/methods_for_level2.py
```diff
@@ -1259,26 +1296,76 @@ def add_post_mission_systems(self, include_landing=True):
                 ],
                 promotes_outputs=[('overall_fuel', Mission.Summary.TOTAL_FUEL_MASS)])
 
-            if 'constrain_range' in self.post_mission_info:
-                if self.post_mission_info['constrain_range']:
-                    target_range = self.post_mission_info['target_range']
+            regular_phases, reserve_phases = self.phase_separator()
```
Kenneth-T-Moore, line 1299: So, i think the mass residual is a little tricky.  If you look at line 1287, you find the execcomp with the equation `overall_fuel = fuel_burned + fuel_reserve`.  The way things are done now, the `fuel_burned = starting mass - mass at end of final phase`. This means that, with the additional reserve phases, reserve mission is considered part of fuel_burned (start of climb - end of final reserve).  The `fuel_reserve` variable here is Mission.Design.RESERVE_FUEL, which is a user-specified amount in lbs, and this might be 0 for a model like this because we have a reserve mission instead.

Maybe it is semantics, but including the reserve fuel in the fuel burned is going to confuse users, who are expecting that variable to be just them mission fuel. A solution to this requires separately computing the fuel burned in the normal and the reserve phases, and adding it into this equation.  (Side note, i recommend adding it as another var, because Jason has proposed that users might want to run a reserve mission AND specify an additional reserve fuel in lbs.)

aviary/mission/flops_based/phases/energy_phase.py
ehariton: All the tests seem to work without it so I removed it.

aviary/examples/run_reserve_mission_cruise_fixedrange.py
ehariton: Done!

aviary/interface/methods_for_level2.py
ehariton: We don't have any sense of connecting anything but mass & distance between the regular phases and the reserve phases. I think we'd need a follow-on PR do start dealing with / letting the user specify tandem phases.

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
ehariton: I think ideally these would be filled in by the phase creator UI to start with, so then if the user went into the phase_info file created by the UI they could have all of these things already set. I do agree it's a lot of items and I don't think we have a sense of "defaults right now". But defaults is more "magic", and "magic" isn't all good.

aviary/interface/methods_for_level2.py
```diff
@@ -433,6 +432,44 @@ def _update_metadata_from_subsystems(self):
             if variable in self.meta_data:
                 self.meta_data.pop(variable)
 
+    def phase_separator(self):
+        """
+        This method checks for reserve=True & False
+        Returns errors if reserve = False no reserve no specified after reserve = True
```
crecine, line 438:
```suggestion
        Returns an error if a non-reserve phase is specified after a reserve phase.
```

aviary/interface/methods_for_level2.py
```diff
@@ -433,6 +432,44 @@ def _update_metadata_from_subsystems(self):
...
+                    # This is a reserve phase
+                    reserve_phases.append(phase_name)
+                    start_reserve = True
+            else:
+                # This is a regular phase by default
+                regular_phases.append(phase_name)
+                if start_reserve is True:
+                    raise_error = True
+
+        if raise_error is True:
```
crecine, line 464:
```suggestion
        if raise_error:
```

aviary/interface/methods_for_level2.py
```diff
@@ -1228,7 +1265,7 @@ def add_post_mission_systems(self, include_landing=True):
                                                 subsystem_postmission)
 
         if self.mission_method is HEIGHT_ENERGY:
-            phases = list(self.phase_info.keys())
+            phases = list(self.phase_info.keys())  # list of all phases included reserve
```
crecine, line 1268:
```suggestion
            phases = list(self.phase_info.keys())  # list of all phases including reserve
```

aviary/interface/methods_for_level2.py
crecine: we could use block_fuel for fuel_burned and reserve_mission_fuel for the fuel used during reserve phases (separate from the fixed reserve_fuel)

aviary/interface/methods_for_level2.py
```diff
@@ -1308,17 +1395,22 @@ def add_post_mission_systems(self, include_landing=True):
...
     def _link_phases_helper_with_options(self, phases, option_name, var, **kwargs):
+        # if there is a phase that has 'option_name' of optimize_mach or optimize_altitude = True,
+        # link that phase to the adacent phases passed in via 'phases'
+
         phases_to_link = []
-        for idx, phase_name in enumerate(self.phase_info):
-            if self.phase_info[phase_name]['user_options'][option_name]:
-                # get the name of the previous phase
+        for idx, phase_name in enumerate(phases):
+            if self.phase_info[phase_name]['user_options'][option_name] is True:
```
crecine, line 1403:
```suggestion
            if self.phase_info[phase_name]['user_options'][option_name]:
```

aviary/interface/methods_for_level2.py
```diff
@@ -1394,13 +1486,24 @@ def add_linkage_constraint(self, phase_a, phase_b, var_b):
...
 
+            regular_phases, reserve_phases = self.phase_separator()
+
+            # connect regular phases with eachother if your are optimizing alt or mach
+            self._link_phases_helper_with_options(
+                regular_phases, 'optimize_altitude', Dynamic.Mission.ALTITUDE, ref=1.e4)
+            self._link_phases_helper_with_options(
+                regular_phases, 'optimize_mach', Dynamic.Mission.MACH)
+
+            # connect reserve phases with eachother if your are optimizing alt or mach
```
crecine, line 1502:
```suggestion
            # connect reserve phases with each other if you're are optimizing alt or mach
```

crecine: it looks like these are happening every time, not just when optimizing alt or mach?

crecine (commented):
a couple of comments/suggestions

aviary/interface/methods_for_level2.py
ehariton: I looked through the variable metadata, apparently there are three reserve fuel names:
RESERVE_FUEL_ADDITIONAL, RESERVE_FUEL_FRACTION, RESERVE_FUEL

aviary/examples/run_reserve_mission_cruise_fixedtime.py
```diff
@@ -0,0 +1,154 @@
...
+            "mach_bounds": ((0.7, 0.74), "unitless"),
+            "initial_altitude": (32000.0, "ft"),
+            "final_altitude": (32000.0, "ft"),
+            "altitude_bounds": ((23000.0, 38000.0), "ft"),
+            "throttle_enforcement": "boundary_constraint",
+            "fix_initial": False,
+            "constrain_final": False,
+            "fix_duration": True,
+            "initial_bounds": ((149.5, 448.5), "min"),
+            "duration_bounds": ((60, 60), "min"),  # set this for fixed duration phases
```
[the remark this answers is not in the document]
ehariton: Yeah I re-wrote this with a comment when I changed up the phase_info import. If you look near the top of this file now you'll see it says:
``` 
# Add reserve phase(s)
time_cruise = [60, 'min']  # fixed cruise time
```

aviary/interface/methods_for_level2.py
crecine: RESERVE_FUEL_ADDITIONAL is a fixed amount (e.g. 5000 lbs)
RESERVE_FUEL_FRACTION is a percentage of the design mission fuel (e.g. 5%)
RESERVE_FUEL is the total (currently of just additional and fractional, but should include reserve_mission_fuel when it is available)

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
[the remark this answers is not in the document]
ehariton: The current commit does not support this functionality because:
Reserve time is phase specific,
Reserve range is based on the summation of all reserve phases,

If we need this functionality we will need to redefine how time and range reserves are implemented.

aviary/mission/flops_based/phases/energy_phase.py
ehariton: non-reserve phases don't need to specify `reserve=false`, already implemented :)

aviary/examples/run_reserve_mission_cruise_fixedtime.py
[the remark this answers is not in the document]
crecine: looks good!

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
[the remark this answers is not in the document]
crecine: doesn't have to be a part of this PR, could be a follow on, but I think that is a pretty important feature.

aviary/interface/methods_for_level2.py
ehariton: I don't want to focus on enabling full tandem phases for this PR. That's out of scope for this job.

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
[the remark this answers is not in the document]
ehariton: Looks like this type of requirement can be derived from FAR 121.639 Fuel supply: All domestic operations.  
https://www.ecfr.gov/current/title-14/chapter-I/subchapter-G/part-121/subpart-U/section-121.639

Fly to distant airport THEN loiter at cruise condition for 45 minutes.

aviary/interface/methods_for_level2.py
ehariton: `_link_phases_helper_with_options()` doesn't connect anything if the phase dictionaries are empty.

aviary/examples/run_reserve_mission_multiphase_fixedtime.py
ehariton: I feel like this is out of scope for the current jog. I like making things easier for people, but the focus is on enabling reserve missions of arbitrary ordering.

crecine: for anything outside the scope of this jog/PR, could you add an issue?

aviary/examples/run_reserve_mission_cruise_fixedtime.py
```diff
@@ -111,13 +34,12 @@
             "throttle_enforcement": "boundary_constraint",
             "fix_initial": False,
             "constrain_final": False,
-            "fix_duration": True,
+            "fix_duration": True,  # TBD remove in future
             "initial_bounds": ((149.5, 448.5), "min"),
-            "duration_bounds": ((60, 60), "min"),  # set this for fixed duration phases
+            "duration_bounds": ((45, 45), "min"),  # TBD remove in future
```
crecine, line 39: should target_time and duration_bounds match?

aviary/examples/run_reserve_mission_multiphase_time_and_range.py
```diff
@@ -0,0 +1,179 @@
...
+            "altitude_bounds": ((23000.0, 38000.0), "ft"),
+            "throttle_enforcement": "boundary_constraint",
+            "fix_initial": False,
+            "constrain_final": False,
+            "fix_duration": True,  # TBD remove in future
+            "initial_bounds": ((149.5, 448.5), "min"),
+            "duration_bounds": ((0, 0), 'min'),  # TBD remove in future
+        },
+    },
+    "reserve_cruise_fixed_range_2": {
```
crecine, line 94: I really like how this demonstrates the arbitrary ordering of reserve phases that is now possible.

aviary/interface/methods_for_level2.py
```diff
@@ -487,6 +487,57 @@ def check_and_preprocess_inputs(self):
...
+                if 'target_time' in self.phase_info[phase_name]["user_options"]:
+                    target_time = self.phase_info[phase_name]["user_options"]["target_time"]
+                    if target_time[0] <= 0:
+                        raise ValueError(
+                            f"Invalid target_time in phase_info[{phase_name}][user_options]."
+                            f"Current (value: {target_time[0]}), (units: {target_time[1]}) <= 0")
+                    if 'duration_bounds' in self.phase_info[phase_name]["user_options"]:
+                        # raise ValueError(
+                        print(
+                            f"When specifying target_time, duration_bounds for time should be removed. "
```
crecine, line 519: I think the examples have both set?

aviary/interface/methods_for_level2.py
```diff
@@ -1296,9 +1347,8 @@ def add_post_mission_systems(self, include_landing=True):
                 ],
                 promotes_outputs=[('overall_fuel', Mission.Summary.TOTAL_FUEL_MASS)])
 
-            regular_phases, reserve_phases = self.phase_separator()
-
             # If a target range has been specified
+            # range is measured from the first phase to the last normal phase
```
crecine, line 1351:
```suggestion
            # range is measured from the start of the first phase to the end of the last normal phase
```

aviary/examples/run_reserve_mission_cruise_fixedtime.py
ehariton: `duration_bounds` gets over-written in `check_and_preprocess_inputs()` for `target_time()` phases. We're forced to include something for duration_bounds right now because `check_phase_info()` requires it to be set otherwise it errors out.

aviary/interface/methods_for_level2.py
ehariton: The examples currently trigger these errors because `check_phase_info()` still needs a lot of work, more than I can accomplish this job. That's why the `ValueError()` is commented out, otherwise the examples would crash.

aviary/interface/methods_for_level2.py
ehariton: Mission.Summary.FUEL_BURNED and Mission.Summary.RESERVE_FUEL_BURNED have been added to variable hierarchy and to methods for lvl2. Fuel-flow calcs have been re-done to reflect this.

aviary/examples/external_subsystems/simple_weight/run_simple_weight.py
```diff
@@ -18,6 +18,9 @@
 # Allow for user overrides here
 prob.load_inputs('models/test_aircraft/aircraft_for_bench_FwFm.csv', phase_info)
 
+# Preprocess inputs
+prob.check_and_preprocess_inputs()
```
Kenneth-T-Moore, line 22: good catch; we definitely need to always call this.

aviary/examples/run_reserve_mission_multiphase_time_and_range.py
```diff
@@ -0,0 +1,179 @@
...
+This exposes more options and flexibility to the user and uses the "Level 2" API within Aviary.
+
+We define a `phase_info` object, which tells Aviary how to model the mission.
+Here we have climb, cruise, and descent phases.
+We then call the correct methods in order to set up and run an Aviary optimization problem.
+This performs a coupled design-mission optimization and outputs the results from Aviary into the `reports` folder.
+"""
+import aviary.api as av
+from example_phase_info import phase_info
+
```
Kenneth-T-Moore, line 13: I recommend doing a deepcopy on the phase_info after importing it.

If you don't do this, the phase_info in memory in your process is modified, and will end up with the reserve missions in it. You can end up with a situation where a test that runs after the test and imports example_phase_info could fail because of the extra unexpected phases in it.

aviary/mission/flops_based/phases/energy_phase.py
```diff
@@ -368,6 +370,15 @@ def _extra_ode_init_kwargs(self):
...
 
 
+EnergyPhase._add_meta_data(
+    'reserve', val=False, desc='this phase is part of the reserve mission.')
+
+EnergyPhase._add_meta_data(
+    'target_distance', val={}, desc='the amount of distance traveled in this phase added as a constraint')
+
+EnergyPhase._add_meta_data(
+    'target_time', val={}, desc='the amount of time taken by this phase added as a constraint')
```
Kenneth-T-Moore, line 380: I wonder if "target_duration" is a better name for this, given that we call it duration in the dymos options, and have adopted that in aviary as well.

aviary/interface/methods_for_level2.py
```diff
@@ -450,6 +487,57 @@ def check_and_preprocess_inputs(self):
...
+                            f"Invalid target_distance in [{phase_name}].[user_options]. "
+                            f"Current (value: {target_distance[0]}), (units: {target_distance[1]}) <= 0")
+
+        # Target_time verification for all phases
+        # Checks to make sure target_time is positive,
+        # duration_bounds and initial_guesses for time have not been set,
+        # Then sets duration_bounds, initial_guesses, and fixed_duration
+        for idx, phase_name in enumerate(self.phase_info):
+            if 'user_options' in self.phase_info[phase_name]:
+                if 'target_time' in self.phase_info[phase_name]["user_options"]:
```
Kenneth-T-Moore, line 510: I just noticed that we already have "fix_duration" in the phase info.  I think we can get rid of "target_time", and just require the user to set the "target" time as the "times" in "initial_guesses".

Also, given that fix_duration is already there, you might not have to do any other processing, since something should already be in aviary (though whether it works is a question).

aviary/interface/methods_for_level2.py
```diff
@@ -1394,13 +1533,22 @@ def add_linkage_constraint(self, phase_a, phase_b, var_b):
...
                 phases, ["time", Dynamic.Mission.MASS, Dynamic.Mission.DISTANCE], connected=True)
 
+            # connect regular phases with each other if your are optimizing alt or mach
+            self._link_phases_helper_with_options(
+                self.regular_phases, 'optimize_altitude', Dynamic.Mission.ALTITUDE, ref=1.e4)
+            self._link_phases_helper_with_options(
+                self.regular_phases, 'optimize_mach', Dynamic.Mission.MACH)
+
+            # connect reserve phases with each other if your are optimizing alt or mach
             self._link_phases_helper_with_options(
```
Kenneth-T-Moore, line 1548: Should these be under an "If self.reserve_phases" so that they don't run otherwise?

aviary/interface/methods_for_level2.py
```diff
@@ -1504,24 +1652,19 @@ def add_linkage_constraint(self, phase_a, phase_b, var_a, var_b, connected,
                 self.model.connect(
                     "traj.ascent.timeseries.altitude", "h_fit.h_cp")
 
-                self.model.connect(
-                    "traj.desc2.timeseries.mass",
-                    "landing.mass",
-                    src_indices=[-1],
-                    flat_src_indices=True,
-                )
+                last_flight_phase_name = list(self.phase_info.keys())[-1]
```
Kenneth-T-Moore, line 1655: Isn't this a lot of steps to just get `self.regular_phases[-1].pathname` ?

aviary/examples/external_subsystems/simple_weight/run_simple_weight.py
ehariton: Yes, any height-energy mission that doesn't run `check_and_preprocess_inputs()` will fail now because that's what we use to separate the phases into normal / reserve.

aviary/interface/methods_for_level2.py
```diff
@@ -231,6 +231,9 @@ def __init__(self, analysis_scheme=AnalysisScheme.COLLOCATION, **kwargs):
 
         self.analysis_scheme = analysis_scheme
 
+        self.regular_phases = []
```
Kenneth-T-Moore, line 234: I wonder if "flight_phases" might be better name than "regular_phases".

"Look, Big Daddy! It's regular Daddy! -- Ralph Wiggum

aviary/interface/methods_for_level2.py
```diff
@@ -1316,39 +1316,55 @@ def add_post_mission_systems(self, include_landing=True):
...
             # TODO: the overall_fuel variable is the burned fuel plus the reserve, but should
             # also include the unused fuel, and the hierarchy variable name should be more clear
-            ecomp = om.ExecComp('overall_fuel = fuel_burned + fuel_reserve',
-                                fuel_burned={'units': 'lbm', 'shape': 1},
-                                fuel_reserve={'units': 'lbm', 'shape': 1},
-                                overall_fuel={'units': 'lbm'})
+            ecomp = om.ExecComp('overall_fuel = fuel_burned + reserve_fuel',
+                                overall_fuel={'units': 'lbm', 'shape': 1},
+                                fuel_burned={'units': 'lbm'},  # from regular_phases only
+                                reserve_fuel={'units': 'lbm', 'shape': 1},
```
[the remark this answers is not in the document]
ehariton: I added a better description of this in `variable_meta_data` for `Mission.Design.RESERVE_FUEL`
``` 
desc='the total fuel reserves which is the sum of:'
         'RESERVE_FUEL_BURNED, RESERVE_FUEL_ADDITIONAL, RESERVE_FUEL_FRACTION',
```

aviary/interface/methods_for_level2.py
```diff
@@ -2610,104 +2706,33 @@ def _add_two_dof_landing_systems(self):
...
-            equation_string += f"{reserves_frac}*(takeoff_mass - final_mass)"
-        if reserves_val != 0:
-            equation_string += f" + {reserves_val}"
+    def _add_fuel_reserve_component(self):
+        if self.aviary_inputs.get_val(Aircraft.Design.RESERVE_FUEL_FRACTION, units='unitless') != 0:
+            self.post_mission.add_subsystem(
+                "reserve_fuel_frac",
+                om.ExecComp('reserve_fuel_frac_mass = reserve_fuel_fraction * (takeoff_mass - final_mass)',
+                            reserve_fuel_frac_mass={"units": "lbm"},
+                            reserve_fuel_fraction={"units": "lbm", "val": self.aviary_inputs.get_val(
```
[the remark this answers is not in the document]
ehariton: Change added to latest commit.

aviary/interface/methods_for_level2.py
ehariton: `_link_phases_helper_with_options()` checks if the phases dictionary that you pass it has 1 or less phases and doesn't make any connections if it does. We're only passing reserve_phases or regular_phases, to the function, so reserve and regular can't connect to each other by design.

aviary/interface/methods_for_level2.py
ehariton: in 2DOF we actually don't separate regular_phases and normal_phases yet.

aviary/interface/methods_for_level2.py
ehariton: that's a fair point. as long as we're consistent I'm ok.

crecine: flight_phases kind of implies that it excludes taxi and groundroll, so I think "regular", "nominal", or "primary_mission" would be better

aviary/mission/flops_based/phases/energy_phase.py
ehariton: After a long conversation with Ken, 
target_duration has been implemented instead of target_time.

aviary/interface/methods_for_level2.py
ehariton: we're not going to use fix_duration because it only allows a true/false flag whereas I want to allow the user to directly specify a value and units for that duration in a single place. And if `target_duration` is true, then that also sets the `fix_duration` flag.

aviary/interface/methods_for_level2.py
ehariton: We don't want to exclude taxi / groundroll. So I'm thinking we switch this back to regular_phases.

ehariton: primary_mission seems like longer than needed.

Kenneth-T-Moore (approved):

crecine (approved):
Looking good, just a few minor comments/suggestions

crecine:
Dymos commit today ([69a70ab](https://github.com/OpenMDAO/dymos/commit/69a70ab2a07dc96561a695d47806c62e89f829ab)) is the cause of the test failure, the last successful test used [ed640d7](https://github.com/OpenMDAO/dymos/commit/ed640d7cef5b28770a4d1958a7a9cc0b6d49811f)

aviary/examples/external_subsystems/simple_weight/run_simple_weight.py
ehariton: Current Error Message if you use height_energy / solved_2DOF and don't call check-preprocess:

```
Traceback (most recent call last):
  File "/mnt/c/OMDAO/om-Aviary/aviary/examples/run_reserve_mission_cruise_fixedrange.py", line 61, in <module>
    prob.add_post_mission_systems()
  File "/mnt/c/OMDAO/om-Aviary/aviary/interface/methods_for_level2.py", line 1385, in add_post_mission_systems
    self.model.connect(f"traj.{self.regular_phases[0]}.timeseries.mass",
IndexError: list index out of range
```

aviary/interface/methods_for_level2.py
```diff
@@ -2515,9 +2657,11 @@ def _add_height_energy_landing_systems(self):
         else:
             control_type_string = 'control_values'
 
-        self.model.connect(f'traj.{last_flight_phase_name}.states:mass',
+        last_flight_phase = self.regular_phases[-1]
```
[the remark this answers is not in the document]
ehariton: good point, in this instance yes, this should be changed, in many of the other instances where we use the terminology of `last_flight_phase` it's for non-height-energy or it's mean to look at all the phases and choose the last one.

PR #163, opened 2024-03-06
Some changes for running under MPI.

Kenneth-T-Moore:
### Summary

This is a series of changes to help with running external mission subsystems under MPI.

1. Moved some solvers out of traj.phases.
When dymos is run under MPI, phases is a parallel group. Currently, if any phase has an initial state that is connected to something upstream, a solver is required to update that value into the state via an implicit component. Unfortunately, dymos places that solver in traj.phases to address other corner cases. In this PR, we move that solver into the states independent component so that it is not above the ODE.
2. Severed the connections between phases when running under MPI.
Ideally, you shouldn't parallelize systems that are connected to each other. It will work if you add a nonlinear and linear solver, but this is kind of janky and leads to another level of solvers in your model. We can sever the connection by setting connected=False in the call to link_phases. When we do that, the optimizer handles the removed connections using constraints.
3. Added a few refs for the new constraints.

### Related Issues

- Resolves #

None

### New Dependencies

None


johnjasa (approved):
Great PR, thanks for this Ken. This all seems helpful and useful. I love your comments around your changes in the code that clearly delineate why these changes were needed and what it means for the future!

crecine (approved):

PR #169, opened 2024-03-11
Fix for benchmark regression (hopefully).

Kenneth-T-Moore:
### Summary

1. Fixed a regression that prevented phase time duration refs from being set properly.
2. Added support for setting time bounds and refs on phases where time isn't the integration variable.
- The regression occurred due to an attempt to get around this lack of support.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/methods_for_level2.py
```diff
@@ -836,36 +840,38 @@ def _get_phase(self, phase_name, phase_idx):
                 user_options.set_val('initial_ref', init_bounds[0][1],
                                      units=init_bounds[1])
             else:
-                user_options.set_val('initial_ref', 10., 'min')
+                user_options.set_val('initial_ref', 600., time_units)
```
crecine, line 843: Should we do something to make sure that the ref value is something reasonable?
```
time_val = openmdao.utils.units.convert_units(600, 's', new_units=time_units)
user_options.set_val('initial_ref', time_val, time_units)
```

Kenneth-T-Moore: I really don't know the provenance of this line, or why it is needed. We can't use an actual unit here, because some of our phases have a custom integration variable with different units, and we will get an exception if the ref has units incompatible with the other "time" units.

johnjasa (approved):

crecine (approved):

PR #182, opened 2024-03-15
Using verbosity settings to control print level

crecine:
### Summary

adjusted print level of driver based on verbosity
made BRIEF the default of add_driver and run_aviary
switched tests to use QUIET
	(Examples still use BRIEF)

This makes warnings and errors easier to find and debug

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/docs/developer_guide/coding_standards.md
+++ b/aviary/docs/developer_guide/coding_standards.md
@@ -16,6 +16,9 @@ To set up pre-commit in your development python environment, there are a few one
 
 The Aviary repository contains a configuration file that defines what is run when commits are made and with what options enabled. Currently this is limited to autopep8 with a max line length restriction.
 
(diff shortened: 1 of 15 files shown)


johnjasa (approved):
Awesome, love these changes!

Since Verbosity is now pretty user-facing, could you please add a note to the [Troubleshooting doc page](https://openmdao.github.io/Aviary/user_guide/troubleshooting.html) about these changes/capabilities? Maybe in [Outputs and How to Read Them](https://openmdao.github.io/Aviary/user_guide/outputs_and_how_to_read_them.html) too but that's up to you.

jkirk5 (approved):
I agree with these default settings. If you are able to add the documentation John asked about that would be great, but I will also approve now to avoid holding things up

PR #183, opened 2024-03-17
Update Hamilton Standard model and add integration test

johnjasa:
### Summary

Bevy of changes by myself and @xjjiang that move us towards a complete turboprop modeling capability:
- vectorized the unit test for HS model
- add partials checks to HS unit tests
- set up the infrastructure for and created a turboprop+mission case that runs with SLSQP

Future work will:
- have updated HS model that Xun has worked on with Jeff
- expose the turboprop propulsion setup in a nice way through user input files (without using the custom engine builder)
- add additional phases to the mission integration test

### Related Issues

- Resolves #

### Backwards incompatibilities

None

### New Dependencies

None

PR #184, opened 2024-03-18
Update Hamilton Standard model based on Jeff's enhancement

xjjiang:
### Summary

1. Vectorized Hamilton Standard model.
2. Unit testing for vectorized Hamilton Standard model.
3. Enhanced Hamilton Standard model. The original Hamilton Standard model sometimes fails to work for high blade integrated lift coefficient (`CLI`). A few tables are modified; lower limit on equivalent power coefficient (`CPE`) for interpolation warning; allow 10 iterations (instead of 5) for the Newton-Rhapson method to find the corresponding equivalent thrust coefficient (`CTE`).
4. In GAST, the "Warning" messages are Suppressed for flight idle power calculations but this is not implemented in Aviary.
5. A unit test for high CLI is added. The checking of partial derivative of `'comp_tip_loss_factor'` with respect to `'aircraft:engine:propeller_integrated_lift_coefficient'` is skipped because the `calc mag.` is 0.0 and the `calc mag.` is very close to zero (`1.1102e-10`). As a result, `r(cal-chk)` is always `1.0`.

### Related Issues

- Resolves #

None

### New Dependencies

None


xjjiang:
I will wait for John's [PR #183](https://github.com/OpenMDAO/Aviary/pull/183) to merge into main.

johnjasa (approved):
Good changes, nice catch with the variable metadata fix for prop tip speed!

crecine (approved):

PR #194, opened 2024-03-25
Improvements to variable tests + minor cleanup

jkirk5:
### Summary

- Improvements to variable testing
    - now comparing both hierarchy and metadata with each other (both lists must match)
    - All variables in hierarchy are now included in all variable tests
- Renamed fortran_to_aviary.py for PEP8 compliance
- Added SPECIFIC_ENERGY and MACH_RATE to be consistent with other variables that include both state and state rate in the variable hierarchy
- Added tempdirs flag to some tests to reduce clutter when running tests
- Updates to converters
    - Added timestamp & user heading to fortran_to_aviary
    - Fixed engine type choices not properly rendering when using `-h`/`--help` for engine conversion utility
    - Merged `-v` and `-vv` args, with the expectation that we eventually support QUIET debug mode too, so having unique flags for each will be unwieldly


Kenneth-T-Moore (approved):

johnjasa (approved):

aviary/utils/fortran_to_aviary.py
```diff
@@ -88,15 +97,16 @@ def create_aviary_deck(fortran_deck: str, legacy_code=None, defaults_deck=None,
         out_file = fortran_deck.parent / out_file
 
     if out_file.is_file():
-        if force:
+        if force and verbosity.value >= 1:
```
crecine, line 100: won't this cause an error if verbosity is quiet, even if force is True?

PR #219, opened 2024-03-28
Propeller speed is computed now

xjjiang:
### Summary

In the current design, propeller speed is a user given input. Now, it is computed from two new variables: `MAX_TIP_SPEED` and `PERCENT_ROTOR_RPM_CORRECTED`.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/subsystems/propulsion/prop_performance.py
```diff
@@ -195,6 +178,59 @@ def setup(self):
...
+                ('T0', Dynamic.Mission.TEMPERATURE),
+                ('mach', Dynamic.Mission.MACH),
+            ],
+            promotes_outputs=['theta_T'],
+        )
+
+        self.add_subsystem(
+            'prop_tip_spd',
+            om.ExecComp(
+                'prop_tip_speed = minimum(pc_rotor_rpm_corrected * theta_T**.5 * max_prop_tip_spd, max_prop_tip_spd)',
```
johnjasa, line 201: Do you expect the this minimum function to switch between the expression or the `max_prop_tip_spd` frequently during an optimization? If so, using a [smooth minimum](https://en.wikipedia.org/wiki/Smooth_maximum) function would improve convergence.

Based on the calculations here, I wouldn't expect using a regular minimum function to be a problem in this case. This is more of a general note and something to consider when using gradient-based optimization.

aviary/subsystems/propulsion/test/test_prop_performance.py
```diff
@@ -103,6 +103,10 @@ def test_case_0_1_2(self):
         prob.set_val(Dynamic.Mission.PROPELLER_TIP_SPEED,
                      [800.0, 800.0, 750.0], units="ft/s")
```
johnjasa, line 104: Can you remove all of these `PROPELLER_TIP_SPEED` `set_val` calls in this test now?

johnjasa (approved):
Approving so I don't hold this up! Two quick comments.

aviary/subsystems/propulsion/test/test_prop_performance.py
```diff
@@ -103,6 +103,10 @@ def test_case_0_1_2(self):
...
         prob.set_val(Dynamic.Mission.SHAFT_POWER, [1850.0, 1850.0, 900.0], units="hp")
+        prob.set_val(Dynamic.Mission.PERCENT_ROTOR_RPM_CORRECTED,
+                     [1.0], units="unitless")
```
crecine, line 107: here and below, should this be `[1,1,1]`?

crecine (approved):

aviary/subsystems/propulsion/prop_performance.py
xjjiang: For takeoff, I expect that it is always at maximum tip speed. For cruise, I expect that it is always less than maximum tip speed. I use `maximum()` because it is a built-in function for `ExecComp`. I am not sure if a smooth function can replace it.

aviary/subsystems/propulsion/test/test_prop_performance.py
xjjiang: You are right. I removed it.

aviary/subsystems/propulsion/test/test_prop_performance.py
xjjiang: I guess we can assume that this parameter varies but in GASP, it is a constant in each phase (takeoff, climb, cruise). I changed it to [1,1,1].

crecine: thanks

Issue #157, opened 2024-02-23
Engines require outputting all possible variables


jkirk5:
### Desired capability or behavior.

It may be possible to eliminate this requirement by more intelligent code in the propulsion mission component. Adding try statements (or checking the engine's `self.use_XXXX` flag) to connections instead of assuming they exist, and ensuring that a value of zero is passed to the MuxComp if the variable is not provided by the engine is a possible solution.

### Is your feature request related to a problem? Please describe.

Currently, due to the way all engine outputs are muxed together to compute vehicle-level totals, all engines are required to output all possible engine outputs, even if they are not produced by that engine and get defaulted to zero (e.g. outputting 0 kWh for a fuel turbine). This is a hassle for creating new engine models.

PR #223, opened 2024-04-01
Full turboprop model

jkirk5:
### Summary

Added integrated "propeller propulsor" EngineModel (shaft power model + propeller model). Defaults to an EngineDeck and the Hamilton Standard based propeller model. Users can provide substitute models instead of these defaults.

NOTE: This model has issues related to issue #157. Engine features such as max thrust, reliable triggering of uncorrecting shaft power, and others features will not work properly until the issue is fixed.

Internal to the turboprop model, the promotions and connections will need to be overhauled, which will happen as part of upcoming work for multi-engine capabilities

### Related Issues

- Resolves #

None

### New Dependencies

None


xjjiang:
It looks like you need to take the new changes in main to your branch.

aviary/docs/user_guide/hamilton_standard.md
```diff
@@ -97,7 +97,7 @@ The newly added aviary options and variables are:
 Aircraft.Engine.PROPELLER_DIAMETER
```
ehariton, line 97: I'm seeing a bunch of stuff with PROPELLER as the first part of the name, I think it may be time to make this `.PROPELLER.` i.e. Aircraft.Engine.Propeller.DIAMETER

jkirk5: I also think it makes sense to separate out the propeller analysis, and making it a subsection of engine probably makes the most sense (at least in the short term). Adding a new `Aircraft.Propeller` section will add extra work to the multi-engine effort

ehariton: It looks like there are actually only 8 names that fit in this category:

Aircraft.Engine.PROPELLER_INTEGRATED_LIFT_COEFFICIENT
Aircraft.Engine.PROPELLER_ACTIVITY_FACTOR
Aircraft.Engine.NUM_PROPELLER_BLADES
Aircraft.Engine.PROPELLER_DIAMETER
Aircraft.Design.MAX_PROPELLER_TIP_SPEED
Dynamic.Mission.PROPELLER_TIP_SPEED
Aircraft.Engine.HAS_PROPELLERS
Dynamic.Mission.INSTALLATION_LOSS_FACTOR

aviary/subsystems/propulsion/hamilton_standard.py
```diff
@@ -689,11 +688,11 @@ def compute(self, inputs, outputs):
                             CP_CLi_table[CL_tab_idx][:cli_len], XPCLI[CL_tab_idx], CPE1X)
                         if (run_flag == 1):
                             ichck = ichck + 1
-                        if (self.options[Settings.VERBOSITY] != Verbosity.DEBUG):
+                        if (verbosity != Verbosity.DEBUG):
```
crecine, line 691: should this be `not Quiet` or `is Debug`?

aviary/subsystems/propulsion/propeller_performance.py
```diff
@@ -111,7 +110,7 @@ def setup(self):
         )
 
         self.add_subsystem(
-            name='compute_install_loss_factor',
+            name='install_loss_factor',
```
crecine, line 113:
```suggestion
            name='installation_loss_factor',
```

aviary/subsystems/propulsion/propeller_performance.py
```diff
@@ -276,12 +241,76 @@ def setup(self):
...
             promotes_outputs=[
                 "thrust_coefficient_comp_loss",
-                "prop_thrust",
+                "propeller_thrust",
                 "propeller_efficiency",
                 "install_efficiency",
             ])
+
+
+class UncorrectShaftPower(om.Group):
```
crecine, line 255: I believe this uncorrection will be useful for shaft RPM and thrust, so it might be better to have a more generalized one that can be used.

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -0,0 +1,193 @@
...
+        turboprop_group = om.Group()
+        # TODO engine scaling for turboshafts requires EngineSizing to be refactored to
+        # accept target scaling variable as an option, skipping for now
+        if type(shp_model) is not EngineDeck:
+            shp_model_pre_mission = shp_model.build_pre_mission(aviary_inputs, **kwargs)
+            turboprop_group.add_subsystem(shp_model_pre_mission.name,
+                                          subsys=shp_model_pre_mission,
+                                          promotes=['*'])
+
+        if self.propeller_model is not None:
```
crecine, line 90:
```suggestion
        if prop_model is not None:
```

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -0,0 +1,193 @@
...
+                                                               Dynamic.Mission.MACH],
+                                              promotes_outputs=[Dynamic.Mission.SHAFT_POWER]),
+
+        if prop_model is not None:  # must assume user-provide propeller group has everything it needs
+            prop_model_mission = prop_model.build_mission(
+                num_nodes, self.options, **kwargs)
+            turboprop_group.add_subsystem(prop_model.name,
+                                          subsys=prop_model_mission,
+                                          promotes_inputs=['*'],
+                                          promotes_outputs=['*'])
```
crecine, line 130: Can we assume that user provided models have a build_mission function? Seems like it might be better to expect them to just pass a preconfigured subsystem

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # connect shaftpower model and propeller model to thrust adder
+        # turboprop_group.connect('turboshaft_thrust',
+        #                         'thrust_adder.tailpipe_thrust')
+        # if prop_model is not None:
+        #     turboprop_group.connect(prop_model.name + '.' + 'propeller_thrust',
+        #                             'thrust_adder.propeller_thrust')
+        # else:
+        #     turboprop_group.connect('propeller_model.propeller_thrust',
+        #                             'thrust_adder.propeller_thrust')
```
crecine, line 172: remove?

aviary/subsystems/propulsion/test/test_turboprop_model.py
```diff
@@ -40,14 +43,16 @@ def prepare_model(self, filename, test_points=[(0, 0, 0), (0, 0, 1)], prop_model
...
-        engine = TurboPropDeck(options=options, prop_model=prop_model)
+        num_nodes = len(test_points)
+
+        engine = TurbopropModel(options=options, propeller_model=prop_model)
         preprocess_propulsion(options, [engine])
 
-        self.prob.model = PropulsionMission(
-            num_nodes=len(test_points), aviary_options=options)
+        # self.prob.model = PropulsionMission(
+        #     num_nodes=len(test_points), aviary_options=options)
```
crecine, line 55: remove?

aviary/subsystems/propulsion/test/test_turboprop_model.py
```diff
@@ -185,25 +157,85 @@ def test_case_3(self):
...
+
+        num_nodes = len(test_points)
 
         self.prob.set_val(Aircraft.Engine.PROPELLER_DIAMETER, 10.5, units="ft")
         self.prob.set_val(Aircraft.Engine.PROPELLER_ACTIVITY_FACTOR,
                           114.0, units="unitless")
         self.prob.set_val(
             Aircraft.Engine.PROPELLER_INTEGRATED_LIFT_COEFFICIENT, 0.5, units="unitless")
+        self.prob.set_val(Dynamic.Mission.PROPELLER_TIP_SPEED,
+                          800.*np.ones(num_nodes), units="ft/s")
```
crecine, line 177: This should be a variable based on the test point

aviary/subsystems/propulsion/test/test_turboprop_model.py
```diff
@@ -185,25 +157,85 @@ def test_case_3(self):
...
+                              np.ones(num_nodes), units="degR")
+
+        return prop_group
+
 
 if __name__ == "__main__":
     unittest.main()
+    # test = TurbopropTest()
+    # test.setUp()
+    # test.test_case_1()
```
crecine, line 241: remove

aviary/subsystems/propulsion/engine_scaling.py
```diff
@@ -50,6 +47,12 @@ def setup(self):
         self.add_input('nox_rate_unscaled', val=np.zeros(nn), units='lbm/h',
                        desc='Current NOx emission rate (unscaled)')
 
+        self.add_input('shaft_power_unscaled', val=np.zeros(nn),
+                       units='hp', desc='Current shaft power produced (unscaled)')
+
+        self.add_input('shaft_power_corrected_unscaled', val=np.zeros(nn),
+                       units='hp', desc='Current shaft power produced, corrected (unscaled)')
+
```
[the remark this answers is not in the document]
jkirk5: I think so, but that is going to take a lot of additional work (i.e. a separate PR) - I already have an issue up for this, #157

aviary/subsystems/propulsion/hamilton_standard.py
jkirk5: `not Quiet`, `is Debug`, and the line here (`!= Debug`) are all different - are you asking if we should change the behavior, or rewriting it to use identity comparisons instead of equalities? I double-checked and apparently it is not recommended to use equalities for Enums so I'll change it to `not Debug` unless we want to change behavior.

aviary/subsystems/propulsion/propeller_performance.py
jkirk5: Good idea, I can move it to propulsion's utils.py and have it uncorrect any data you feed it

aviary/subsystems/propulsion/turboprop_model.py
jkirk5: Yeah I think this is safe to remove now

aviary/subsystems/propulsion/test/test_turboprop_model.py
jkirk5: I'm not quite following, can you reword?

aviary/subsystems/propulsion/hamilton_standard.py
crecine: right now it is printing for everything other than debug, and not printing if it is debug.
I was wondering if that was intended, or if it was meant to print only for debug, or for everything other than quiet.

aviary/subsystems/propulsion/test/test_turboprop_model.py
crecine: the propeller_tip_speed should be a dynamic variable that changes with RPM, here (and I think elsewhere) it is treated as a constant (`800.*np.ones(num_nodes)`)

aviary/subsystems/propulsion/utils.py
```diff
@@ -98,6 +104,65 @@ def convert_geopotential_altitude(altitude):
...
+        )
+
+        self.add_subsystem(
+            'uncorrection',
+            om.ExecComp(
+                'data_uncorrected = data_corrected * (delta_T + theta_T**.5)',
+                shaft_power={'units': "hp", 'shape': num_nodes},
+                delta_T={'units': "unitless", 'shape': num_nodes},
+                theta_T={'units': "unitless", 'shape': num_nodes},
+                shaft_power_corrected={'units': "hp", 'shape': num_nodes},
```
crecine, line 159: `shaft_power` and `shaft_power_corrected` need to be changed to `data`
I'd prefer `corrected_data` and `uncorrected_data`, since it is more natural than `data_corrected` and `data_uncorrected`. It will always be alias promoted to some actual variable, so I don't think we need to worry about keeping them together alphabetically.
Either way is fine

aviary/subsystems/propulsion/hamilton_standard.py
jkirk5: This isn't my code so I can't speak to the original intent, but I believe this is specifically intended. There are two variants of this output, which is controlled by multiple factors I don't know the origin of - one "normal" printout for everything up to debug level, and a debug specific variant (the `else` part of that comparison).

aviary/subsystems/propulsion/test/test_turboprop_model.py
jkirk5: 🤷 I just renamed the existing test w/o changing the inputs. While it's possible for RPM to change, does implementing that actually impact this unittest? You'll get different output values, but it won't getting any different code coverage than the fixed speed prop the test originally had.

aviary/subsystems/propulsion/utils.py
jkirk5: Thanks for catching that, I'll update those variables. I'm fine with either order for the data variable names, so `XXX_data` is fine with me

aviary/subsystems/propulsion/hamilton_standard.py
crecine: Thanks, I took another look at this, and it looks like the printouts are the exact same, so this could be simplified to an
`if DEBUG or ichck <= 1` with no else

crecine (approved):
Approving, assuming the execcomp gets fixed.
It would also be nice if that printing section could be condensed

PR #227, opened 2024-04-03
Changed the lower bound of the initial guess for the GTOW to facilita…

errordynamicist:
Set lower limit for gross mass initial guess to allow modeling turboprops under 100k lbf GTOW

### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -1650,7 +1650,7 @@ class AviaryProblem(om.Problem):
             optimize_mass = self.pre_mission_info.get('optimize_mass')
             if optimize_mass:
                 self.model.add_design_var(Mission.Design.GROSS_MASS, units='lbm',
-                                          lower=100.e3, upper=200.e3, ref=135.e3)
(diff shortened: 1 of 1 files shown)


johnjasa (approved):

aviary/interface/methods_for_level2.py
```diff
@@ -1650,7 +1650,7 @@ def add_design_variables(self):
             optimize_mass = self.pre_mission_info.get('optimize_mass')
             if optimize_mass:
                 self.model.add_design_var(Mission.Design.GROSS_MASS, units='lbm',
-                                          lower=100.e3, upper=200.e3, ref=135.e3)
+                                          lower=100.e2, upper=200.e3, ref=135.e3)
```
Kenneth-T-Moore, line 1653: I think eventually we want to allow the user to set these through the interface, though we probably would still have default bounds and refs, and these are better values.

(Note: IPOPT likes tighter bounds in general.)

Kenneth-T-Moore (approved):

PR #246, opened 2024-04-19
Converted LTF file

errordynamicist:
### Summary

Converted the GASP LTF deck to an Aviary .csv file. 

### Related Issues

- Resolves #

None

### New Dependencies

None


crecine (approved):

aviary/models/large_turboprop_freighter/LTF.csv
jkirk5, line 1: Can you mirror the filename here? So it would become `large_turboprop_freighter.csv`

aviary/models/large_turboprop_freighter/LTF.csv
```diff
@@ -0,0 +1,238 @@
+# Input Values
```
jkirk5, line 1: Can you add back the header data that contains the date/time of conversion and the original filename? That info is occasionally useful. Stripping out username is ok, I do that for all converted files in the repo

jkirk5 (changes requested):
Looks good, just need some minor changes for consistency with the other converted input files in the repo

aviary/models/large_turboprop_freighter/LTF.csv
crecine: the section in create_aviary_deck that gets the time and user, never writes the comments to the output file

crecine: I've added a fix for this to #248

jkirk5 (approved):

PR #248, opened 2024-04-22
Turboprop fix

crecine:
### Summary

Fixed some issues on the vehicle side when running a mission with a turboprop aircraft.
There are still some issues with engine vectorization.
This builds off of #246 

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/methods_for_level2.py
```diff
@@ -381,8 +382,12 @@ def load_inputs(self, aviary_inputs, phase_info=None, engine_builder=None, verbo
         self.ode_args = dict(aviary_options=aviary_inputs,
                              core_subsystems=default_mission_subsystems)
 
+        self.engine_type = aviary_inputs.get_val(Aircraft.Engine.TYPE)
```
jkirk5, line 385: Engine type is a GASP mass-only variable that is getting made required here and will break all non-GASP mass aircraft.

The GASPEngineType enum has a ton of types that don't do anything at this level (a user will probably get confused why turboprop adds on the prop automatically, but turboshaft or any of the recip engines don't

aviary/variable_info/variable_meta_data.py
```diff
@@ -2099,6 +2099,18 @@
...
 
+add_meta_data(
+    Aircraft.Engine.SLS_SHAFT_POWER_MAX,
+    meta_data=_MetaData,
+    historical_name={"GASP": 'INPROP.HPMSLS',
+                     "FLOPS": None,
+                     "LEAPS1": None
+                     },
+    units='hp',
+    desc='maximum shaft power of a turboprop',
```
jkirk5, line 2110:
```suggestion
    desc='maximum shaft power of an engine',
```
anything from a turboshaft, recip engine, to even a black-box motor model can use this variable

jkirk5:
Closing this draft - many changes have been incorporated into other PRs, and I'd prefer to try a different approach for getting a turboprop model setup using L1 interface

PR #265, opened 2024-05-07
Import statement using absolute path

xjjiang:
### Summary

In some Aviary source files, some import statements for Aviary modules are relative. Some of these source files are runnable as standalone scripts - which typically do not work well with relative import statements. For this reason, I converted all relative paths to absolute paths. I also added a guideline in developer guide.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):
I am a big fan of absolute path imports.

PR #269, opened 2024-05-09
When using aviary level1, create a subdirectory within reports using the input deck name

xjjiang:
### Summary

The issue is: when running the aviary command with level1, the reports that are generated go into the ./reports/aviary directory. The name of the input deck is not used. So when running different input decks, the results overwrite previous results from different decks. This PR fixes the problem.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/methods_for_level1.py
```diff
@@ -185,6 +185,9 @@ def _exec_level1(args, user_args):
     if isinstance(args.input_deck, list):
         args.input_deck = args.input_deck[0]
 
+    file_name_stem = Path(args.input_deck).stem
+    args.outdir = args.outdir + os.sep + file_name_stem
```
johnjasa, line 189: Does this always overwrite an existing `outdir` if the user supplies it? @crecine does this change your expected behavior?

xjjiang: You are right. I should consider the scenario that the user provide a special `outdir`. Now, I added a logic that check whether `outdir` is the default directory. If it is, add input deck name to it.

aviary/interface/methods_for_level1.py
```diff
@@ -185,6 +185,12 @@ def _exec_level1(args, user_args):
     if isinstance(args.input_deck, list):
         args.input_deck = args.input_deck[0]
 
+    curr_dir = os.path.abspath(os.getcwd())
+    if args.outdir == curr_dir + os.sep + 'output':
```
crecine, line 189:
```suggestion
    if args.outdir == os.path.join(os.getcwd(), "output"):
```
it is probably safer to use the same command as before for the comparison, just in case one file path is relative and the other is absolute, but point to the same directory.

crecine (approved):
one suggestion, but other than that, looks good.

johnjasa (approved):

PR #276, opened 2024-05-14
In LowSpeedAero, alpha is NOT an output

xjjiang:
### Summary

In `LowSpeedAero` (GASP based), alpha is NOT an output even though the code indicates that `output_alpha` is `True`. This was a comment and is replaced by a warning message. Currently, this option (`output_alpha == True`) is not used. If it is needed, GASP has a subroutine to compute alpha. We can implement it in Aviary.

### Related Issues

- Resolves [deprecated Aviary #343](https://github.com/OpenMDAO/Aviary_deprecated/issues/343)

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/aerodynamics/gasp_based/gaspaero.py
+++ b/aviary/subsystems/aerodynamics/gasp_based/gaspaero.py
@@ -1,4 +1,6 @@
 import numpy as np
+import warnings
+
 import openmdao.api as om
 from openmdao.utils import cs_safe as cs
 
@@ -1371,7 +1373,7 @@ class LowSpeedAero(om.Group):
                 # so ensure this is what's passed to DragCoef
                 promotes_outputs=[("CL", "CL_full_flaps")],
             )
-            # Note: alpha is not an output
+            warnings.warn("Alpha is NOT an output from LowSpeedAero.")
         else:
             # TODO this makes "output_alpha" a bit of a misnomer, would computing an
             # alpha be useful? an issue is ground effects depend on alpha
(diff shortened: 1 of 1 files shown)


johnjasa (approved):

Kenneth-T-Moore (approved):

aviary/subsystems/aerodynamics/gasp_based/gaspaero.py
```diff
@@ -1389,10 +1391,8 @@ def setup(self):
                 # so ensure this is what's passed to DragCoef
                 promotes_outputs=[("CL", "CL_full_flaps")],
             )
-            # Note: alpha is not an output
+            warnings.warn("Alpha is NOT an output from LowSpeedAero.")
```
crecine, line 1394: @johnjasa and @Kenneth-T-Moore
could we add a balance comp here (or in CLFromLift) so that this does actually output alpha?
image

Kenneth-T-Moore: @crecine : I am not familiar enough with Gasp aero, but I think that the issue is that the component that computes alpha from the lift_curve_slope, but we only do this for the clean aero, and this is low speed aero.  I'm not sure if that calculation is valid in low speed.

What you suggest about the balance comp would require a solver, and it would have to actually do something if we varied alpha (and we would have to have the right independents to watch.)  Probably way beyond this story.

crecine: Ahh, that makes sense. Thanks

PR #280, opened 2024-05-15
Fixes to allow use of external engine models & multiple engines using L2 interface

jkirk5:
NOTE: This PR builds on #260 and requires that PR to be merged first
### Summary
The old `engine_builder` parameter in methods for L2 was storing engines provided there with the external subsystems to allow for methods such as `get_controls()`, `get_parameters()`, etc. to be called. This broke with recent propulsion changes, and did not support multiple engine types. This PR makes the following changes:
- Engine model list was moved from undocumented `engine_models` variable in AviaryValues to an attribute in propulsion builder
- This allows the `engine_builders` parameter in `AviaryProblem` to now correctly adds engine models to the propulsion subsystem
- In turn this enables the propulsion subsystem builder to collect method calls such as `get_controls()` for all provided engines
- The default subsystems created in default phase info files was removed (as we now needs engine information to build propulsion subsystem). New helper functions were created to help set up tests
- All tests were updated with interface changes

Known issues, out of scope for now:
- The method of collecting all the parameters, controls, bus vars, etc. can break in edge cases of multiengine, where if two engines offer the same variable but with different values or options, only one will be kept
- build_engine_deck() only creates a single engine deck object from the provided options - in the future this should support building multiple engines if provided options are vectorized


aviary/interface/methods_for_level2.py
```diff
@@ -328,76 +328,17 @@ def load_inputs(self, aviary_inputs, phase_info=None, engine_builder=None, verbo
...
+        self.engine_builders = engine_builders
 
         self.aviary_inputs = aviary_inputs
         return aviary_inputs
 
     def _update_metadata_from_subsystems(self):
         self.meta_data = BaseMetaData.copy()
 
-        variables_to_pop = []
+        # variables_to_pop = []
```
Kenneth-T-Moore, line 341: I think this is a line I forgot to remove.

aviary/mission/flight_phase_builder.py
```diff
@@ -11,6 +11,7 @@
 from aviary.variable_info.variables import Aircraft, Dynamic
 from aviary.mission.flops_based.ode.mission_ODE import MissionODE
 from aviary.variable_info.enums import EquationsOfMotion, ThrottleAllocation
+from aviary.variable_info.variables import Aircraft
```
Kenneth-T-Moore, line 14: Doesn't look like we need this.

aviary/mission/flops_based/ode/mission_ODE.py
```diff
@@ -67,7 +67,7 @@ def setup(self):
         aviary_options = options['aviary_options']
         core_subsystems = options['core_subsystems']
         subsystem_options = options['subsystem_options']
-        num_engine_type = len(aviary_options.get_val('engine_models'))
+        engine_count = len(aviary_options.get_val(Aircraft.Engine.NUM_ENGINES))
```
Kenneth-T-Moore, line 70: The name here relates to that whole discussion about number of hetergenous engines, etc. I guess it is best addressed in a future PR.

aviary/mission/flops_based/ode/test/test_landing_ode.py
```diff
@@ -13,11 +14,13 @@
 class FlareODETest(unittest.TestCase):
     def test_case(self):
         prob = om.Problem()
-
         time, _ = detailed_landing_flare.get_item('time')
         nn = len(time)
         aviary_options = inputs
 
+        default_mission_subsystems = get_default_mission_subsystems(
```
Kenneth-T-Moore, line 21: We kind of don't really even need these lines here, by the look of it. the old default_mission_subsystems as an unused import.

aviary/mission/flops_based/ode/test/test_takeoff_ode.py
```diff
@@ -70,6 +71,9 @@ def _make_prob(climbing):
         nn = len(time)
         aviary_options = inputs
 
+        default_mission_subsystems = get_default_mission_subsystems(
```
Kenneth-T-Moore, line 74: same here

aviary/mission/gasp_based/ode/unsteady_solved/test/test_unsteady_alpha_thrust_iter_group.py
```diff
@@ -5,22 +5,28 @@
...
+from aviary.subsystems.aerodynamics.aerodynamics_builder import CoreAerodynamicsBuilder
+from aviary.variable_info.enums import LegacyCode
 
 
 class TestUnsteadyAlphaThrustIterGroup(unittest.TestCase):
 
     def _test_unsteady_alpha_thrust_iter_group(self, ground_roll=False):
         nn = 5
 
+        # just need aero subsystem
```
Kenneth-T-Moore, line 27: I don't think this or any of the new imports are needed here.

Note: we do have to test that CoreAerodynmicsBuilder runs when called, but that would belong in a unit test, and the return should be checked.

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1010,24 +1008,30 @@ def build_mission(self, num_nodes, aviary_inputs) -> om.Group:
...
         outputs = [Aircraft.Engine.NUM_ENGINES,
                    Aircraft.Engine.SCALED_SLS_THRUST,
                    Aircraft.Engine.SCALE_FACTOR]
 
         # determine which index in problem-level aviary values corresponds to this engine
-        engine_idx = None
-        for idx, engine in enumerate(problem.aviary_inputs.get_val('engine_models')):
-            if engine.name == self.name:
-                engine_idx = idx
+        # engine_idx = None
```
Kenneth-T-Moore, line 1025: You an remove these commented out lines now that the engine number is passed into this report.

aviary/subsystems/propulsion/propulsion_mission.py
```diff
@@ -16,27 +16,30 @@ class PropulsionMission(om.Group):
...
-        # TODO what if "engine" is not an EngineModel object? Type is never checked/enforced
-
         if num_engine_type > 1:
 
             # We need a single component with scale_factor. Dymos can't find it when it is
             # already sliced across several component.
+            # TODO this only works for engine decks. Need to fix problem in generic way
             comp = om.ExecComp(
                 "y=x",
                 y={'val': np.ones(num_engine_type), 'units': 'unitless'},
```
Kenneth-T-Moore, line 45: While we are looking here, please add `has_diag_partials=True` so the execomp arguments.

aviary/subsystems/propulsion/utils.py
```diff
@@ -104,6 +107,54 @@ def convert_geopotential_altitude(altitude):
...
+    for entry in Aircraft.Engine.__dict__:
+        var = getattr(Aircraft.Engine, entry)
+        # check if this variable exist with useable metadata
+        try:
+            units = _MetaData[var]['units']
+            try:
+                # add value from aviary_options to engine_options
+                aviary_val = aviary_options.get_val(var, units)
+                if isinstance(aviary_val, np.ndarray):
+                    # "Convert" numpy types to standard Python types. Wrap first
```
Kenneth-T-Moore, line 140: Are we ever going to have an engine model that has a true numpy vector has one of the input/options? if so, then grabbing the 0th element won't be the right thing to do.

aviary/subsystems/propulsion/utils.py
```diff
@@ -104,6 +107,54 @@ def convert_geopotential_altitude(altitude):
...
+                    # non-standard types, such as objects
+                    aviary_val = aviary_val[0].item()
+                elif isinstance(aviary_val, (list, tuple)):
+                    aviary_val = aviary_val[0]
+                engine_options.set_val(var, aviary_val, units)
+            # if not, use default value from _MetaData
+            except KeyError:
+                # engine_options.set_val(var, _MetaData[var]['default_value'], units)
+                continue
+        except (KeyError, TypeError):
```
Kenneth-T-Moore, line 151: A TODO, but it seems like we do this whole operation (loop over a dict, get units from metadata dict, get value + units from aviary inputs if it is in there) a few places in the code, so we should unify if we can.

aviary/subsystems/propulsion/propulsion_builder.py
```diff
@@ -41,34 +46,195 @@ def mission_outputs(self, **kwargs):
...
+                #      engine locations
+                params[var]['shape'] = (num_engine_type,)
+                params[var]['static_target'] = True
 
-        params = {}  # For now
-        params[Aircraft.Engine.SCALE_FACTOR] = {'shape': (num_engine_type, ),
-                                                'static_target': True}
+        # params = {}  # For now
+        # params[Aircraft.Engine.SCALE_FACTOR] = {'shape': (engine_count, ),
+        #                                         'static_target': True}
```
crecine, line 122: remove?

aviary/subsystems/propulsion/test/test_custom_engine_model.py
```diff
@@ -309,3 +307,5 @@ def test_turboprop(self):
 
 if __name__ == '__main__':
     unittest.main()
+    # test = CustomEngineTest()
+    # test.test_custom_engine()
```
crecine, line 311: remove

aviary/subsystems/propulsion/test/test_propulsion_mission.py
```diff
@@ -209,7 +211,7 @@ def test_case_multiengine(self):
...
 
 if __name__ == "__main__":
-    # unittest.main()
-    test = PropulsionMissionTest()
-    test.setUp()
-    test.test_case_multiengine()
+    unittest.main()
+    # test = PropulsionMissionTest()
+    # test.setUp()
+    # test.test_case_multiengine()
```
crecine, line 217: remove

aviary/subsystems/propulsion/test/test_propulsion_premission.py
```diff
@@ -89,7 +96,7 @@ def test_propulsion_sum(self):
...
 
 if __name__ == "__main__":
-    # unittest.main()
-    test = PropulsionPreMissionTest()
-    test.setUp()
-    test.test_multi_engine()
+    unittest.main()
+    # test = PropulsionPreMissionTest()
+    # test.setUp()
+    # test.test_case()
```
crecine, line 102: remove

aviary/subsystems/test/test_premission.py
```diff
@@ -351,3 +355,6 @@ def test_manual_override(self):
 
 if __name__ == "__main__":
     unittest.main()
+    # test = PreMissionTestCase()
+    # test.setUp()
+    # test.test_GASP_mass_FLOPS_everything_else()
```
crecine, line 360: remove

aviary/utils/preprocessors.py
```diff
@@ -303,16 +256,23 @@ def preprocess_propulsion(aviary_options: AviaryValues, engine_models: list = No
...
+            # more defined engine locations than number of engines - increase num engines
             eng_name = engine.name
             num_engines_all[i] = total_engines_calc
             warnings.warn(
                 'Sum of aircraft:engine:num_fueslage_engines and '
                 'aircraft:engine:num_wing_engines do not match '
                 f'aircraft:engine:num_engines for EngineModel <{eng_name}>. Overwriting '
                 'with the sum of wing and fuselage mounted engines.')
+        elif total_engines_calc < num_engines:
+            # fewer defined locations than num_engines - assume rest are wing mounted
```
crecine, line 270: if fewer engine locations are specified than there are engines, wouldn't it make more sense to assume that the unspecified ones are on the fuselage?

aviary/utils/test_utils/default_subsystems.py
```diff
@@ -0,0 +1,24 @@
+from aviary.subsystems.propulsion.propulsion_builder import CorePropulsionBuilder
+from aviary.subsystems.geometry.geometry_builder import CoreGeometryBuilder
+from aviary.subsystems.mass.mass_builder import CoreMassBuilder
+from aviary.subsystems.aerodynamics.aerodynamics_builder import CoreAerodynamicsBuilder
+from aviary.variable_info.variable_meta_data import _MetaData as BaseMetaData
+from aviary.variable_info.enums import LegacyCode
+
+
+def get_default_premission_subsystems(legacy_code=None, engines=None):
+    legacy_code = LegacyCode(legacy_code)
```
crecine, line 10: Does `LegacyCode(None)` cause problems? If so, should we have a better default or just no default?

aviary/validation_cases/benchmark_tests/test_bench_multiengine.py
```diff
@@ -111,7 +120,7 @@ def test_multiengine_static(self):
         alloc_descent = prob.get_val('traj.descent.parameter_vals:throttle_allocations')
 
         assert_near_equal(alloc_climb[0], 0.5, tolerance=1e-2)
-        assert_near_equal(alloc_cruise[0], 0.56, tolerance=1e-2)
+        assert_near_equal(alloc_cruise[0], 0.64, tolerance=1e-2)
```
crecine, line 123: why did this change?

aviary/mission/flight_phase_builder.py
jkirk5: gets uses on line 77

aviary/mission/flops_based/ode/mission_ODE.py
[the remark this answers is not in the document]
jkirk5: probably a git merge thing that happened here since both 260 and 280 changed this line :/

aviary/mission/flops_based/phases/test/test_time_integration_phases.py
```diff
@@ -35,8 +35,12 @@ def setUp(self):
                               val=0.35, units="unitless")
         aviary_inputs.set_val(Settings.EQUATIONS_OF_MOTION,
                               val=EquationsOfMotion.SOLVED_2DOF)
-        ode_args = dict(aviary_options=aviary_inputs, core_subsystems=[prop, geom, aero])
-        preprocess_propulsion(aviary_inputs, [EngineDeck(options=aviary_inputs)])
+
+        engines = build_engine_deck(aviary_inputs)
+        # don't need mass
+        core_subsystems = get_default_premission_subsystems('FLOPS', engines)[:-1]
```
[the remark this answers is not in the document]
jkirk5: We don't need to run mass premission to test individual mission segments here since we are directly defining start of phase mass

aviary/subsystems/propulsion/utils.py
jkirk5: ...I forgot to wrap the 0th index in a numpy array before calling `item()`, even though the comment says to.
That will work for any arbitrary numpy array of any dtype. I think your comment was targeting something else though

aviary/subsystems/propulsion/utils.py
```diff
@@ -104,6 +107,54 @@ def convert_geopotential_altitude(altitude):
...
+                    # "Convert" numpy types to standard Python types. Wrap first
+                    # index in numpy array before calling item() to safeguard against
+                    # non-standard types, such as objects
+                    aviary_val = aviary_val[0].item()
+                elif isinstance(aviary_val, (list, tuple)):
+                    aviary_val = aviary_val[0]
+                engine_options.set_val(var, aviary_val, units)
+            # if not, use default value from _MetaData
+            except KeyError:
+                # engine_options.set_val(var, _MetaData[var]['default_value'], units)
```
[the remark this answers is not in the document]
jkirk5: I'm still not sure if we want this or not - right now it isn't needed because all our tests are well-defined, but we might want to default variables users don't input for engine decks. However, it's equally reasonable to not automatically add in variables the user didn't specify, since there's a chance that has unforseen consequences.

Right now the setup is if you need a var but didn't provide it things break, rather than using a default. Is that a good idea? I'm not sure

aviary/subsystems/propulsion/utils.py
jkirk5: Yeah it pops up a few times. I'm not sure it's 100% the same each time but we could probably refactor around the common core there. I'll make an issue

aviary/mission/flops_based/phases/test/test_time_integration_phases.py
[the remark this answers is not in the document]
crecine: ahh, that makes sense. Thanks

aviary/mission/gasp_based/ode/unsteady_solved/test/test_unsteady_alpha_thrust_iter_group.py
```diff
@@ -5,22 +5,28 @@
...
 from aviary.constants import GRAV_ENGLISH_LBM
-from aviary.interface.default_phase_info.two_dof import aero
 from aviary.mission.gasp_based.ode.params import ParamPort
 from aviary.mission.gasp_based.ode.unsteady_solved.unsteady_control_iter_group import \
     UnsteadyControlIterGroup
 from aviary.mission.gasp_based.ode.unsteady_solved.unsteady_solved_flight_conditions import \
     UnsteadySolvedFlightConditions
 from aviary.variable_info.enums import SpeedType
-from aviary.variable_info.variables import Aircraft, Dynamic
+from aviary.variable_info.options import get_option_defaults
```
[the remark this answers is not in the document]
crecine: nevermind, duplicate with line 15

aviary/subsystems/propulsion/utils.py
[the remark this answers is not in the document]
crecine: In general I like having defaults, like having the scale factors default to 1, to reduce workload on users, especially if we want this to be accessible to new users/users of the legacy codes.

Maybe we can set default values for some (but not all) of the engine related variables, and use the set_val only if there is some default value.

aviary/utils/preprocessors.py
jkirk5: My logic is that I think the most common scenario this will happen is when users set up their engines but forget to specify location. Since the majority of aircraft Aviary's methods are designed for have wing-mounted engines I chose that as the default. Since fuselage-mounted engines are typically the exception to the norm I'd expect users to be more likely to remember to specify mount location. Even more so for a mixed-mount engine setup.

aviary/variable_info/variable_meta_data.py
```diff
@@ -1666,7 +1666,7 @@
                      "LEAPS1": None
                      },
     units='unitless',
-    types=(str, Path, None),
```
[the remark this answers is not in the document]
jkirk5: I'm not sure the best way to handle it, but I don't want there to be an actual "default" data file that gets used if the user doesn't specify. There has to be a `default_value` because of the way Aviary is set up so I put None. But None isn't actually a acceptable type for `DATA_FILE`, if in your csv you put `DATA_FILE = None` you should get told that's not valid. An error's going to get triggered somewhere along the way, the question is where that would be the most helpful to the user to occur.

aviary/validation_cases/benchmark_tests/test_bench_multiengine.py
jkirk5: Engine sizing is working now! The engines are actually sized now, so instead of the initial guess of 0.5 scale factor for each engine, they get optimized and end up with one around ~0.35 and the other ~0.65

aviary/interface/test/test_phase_info.py
```diff
@@ -156,6 +156,6 @@ def test_phase_info_parameterization_height_energy(self):
 
 # To run the tests
 if __name__ == '__main__':
-    # unittest.main()
-    test = TestParameterizePhaseInfo()
-    test.test_phase_info_parameterization_two_dof()
+    unittest.main()
+    # test = TestParameterizePhaseInfo()
+    # test.test_phase_info_parameterization_two_dof()
```
crecine, line 161: remove

aviary/mission/gasp_based/ode/unsteady_solved/test/test_unsteady_alpha_thrust_iter_group.py
```diff
@@ -14,7 +14,6 @@
 from aviary.variable_info.options import get_option_defaults
 from variable_info.options import get_option_defaults
```
crecine, line 15:
```suggestion
```

aviary/validation_cases/benchmark_tests/test_bench_multiengine.py
```diff
@@ -169,6 +169,6 @@ def test_multiengine_dynamic(self):
 
 
 if __name__ == '__main__':
-    unittest.main()
-    # test = MultiengineTestcase()
-    # test.test_multiengine_static()
+    # unittest.main()
+    test = MultiengineTestcase()
+    test.test_multiengine_static()
```
crecine, line 174: revert/remove when done

crecine (approved):

aviary/utils/preprocessors.py
crecine: but if they haven't specified a location, what do we use as the value? it seems like the only time someone wouldn't bother specifying a wing location is if it isn't on the wing.
Maybe this should be an error then.

aviary/utils/test_utils/default_subsystems.py
crecine: what about `get_default_premission_subsystems(legacy_code, engines=None)`, so it's clear they need to specify something

crecine (commented):
a couple more comments on the preprocessor and default_subsystems

Kenneth-T-Moore (approved):

aviary/utils/preprocessors.py
jkirk5: Inherited from FLOPS, we have `NUM_ENGINES`, `NUM_WING_ENGINES`, and `NUM_FUSELAGE_ENGINES` - `NUM_ENGINES` is supposed to be the sum of wing and fuselage engines. I think it will be pretty common for users to incorrectly set up these values, either forgetting to specify one or more, or provide conflicting values. I set up catches for as many incorrect combinations as I could that are "salvageable", this one is for when the user only provided `NUM_ENGINES` and forgot to specify anything else. FLOPS defaults to wing-mounted in most cases, so I followed that logic.

aviary/subsystems/propulsion/utils.py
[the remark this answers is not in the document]
crecine: just noticed that the comments I left weren't marked as part of my latest review

aviary/utils/preprocessors.py
crecine: Do the span locations impact the wing weight calculation like they do with GASP? If we default to adding these as wing_engines, how do we decide where on the wing they go?

PR #282, opened 2024-05-15
Initial guesses initialization

cmbenne3:
### Summary

Changes the location of the initialization of 'initial_guesses'. This fixes an error with load_inputs() when called with an AviaryValues object, rather than a .csv input file.

Change to parse_inputs() to allow for optional arguments, matching best practice elsewhere in aviary codebase.

None

### New Dependencies

None


aviary/utils/process_input_decks.py
```diff
@@ -135,9 +142,15 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
 
             elif var_name in guess_names:
                 # all initial guesses take only a single value
+                # get values from supplied dictionary
                 initial_guesses[var_name] = float(var_values[0])
                 continue
 
+            elif var_name.startswith('initial_guesses:'):
+                # get values labelled as initial_guesses in .csv input file
+                initial_guesses[var_name[len('initial_guesses:'):]] = float(var_values[0])
```
crecine, line 151: I think you are trying to get the part of the string after `initial_guesses:`, right?
If so, using a `var_name.split(':')[-1]` is probably more clear

cmbenne3: Yes - that was the idea. 
I'm not really sure if this is necessary at all. 
The .csv input files currently has the initial_guesses listed in this way:
image

I was wondering if it would be helpful for someone running 'parse_inputs' manually to be able to specify the initial_guesses in the input file without also needing to send the correct dictionary to the parse_inputs function.

I can either remove this section completely or change it as you suggest above.

crecine: I think this is a good addition, but should be changed for clarity

cmbenne3: @crecine - I've made and committed your suggested change, tested it, and pushed to my fork... it seems that there's now a merge conflict - possibly just something to do with formatting... please let me know what I should do to correct it!

aviary/utils/process_input_decks.py
```diff
@@ -59,18 +59,29 @@ def create_vehicle(vehicle_deck='', verbosity=Verbosity.BRIEF):
...
     aircraft_values.set_val(Settings.PROBLEM_TYPE, val=ProblemType.SIZING)
     aircraft_values.set_val(Aircraft.Electrical.HAS_HYBRID_SYSTEM, val=False)
-
+    
+    initial_guesses = {
+        # initial_guesses is a dictionary that contains values used to initialize the trajectory
+        'actual_takeoff_mass': 0,
+        'rotation_mass': .99,
+        'operating_empty_mass': 0,
+        'fuel_burn_per_passenger_mile': 0.1,
```
crecine, line 68: something I've been meaning to change, but haven't gotten around to, is changing the default values for rotation_mass and fuel_burn_per_passenger_mile to 0, and then adding something like:
```
if rotation_mass == 0:
	rotation_mass = .99
```
to be consistent with the rest of the variables.
Would you be ok with adding that in here as well?

aviary/utils/process_input_decks.py
crecine: resolved your merge conflict for you

aviary/utils/process_input_decks.py
cmbenne3: I've done this within the same 'create_vehicle' method. I wondered about putting it in 'initial_guessing', but this isn't always called... other option would be to put it at the end of the create_vehicle method, but then if a user has specified a zero, we'd be overwriting that with a default, which I think would be undesirable behavior... (I'm not sure what happens if the initial guess for rotation_mass or fuel_burn_per_passenger is zero, but i'd want it to error if I manually sent it a bad value, rather than overwriting my input with a default)

As an aside it looks like this variable should actually be the absolute mass of the aircraft on rotation (in lbs):
image

It seems we're setting the default to 0.99lbs, which seems unnecessary - could we just use 0?

crecine: sorry, I meant to say that this should be initial_guessing with the rest of them.

In initial_guessing, this will be converted to a mass in pounds
image

I'm envisioning that rotation_mass would be handled similar to cruise_mass_final
image

cmbenne3: ok gotcha - I've updated this now. initial_guessing is only called:
image
So if those conditions aren't met we'll be defaulting the initial_guesses to 0 (which is a change in behavior to what is currently in the code)

crecine (approved):

aviary/utils/process_input_decks.py
crecine: As far as I'm aware, the initial guesses aren't used by the height energy missions or the FLOPS mass calculations, so this shouldn't cause any problems

aviary/utils/process_input_decks.py
```diff
@@ -60,18 +60,30 @@ def create_vehicle(vehicle_deck='', verbosity=Verbosity.BRIEF, meta_data=_MetaDa
...
+    }
+
     if isinstance(vehicle_deck, AviaryValues):
         aircraft_values.update(vehicle_deck)
-        initial_guesses = {}
     else:
         vehicle_deck = get_path(vehicle_deck)
         aircraft_values, initial_guesses = parse_inputs(
-            vehicle_deck, aircraft_values, meta_data=meta_data)
+            vehicle_deck, aircraft_values, initial_guesses, meta_data=meta_data)
```
jdgratz10, line 81: now that this function call is getting more complicated could we add variable names? Ie. `parse_inputs(vehicle_deck=vehicle_deck, aircraft_values=aircraft_values`...?

jdgratz10 (changes requested):
Two cosmetic changes requested, otherwise looks good!

aviary/utils/process_input_decks.py
```diff
@@ -60,18 +60,30 @@ def create_vehicle(vehicle_deck='', verbosity=Verbosity.BRIEF, meta_data=_MetaDa
...
 
     return aircraft_values, initial_guesses
 
 
-def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaData):
+def parse_inputs(vehicle_deck, aircraft_values: AviaryValues = None, initial_guesses=None, meta_data=_MetaData):
```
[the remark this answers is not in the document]
cmbenne3: yes!

aviary/utils/process_input_decks.py
cmbenne3: I've updated this as suggested.

aviary/utils/process_input_decks.py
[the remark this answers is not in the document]
cmbenne3: Actually 'no!' I believe Autopep8 is adding the spaces back in which is why my change wasn't visible!
I've checked and the spaces in "AviaryValues = None" are consistent with all other instances of that in the codebase

crecine:
could you update the default values in aviary/docs/getting_started/input_csv_phase_info.md as well

aviary/docs/getting_started/input_csv_phase_info.md
```diff
@@ -6,18 +6,24 @@
...
 - flight_duration: 0,
 - time_to_climb: 0,
 - climb_range: 0,
 - reserves: 0
 
+In the example vehicle input .csv files there is a section headed '# Initial Guesses' which contains the above list in the format: "actual_takeoff_mass,0"
+The user can also specify these parameters with the prefix 'initial_guesses:'
+e.g. 'initial_guesses:actual_takeoff_mass,150000' would set actual_takeoff_mass in the initial_guesses dictionary to 150000
+
+If mission_method is TWO_DEGREES_OF_FREEDOM or mass_method is GASP then the initial_guessing() method is called and wherever the initial_guesses values are equal to 0, they are updated with calculated estimates based off the problem type (sizing, alternate, fallout) and mass, speed, range, thrust, and payload data specified in the vehicle input .csv file.
```
crecine, line 25: great description, thanks

PR #287, opened 2024-05-28
fixed bugs related to specifying phase_info from the CLI

crecine:
### Summary

phase_info was being treated as a dict when it was passed to run_aviary
added a preprocessor to run_level_1 that extracts phase_info and phase_info_parameterization (if present) from the file path that is specified through the CLI and then passes them to run_aviary.
Also prevented the FileNotFoundError raised in get_model from causing get_path to fail prematurely.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/test/test_cmd_entry_points.py
+++ b/aviary/interface/test/test_cmd_entry_points.py
@@ -34,9 +34,9 @@ class CommandEntryPointsTestCases(unittest.TestCase):
             ' --optimizer IPOPT --max_iter 1 --shooting'
         self.run_and_test_cmd(cmd)
 
-    @require_pyoptsparse(optimizer="SNOPT")
+    @require_pyoptsparse(optimizer="IPOPT")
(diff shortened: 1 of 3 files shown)


aviary/interface/test/test_cmd_entry_points.py
```diff
@@ -34,6 +34,12 @@ def bench_test_IPOPT_cmd(self):
             ' --optimizer IPOPT --max_iter 1 --shooting'
         self.run_and_test_cmd(cmd)
 
+    @require_pyoptsparse(optimizer="SNOPT")
```
johnjasa, line 37: nitpick: can you make this IPOPT so it's tested on CI please?

johnjasa (approved):
Love it, this addresses a point of confusion that @cmbenne3 had as well! It's helpful to allow `phase_info` to be a filepath at the `run_aviary` level. Thanks for this, Carl.

jkirk5 (approved):

PR #298, opened 2024-05-29
added a warning if a variable is not in meta_data nor in guess_names

### LINKED ISSUE #56: There is no checking on hierarchy variables read in from a csv file.
### Description

If I typo one of the aircraft:* or mission:* hierarchy names, there is no warning or error that I have made a mistake.

Note: the aviary reader is going to see this as a non-hierarchy variable, since you can set any top-level variable in the csv. Maybe we should check the list of var names that get read in, and raise a warning or error for any that are found not to be in the aggregate hiearchy metadata and/or the model.

### Example

"aircraft:wing:mass_scaler" which I could typo as "aircraft:wing:mass_scalar".
It is not a required variable, so I would never know without careful checking versus list_inputs.

### Aviary Version

0.9.2-dev

_No response_

xjjiang:
### Summary

This is to respond to issue #56. If a user variable in csv file is not in meta_data nor in guess_names, then print a warning message. This variable is ignored.

### Related Issues

- Resolves #56

None

### New Dependencies

None


aviary/utils/process_input_decks.py
```diff
@@ -139,6 +139,11 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
                 initial_guesses[var_name] = float(var_values[0])
                 continue
```
Kenneth-T-Moore, line 141: Can you add a test for this? The test should assert that the right warning is raised. You can see an example of this here:

https://github.com/OpenMDAO/OpenMDAO/blob/c78c82526cdc2c0edd88a8d3695e6704cee0d8a1/openmdao/core/tests/test_problem.py#L1237

You won't need to actually run an optimization or anything. This is raised earlier than that.

aviary/utils/process_input_decks.py
```diff
@@ -139,6 +139,11 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
...
 
+            else:
+                warnings.warn(
+                    f"Variable '{var_name}' is not in meta_data nor in 'guess_names'. It is ignored.")
+                continue
+
             if aircraft_values.get_val('verbosity').value >= 2:
                 print('Unused:', var_name, var_values, comment)
```
crecine, line 148: we originally decided to use a print instead of a warning because there are many unused variables we don't want warnings for. Maybe we could change this to be a warning if the variable name contains a ":"

crecine: or if the variable name starts with "aircraft" or "mission"

xjjiang: I use ":" to do the test.

Kenneth-T-Moore: @crecine

I guess you mean the sections of unconverted variables that are placed at the bottom of the csv file by the converter?  I'm not sure we should design around that. Maybe we should comment out those lines instead.

aviary/utils/process_input_decks.py
xjjiang: A unit test is added.

Kenneth-T-Moore (approved):

crecine (approved):

aviary/utils/process_input_decks.py
```diff
@@ -139,7 +139,13 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
                 initial_guesses[var_name] = float(var_values[0])
                 continue
 
-            if aircraft_values.get_val('verbosity').value >= 2:
+            elif ":" in var_name:
+                warnings.warn(
+                    f"Variable '{var_name}' is not in meta_data nor in 'guess_names'. It is ignored.",
```
crecine, line 144:
```suggestion
                    f"Variable '{var_name}' is not in meta_data nor in 'guess_names'. It will be ignored.",
```

xjjiang:
I replaced `verbosity` to `settings:verbosity` in `utils/process_input_decks.py`.

xjjiang:
In Aviary, we set `verbosity` in two ways: one in .cvs file, it is `settings:verbosity`, the other is as a function attribute, it is `verbosity`. We should use the Settings.VERBOSITY everywhere and not the string itself. I will work on that later.

aviary/utils/process_input_decks.py
```diff
@@ -183,7 +189,7 @@ def update_GASP_options(aircraft_values: AviaryValues()):
         aircraft_values.set_val(
             Aircraft.Wing.FOLD_DIMENSIONAL_LOCATION_SPECIFIED, val=False)
 
-    if aircraft_values.get_val('verbosity').value >= 2:
+    if aircraft_values.get_val('settings:verbosity').value >= 2:
```
crecine, line 192: Here and elsewhere in this file:
```suggestion
    if aircraft_values.get_val(Settings.VERBOSITY).value >= 2:
```

aviary/utils/process_input_decks.py
```diff
@@ -333,7 +339,7 @@ def initial_guessing(aircraft_values: AviaryValues(), initial_guesses):
         initial_guesses['climb_range'] = initial_guesses['time_to_climb'] / \
             (60 * 60) * (avg_speed_guess * np.cos(gamma_guess))
 
-    if aircraft_values.get_val('verbosity').value >= 2:
+    if aircraft_values.get_val('settings:verbosity').value >= 2:
```
crecine, line 342:
```suggestion
    if aircraft_values.get_val(Settings.VERBOSITY).value >= 2:
```

PR #300, opened 2024-05-29
Command line verbosity option

### LINKED ISSUE #284: Command line verbosity option error for fortran_to_aviary
### Description

Unable to select different verbosity options using the aviary command line interface when running the fortran_to_aviary function.
The documentation (and help command) suggests that you can parse an optional integer to change the verbosity -v {0,1,2,3}, but when attempting to do this it results in an error.

### Example

> aviary fortran_to_aviary -o aviary_input.csv -l FLOPS -v 3 --force "Flops_Input_example"

returns following error:
> usage: aviary fortran_to_aviary [-h] [-o OUT_FILE] -l {FLOPS,GASP} [-d DEFAULTS_DECK] [--force] [-v {0,1,2,3}] input_deck
aviary fortran_to_aviary: error: argument -v/--verbosity: invalid Verbosity value: '3'

### Aviary Version

0.9.4-dev

Either the documentation or acceptable options list needs updating with the correct way to input verbosity, or the function _setup_F2A_parser() needs modification:

Suggested modification change:
>parser.add_argument(
        "-v",
        "--verbosity",
        type=Verbosity,
        choices=list(Verbosity),
        default=1,
        help="Set level of print statements",
    )

to:
>parser.add_argument(
        "-v",
        "--verbosity",
        type=int,
        choices={0,1,2,3},
        default=1,
        help="Set level of print statements",
    )

On testing this seems to work, but there is probably a better way to do this.

xjjiang:
### Summary

This is to fix issue #284. It is reported that the following command results in error:

`aviary fortran_to_aviary -o aviary_input.csv -l FLOPS -v 3 --force "Flops_Input_example"`

This PR tries to fix this error.

### Related Issues

- Resolves #284.

None

### New Dependencies

None


xjjiang:
@cmbenne3 Is this good enough?

xjjiang:
@cmbenne3 , we don't want to hard code `{0, 1, 2, 3}`. Instead, we want to get them from Verbosity class. That was my motivation for the change. Let's see how others respond.

cmbenne3:
yes - I agree that we don't want to hardcode the verbosity options in case we change the class - good thinking!

crecine (approved):

jkirk5:
I tinkered around a little and couldn't find a quick solution other than the lambda function you made, which is hard to parse but I think I get the gist of what it does

jkirk5 (approved):

crecine:
> I tinkered around a little and couldn't find a quick solution other than the lambda function you made, which is hard to parse but I think I get the gist of what it does
would you prefer?
`return [c.value for c in cls]`

xjjiang:
Done as suggested by Carl.

aviary/variable_info/enums.py
```diff
@@ -195,3 +195,7 @@ class Verbosity(Enum):
 
     def __str__(self):
         return str(self.value)
+
+    @classmethod
+    def set(cls):
```
crecine, line 200: I think `values` would be a better name than `set` for this method.

xjjiang: agree. done.

PR #303, opened 2024-06-04
Combine verbosity

xjjiang:
### Summary

In Aviary, we set verbosity in two ways: one in .cvs file, it is `settings:verbosity`, the other is as a function attribute, it is `verbosity`. As a result, both of them are written into `aircraft_values`. This PR will combine them into a single value - `settings:verbosity`.

I didn't touch the string `verbosity` in `fortran_to_aviary.py` because they seem to have a different meaning. They are not supposed to write into .csv files.

Please ignore `methods_for_level1.py` and `api.py`. They should not be in this PR.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/utils/process_input_decks.py
+++ b/aviary/utils/process_input_decks.py
@@ -60,18 +60,30 @@ def create_vehicle(vehicle_deck='', verbosity=Verbosity.BRIEF, meta_data=_MetaDa
     aircraft_values.set_val(Settings.PROBLEM_TYPE, val=ProblemType.SIZING)
     aircraft_values.set_val(Aircraft.Electrical.HAS_HYBRID_SYSTEM, val=False)
 
+    initial_guesses = {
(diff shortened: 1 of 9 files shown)


aviary/utils/process_input_decks.py
```diff
@@ -139,7 +139,7 @@ def parse_inputs(vehicle_deck, aircraft_values: AviaryValues(), meta_data=_MetaD
                 initial_guesses[var_name] = float(var_values[0])
                 continue
 
-            if aircraft_values.get_val('verbosity').value >= 2:
+            if aircraft_values.get_val('settings:verbosity').value >= 2:
```
crecine, line 142: Can you use `Settings.VERBOSITY` for the three `get_val`s here as well?

xjjiang: Done.

johnjasa (approved):

crecine (approved):

PR #304, opened 2024-06-04
Rename run_aviary

xjjiang:
### Summary

Rename `run_aviary` to `setup_and_run_aviary` because that is what it does.

Note: it will cause backward incompatibility. Should we make this change?

### Related Issues

- Resolves #

updating the api will cause backwards incompatibility for any users currently importing run_aviary from the api

### New Dependencies

None

### DIFF
--- a/aviary/api.py
+++ b/aviary/api.py
@@ -35,7 +35,7 @@ from aviary.interface.default_phase_info.two_dof import phase_info as default_2D
 from aviary.interface.default_phase_info.two_dof_fiti import create_2dof_based_ascent_phases, create_2dof_based_descent_phases
 from aviary.interface.default_phase_info.height_energy import phase_info as default_height_energy_phase_info
 from aviary.interface.methods_for_level1 import run_level_1
-from aviary.interface.methods_for_level1 import run_aviary
+from aviary.interface.methods_for_level1 import setup_and_run_aviary
 from aviary.interface.methods_for_level2 import AviaryProblem
 from aviary.interface.utils.check_phase_info import check_phase_info
 from aviary.utils.engine_deck_conversion import EngineDeckConverter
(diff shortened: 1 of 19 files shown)


crecine:
This might be a good opportunity to add a deprecation warning to the API, and support both for now.

xjjiang:
> This might be a good opportunity to add a deprecation warning to the API, and support both for now.

It would be a good practice for the future. Any guideline? @crecine @Kenneth-T-Moore

crecine:
@Kenneth-T-Moore 
I tried adding a deprecation warning, but running `examples/run_basic_aviary_example.py` doesn't actually produce a warning.
https://github.com/xjjiang/om-Aviary/pull/1/files

johnjasa:
I'm not certain if this change is a helpful one for users and I want to discuss it here.

Although the `run_aviary` command does indeed setup and run the model, I think the name `setup_and_run_aviary` is clunky, especially for low-level uses of Aviary. I think `run_aviary` is much more memorable and clear.

@xjjiang could you tell me more about the benefits of renaming here? I welcome input from others too!

Kenneth-T-Moore:
@crecine  
I am not sure what is swallowing that warning. I would have to play around.

@johnjasa 
I am not sure we need this change either. I think the level 1 user should stay unaware of the setup/run distinction until they have learned the basics.

crecine:
I believe this was intended to reduce some confusion regarding the difference between `run_aviary` and `run_aviary_problem`.
Are we expecting level 1 users to be importing from the api? I would think that this would have the biggest effect on level 2 users that want to script/automate running Aviary

xjjiang:
I just thought `setup_and_run_aviary` is what this function actually does. It is okay if we keep as it is.

xjjiang:
I will withdraw this PR in one hour if you are not appose it.

PR #312, opened 2024-06-06
Update OAS Example with figure and link

jdgratz10:
### Summary

This PR fixes a broken URL in the OAS example and adds an xdsm of the pre-mission group.

None

### New Dependencies

None


Kenneth-T-Moore (approved):

aviary/docs/examples/images/OAS_xdsm.PNG
crecine, line 1: "Geometry Parameters" into aerodynamics is plural, but engine and geometry parameter into FLOPS based mass are singular. Is that intentional?
Should there be any outputs from mass, geometry, or aero called out?

jdgratz10: Regarding calling out outputs I don't think that's necessary here. Regarding the plurality, I will push a fix.

crecine (approved):

PR #316, opened 2024-06-10
Enables minimal print for brief and verbose

crecine:
### Summary

when using SNOPT or IPOPT, setting `driver.options['print_results'] = True` prints a significant amount of information to the terminal.
Changing this to `minimal` will display only Variables and Constraints with non-blank exit statuses.
For example, if a variable is on it's lower bound and has the status "l".
This makes it easier to find relevant information when it isn't buried amongst hundreds of lines of output.

### Example

   Objectives
      Index  Name                                   Value
          0  mission:summary:fuel_burned     2.420137E+00

   Variables (c - continuous, i - integer, d - discrete)
      Index  Name                                     Type      Lower Bound            Value      Upper Bound     Status
          1  traj.climb.t_duration_0                     c     5.000000E-01     5.000000E-01     1.500000E+00          l
         32  traj.cruise.t_duration_0                    c     5.000000E-01     1.500000E+00     1.500000E+00          u

   Constraints (i - inequality, e - equality)
      Index  Name                                                              Type          Lower           Value           Upper    Status  Lagrange Multiplier (N/A)


Kenneth-T-Moore (approved):
I think this would be fine. I do use the printed info, but what I look for should be covered in the "minimal" set, provided that active and violated constraints and desvar bounds are all there.

xjjiang:
Should you describe this feature in docs?

crecine:
> Should you describe this feature in docs?

I don't think we need to. This is just using a feature that is available in Openmdao/pyoptsparse based on the value of verbosity, which is already documented.

aviary/examples/reserve_missions/run_reserve_mission_fixedtime.py
```diff
@@ -40,8 +41,7 @@
 
 # Link phases and variables
 prob.link_phases()
-
-prob.add_driver("SLSQP", max_iter=100)
+prob.add_driver("SNOPT", max_iter=100, verbosity=Verbosity.VERBOSE)
```
johnjasa, line 44: Does changing this to SNOPT negatively affect how this script is tested? Or does that not come into play?

johnjasa (approved):

PR #328, opened 2024-06-13
Hydraulic pressure correction

ChrisPsenica:
### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None


ChrisPsenica:
Please note: I also adjusted the upper bound limit of the gross weight from 200000Lbs to 800000Lbs

johnjasa (approved):

aviary/interface/methods_for_level2.py
```diff
@@ -1702,7 +1702,7 @@ def add_design_variables(self):
             optimize_mass = self.pre_mission_info.get('optimize_mass')
             if optimize_mass:
                 self.model.add_design_var(Mission.Design.GROSS_MASS, units='lbm',
-                                          lower=100.e2, upper=200.e3, ref=135.e3)
+                                          lower=100.e2, upper=900.e3, ref=135.e3)
```
crecine, line 1705: in the past we had the upper limit set to `None` to prevent this issue.
@johnjasa should we go back to None?

crecine (approved):

aviary/interface/methods_for_level2.py
johnjasa: I don't think so as IPOPT benefits from having a reasonable upper limit for its auto-scaling

PR #331, opened 2024-06-18
Engine deck conversion bugfix

jkirk5:
### Summary

Engine deck conversion is supposed to verify that data can be converted to float and re-sort data before writing. This formatted data wasn't actually being used, due to a possible typo or oversight


johnjasa (approved):

Kenneth-T-Moore (approved):
Approved, but are the old, improperly-sorted decks generated by this tool still ok to use?

crecine:
Git won't let me add a pr to this branch, but here is some cleanup of some for loops and dictionary handling
```
    # sort data
    # create parallel dict to data that stores floats
    formatted_data = {}
    for key in data:
        formatted_data[key] = data[key].astype(float)

    # convert engine_data from dict to list so it can be sorted
    sorted_values = np.array(list(formatted_data.values())).transpose()

    # Sort by mach, then altitude, then throttle, then hybrid throttle
    sorted_values = sorted_values[np.lexsort(
        [formatted_data[THROTTLE],
         formatted_data[ALTITUDE],
         formatted_data[MACH]])]
    for idx, key in enumerate(formatted_data):
        formatted_data[key] = sorted_values[:, idx]

    # store formatted data into NamedValues object
    write_data = NamedValues()
    for key in data:
        write_data.set_val(header_names[key], formatted_data[key], default_units[key])

    write_data_file(output_file, write_data, comments, include_timestamp=False)
```

jkirk5:
> Approved, but are the old, improperly-sorted decks generated by this tool still ok to use?

Yep! For 2 reasons: first, the FLOPS and GASP decks we are converting from are already sorted, and also the EngineDeck component will also sort the data after reading in from a file.

Basically, this is only important for human-readability of the engine decks

PR #335, opened 2024-06-18
Refactor assert that checks whether the variable hierarchy is in proper alphabetical order.

Kenneth-T-Moore:
### Summary

Refactor the assert that checks whether the variable hierarchy is in proper alphabetical order so that it is cleaner and works with arbitrary depth.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -1867,7 +1867,7 @@ class AviaryProblem(om.Problem):
 
                                 targets = mission_var_name
                                 if '.' in mission_var_name:
-                                    # Support for non-hiearchy variables as parameters.
+                                    # Support for non-hierarchy variables as parameters.
                                     mission_var_name = mission_var_name.split('.')[-1]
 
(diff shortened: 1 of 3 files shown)


crecine (approved):

jkirk5 (changes requested):
Tested with a multi-level variable hierarchy - unexpected behavior. Alphabetization required hierarchy to be formatted like:
```
class A:
    b= 'var1'
    class C:
        x = 'another_var'
    cb='var_cb'
    cx='var_cx'
    class D:
        var='var'
 etc...
```
 
 Which is hard to read. Do we have a clean way to allow for subclass definitions to be group together and not distributed amongst the other variables at that higher class level?

Kenneth-T-Moore:
Thanks for testing it out, Jason. Yeah, that doesn't look ideal, so I reworked the logic to allow the subclasses to be listed after the variables, which is much more readable.

jkirk5 (approved):

PR #336, opened 2024-06-18
Improve reading vector variables from csv

jkirk5:
### Summary

Added some protections when reading a variable from .csv that is supposed to be a vector (e.g. engine wing locations). Attempts to account for the fact that sometimes a vector means each index goes to an engine, but in a single-engine-type aircraft sometimes the whole vector is supposed to pair with the engine


aviary/subsystems/propulsion/utils.py
```diff
@@ -135,13 +136,45 @@ def build_engine_deck(aviary_options: AviaryValues):
...
+                default_value = meta_data[var]['default_value']
+                # special handling for iterables - check if they are multidimensional,
+                # which implies multiple engine models, and only use the value intended
+                # for the first engine model
                 if isinstance(aviary_val, np.ndarray):
+                    try:
+                        expected_dim = default_value.ndim
+                    except AttributeError:
+                        # metadata is not expecting a numpy array, just a single value
+                        expected_dim = 0
```
crecine, line 148: would a getattr with a default value be better here?

aviary/utils/preprocessors.py
```diff
@@ -175,16 +185,16 @@ def preprocess_propulsion(aviary_options: AviaryValues, engine_models: list = No
             # 2. aviary_options
             # 3. default value from metadata
             for i, engine in enumerate(engine_models):
-                if var == Aircraft.Nacelle.AVG_DIAMETER:
-                    pass
+                # if var == Aircraft.Nacelle.AVG_DIAMETER:
+                #     pass
```
crecine, line 189: remove?

crecine (approved):

aviary/subsystems/propulsion/utils.py
jkirk5: This is just checking if we are expecting an array by checking metadata, and if so how many dimensions that array should be. Never actually overwriting provided values with a default

jkirk5: FYI I just overhauled the try/except statement structure so this code is about to become outdated

johnjasa (approved):
Does it make sense to add a test for these multidimensional array cases? Or are those tests not relevant yet?

jkirk5:
> Does it make sense to add a test for these multidimensional array cases? Or are those tests not relevant yet?

Yes, it would make sense to have a test for this - we currently don't have a working aircraft w/ 2 or more engines per wing without running into other issues (multiple engine types and turboprops currently not supported from csv level). So I'll say future work (to either create a new test case or add these features to level 1 and change an existing test case to read from csv instead of L2 interface)

PR #334, opened 2024-06-18
Convert GASP propeller map to Aviary CSV format

xjjiang:
### Summary

1. Add an option `convert_prop_table` to aviary commands. A sample command is: `aviary convert_prop_table -f GASP gasp_prop.map new_prop.prop`. Right now, `-f GASP` does not seem to be necessary but in the future we may add other formats to be converted.
2. Add a utility to convert GASP propeller map to Aviary CSV format.
3. Add a unit test for the new feature.
4. Add two sample propeller maps - one for Propfan (using Mach number) and the other for King Air90 propeller (using Helical Mach at 75% Radius).
5. There are some duplicate code shared by `engine_deck_conversion.py` and `propeller_map_conversion.py`. They are moved to `conversion_utils.py`.
6. Add the new feature to docs.

### Related Issues

- Resolves #

### Backwards incompatibilities

None

### New Dependencies

None

PR #340, opened 2024-06-24
Correcting the example commands for the engine deck conversion and adding a notice for future users

ChrisPsenica:
The Aviary website gives instructions on how to convert an engine deck using the convert_engine command in Aviary. These commands are not entirely correct however. When running the command exactly as it is from the Aviary website (using the FLOPS conversion for example) there is an error messaged returned saying that FLOPS is not a recognized command:

Screenshot 2024-06-24 113450

This happens for all conversions (FLOPS, GASP, GASP_TP). However this issue is solved when modifying the given command from the Aviary website from "aviary convert_engine turbofan_22k.eng turbofan_22k.txt FLOPS" to "aviary convert_engine turbofan_22k.eng turbofan_22k.txt -f FLOPS". The added '-f' before FLOPS allows the command to be recognized. Additionally, I added another bullet point note to the same page where it informs the user that in order for the engine conversion to work the original engine deck file must not have any headers to it. When leaving in the headers an error message is returned stating troubles with converting a string to float: 

Screenshot 2024-06-24 113920

### DIFF
--- a/aviary/docs/user_guide/aviary_commands.ipynb
+++ b/aviary/docs/user_guide/aviary_commands.ipynb
@@ -232,13 +232,14 @@
     "- TurboFan decks for both FLOPS and GASP can be converted\n",
     "- TurboProp decks for GASP can also be converted\n",
     "- Comments can be added using #\n",
+    "- The engine deck file that is to be converted must not have any headers\n",
     "\n",
     "\n",
     "Example usage:\n",
     "```\n",
-    "`aviary convert_engine GASP_turbofan_23k_1.eng turbofan_23k_1_lbm_s.deck GASP` Convert a GASP based turbofan\n",
-    "`aviary convert_engine turbofan_22k.eng turbofan_22k.txt FLOPS` Convert a FLOPS based turbofan\n",
-    "`aviary convert_engine turboprop_4465hp.eng turboprop_4465hp.deck GASP_TP` Convert a GASP based turboprop\n",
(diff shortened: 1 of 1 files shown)


ChrisPsenica:
Pull request to clarify some instructions on the Aviary website

crecine:
This was recently fixed by https://github.com/OpenMDAO/Aviary/pull/334/files however, the comment about the headers could still be useful (though I think it only applies to FLOPS decks)

jkirk5:
Adding a comment here to preserve some side conversations:
Data with headers isn't supported by FLOPS, so a failure here is expected.
IMO it's not necessary to specifically call out "no headers" since headers aren't a part of the FLOPS data format. Basically, the list of things that _can't_ be in the data file is infinite. Instead, I'd rather see more helpful failure messages - right now we don't really check the data to catch issues so we just hard crash with an error like shown above. Adding some line-by-line checks for things like convertible to float, expected no. of columns, etc. would be helpful to the user.

aviary/docs/user_guide/aviary_commands.ipynb
```diff
@@ -232,13 +232,14 @@
     "- TurboFan decks for both FLOPS and GASP can be converted\n",
     "- TurboProp decks for GASP can also be converted\n",
     "- Comments can be added using #\n",
+    "- The engine deck file that is to be converted must not have any headers\n",
```
jkirk5, line 235: I'm requesting this line get removed.
GASP decks actually do have a "header" line with additional variables (not headers in the traditional sense because they don't label the data), so this line might actually cause confusion for GASP users. Also see my other comment on the PR.

jkirk5 (changes requested):
Good catch on docs inconsistency

PR #353, opened 2024-06-26
Height Energy Off-Design

kaushikponnapalli:
### Summary

Modified height energy code to enable alternate and fallout missions.
Changed all mission related components to use summary.gross_mass instead of design.
Unified how the range residual and objectives are set for 2-dof and height energy missions.
Added a basic example case

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/models/test_aircraft/aircraft_for_bench_FwFm.csv
```diff
@@ -146,7 +146,7 @@ aircraft:wing:wetted_area,2396.56,ft**2
 mission:constraints:max_mach,0.785,unitless
 mission:design:cruise_altitude,35000,ft
 mission:design:gross_mass,175400.0,lbm
-mission:design:range,3500,NM
```
Kenneth-T-Moore, line 149: I wouldn't change this mass because multiple test/benches use this file. In your test where you do need 2000 NM, I would just set this variable locally after loading the csv, but before making any other level2 calls. (prob.aviary_inputs holds all of these at that time)

aviary/interface/methods_for_level2.py
```diff
@@ -1711,12 +1695,41 @@ def add_design_variables(self):
...
             if optimize_mass:
                 self.model.add_design_var(Mission.Design.GROSS_MASS, units='lbm',
                                           lower=100.e2, upper=200.e3, ref=135.e3)
 
+        elif self.mission_method is HEIGHT_ENERGY:
+            if self.problem_type is ProblemType.SIZING:
+                self.model.add_design_var(Mission.Design.GROSS_MASS, units='lbm',
+                                          lower=100e2, upper=200e3, ref=135e3)
+
+                self.model.add_design_var(Mission.Summary.GROSS_MASS, units='lbm',
```
Kenneth-T-Moore, line 1709: What is the difference between Design Gross Mass, and Summary Gross Mass? Are they truly independent variables that the optimizer should control separately?

crecine: Mission.Design.GROSS_MASS is the gross weight for the design mission, which drives the structural requirements->OEW->GTOW as part of the design loop.
Mission.Summary.GROSS_MASS is the gross mass at the start of the current mission. For the design mission, this is equal to the design gross mass, for any other mission it may or may not be. For non-design missions, the design gross mass is held constant, but the summary gross mass is allowed to change. This prevents the OEW from resizing

crecine: also, I believe Kaushik is planning to combine this with the 2DOF section below, which has some comments explaining the different constraints/DVs

aviary/interface/methods_for_level2.py
```diff
@@ -1701,103 +1701,77 @@ def add_design_variables(self):
...
                 self.model.add_constraint(
-                    Mission.Constraints.RANGE_RESIDUAL, equals=0, ref=10,
+                    Mission.Constraints.RANGE_RESIDUAL, equals=0, ref=10
                 )
+
             elif self.problem_type is ProblemType.FALLOUT:
                 print('No design variables for Fallout missions')
 
+            if self.mission_method is TWO_DEGREES_OF_FREEDOM:
+                if self.analysis_scheme is AnalysisScheme.COLLOCATION:
```
crecine, line 1760: this could just be 
`if (self.mission_method is TWO_DEGREES_OF_FREEDOM) and (self.analysis_scheme is AnalysisScheme.COLLOCATION):`
right?

aviary/interface/methods_for_level2.py
Kenneth-T-Moore: Thanks. I think this makes sense now.

Kenneth-T-Moore (approved):

aviary/examples/run_aviary_example.py
```diff
@@ -10,6 +10,7 @@
 """
 import aviary.api as av
 from example_phase_info import phase_info
+# phase_info['pre_mission']['include_takeoff'] = True
```
crecine, line 13: remove?

aviary/interface/methods_for_level2.py
kaushikponnapalli: I have absolutely no idea why, but making the change you suggested causes `AviaryMissionTimeseries.test_timeseries_report` to fail. The test passes if I leave it as is.

crecine: I think #433 should have fixed this

crecine: Could you try re-doing this change and pulling in main to see if it passes?

aviary/examples/off_design_example.py
```diff
@@ -0,0 +1,224 @@
...
+            "no_climb": True,
+        },
+        "initial_guesses": {"time": ([241, 30], "min")},
+    },
+    "post_mission": {
+        "include_landing": True,
+        "constrain_range": True,
+        "target_range": (3375.0, "nmi"),
+    },
+}
```
crecine, line 102: would using the default phase info for height energy (modifying values as necessary) work here? I don't think there is anything special here for the design mission, and excluding these lines of code would make the important parts of the example more clear.

aviary/examples/off_design_example.py
```diff
@@ -0,0 +1,224 @@
...
+# Load optimization problem formulation
+# Detail which variables the optimizer can control
+prob.add_objective()
+
+prob.setup()
+
+prob.set_initial_guesses()
+
+prob.run_aviary_problem()
+prob_fallout = av.AviaryProblem()
```
crecine, line 135: could we get some comments here and/or some sort of horizontal line (ie `#########`) to separate the design, fallout, and alternate sections of the code?

crecine (commented):
A couple of pretty minor comments, but this is looking good.

aviary/examples/off_design_example.py
```diff
@@ -101,6 +101,9 @@
     },
 }
 
+################
+# Sizing Mission
+################
```
crecine, line 106:
```suggestion
##################
# Sizing Mission #
##################
```

aviary/examples/off_design_example.py
```diff
@@ -132,6 +135,10 @@
 prob.set_initial_guesses()
 
 prob.run_aviary_problem()
+
+#################
+# Fallout Mission
+#################
```
crecine, line 141:
```suggestion
###################
# Fallout Mission #
###################
```

aviary/examples/off_design_example.py
```diff
@@ -159,6 +166,9 @@
 prob_fallout.set_initial_guesses()
 prob_fallout.run_aviary_problem()
 
+###################
+# Alternate Mission
+###################
```
crecine, line 171:
```suggestion
#####################
# Alternate Mission #
#####################
```

crecine (approved):
minor formatting

PR #311, opened 2024-06-05
Adding more executable code in the docs

crecine:
### Summary

This will help prevent the docs from becoming stale as the code changes, by loading modules, checking for variables, and making sure that values are correct
The first step involves converting .md files to .ipynb files Then after reading through each doc section, determine if there is anything that could become stale (variable names, functions, etc)

### Related Issues

- Resolves #

### Backwards incompatibilities

None

### New Dependencies

None

PR #397, opened 2024-07-01
Trying glue

crecine:
### Summary

glue allows us to connect the text in the markdown cell to the code in the python cells to ensure that the markdown cell doesn't get stale

https://jupyterbook.org/en/stable/content/executable/output-insert.html
https://myst-nb.readthedocs.io/en/latest/render/glue.html

Uses:
```
from myst_nb import glue
from IPython.display import Markdown
```
which (I'm pretty sure) are installed automatically by jupyter-book

### Related Issues

*Note*: this builds off of #311

None

### New Dependencies

None


aviary/docs/user_guide/propulsion.ipynb
```diff
@@ -145,7 +145,7 @@
     "ae = Aircraft.Engine\n",
     "\n",
     "required = (ae.DATA_FILE, ae.SCALE_PERFORMANCE, ae.IGNORE_NEGATIVE_THRUST, ae.GEOPOTENTIAL_ALT, ae.GENERATE_FLIGHT_IDLE)\n",
-    "# ********NOTE******** #\n",
+    "# ****NOTE**** #\n",
     "# these do not match, should the docs be updated?\n",
     "# check_value(required,required_options)"
```
crecine, line 150: @jkirk5

aviary/docs/user_guide/propulsion.ipynb
```diff
@@ -340,7 +340,7 @@
    },
    "outputs": [],
    "source": [
-    "# ********NOTE********\n",
+    "# ****NOTE****\n",
     "# this fails if no engine models are provided:\n",
```
crecine, line 344: @jkirk5

aviary/docs/user_guide/pre_mission_and_mission.ipynb
```diff
@@ -96,7 +96,7 @@
    "outputs": [],
    "source": [
     "# Testing Cell\n",
-    "# ********NOTE******** #\n",
+    "# ****NOTE**** #\n",
     "# Is this supposed to be bus_variables or parameters?\n",
```
crecine, line 100: @johnjasa @jdgratz10

crecine: [jdgratz10](https://github.com/jdgratz10) [yesterday](https://github.com/OpenMDAO/Aviary/pull/311#discussion_r1661355988)
I believe based on the context it should be bus_variables, since this example is walking through how to use the bus. The get_parameters function is for variables that are dymos parameters during mission, while the get_bus_variables function is for variables that are indeed dymos parameters during the mission but that are also calculated during the pre-mission portion of the external subsystem. John, feel free to correct me if something has changed here, but that's the way it was back when I built the bus port initially.

crecine: here the docs say:

>"Variables that begin with 'aircraft:' are connected when you use the get_parameters() method within SubsystemBuilderBase.\n"

however, neither of these methods appears to do that

aviary/docs/user_guide/subsystems.ipynb
```diff
@@ -72,7 +72,7 @@
     "\n",
     "for func in expected_flow:\n",
     "    getattr(AviaryProblem, func)\n",
-    "    # ********NOTE********\n",
+    "    # ****NOTE****\n",
     "    # check functions in each function?"
```
crecine, line 76: @johnjasa

aviary/docs/user_guide/outputs_and_how_to_read_them.ipynb
```diff
@@ -176,7 +176,7 @@
    "outputs": [],
    "source": [
     "# Testing Cell\n",
-    "# ********NOTE********\n",
+    "# ****NOTE****\n",
     "# run with a dummy set of aviary_inputs and check that _list files are only generated for correct verbosity?"
```
crecine, line 180: @johnjasa

crecine (commented):
Moving the conversations from PR #311 so that it can be merged as is and remaining notes can be addressed here

aviary/docs/examples/OAS_subsystem.ipynb
```diff
@@ -191,7 +191,7 @@
    "outputs": [],
    "source": [
     "# Testing Cell\n",
-    "# ********NOTE********\n",
+    "# ****NOTE****\n",
     "# check variables here?"
```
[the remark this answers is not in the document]
Kenneth-T-Moore: To test this, put the OASWingWeightBuilder into a simple aviary model by adding it to the premission subsystems. Then, follow the level 2 interface through `setup`.  No need to go any further and run it. We are testing the cell that shows them how to write this builder, so we really want to make sure that aviary calls build_pre_mission correctly as we update the arguments to this call in the future.  No need to test any values; other tests should cover execution.

aviary/docs/user_guide/variable_metadata.ipynb
```diff
@@ -292,7 +292,7 @@
     "# Testing Cell\n",
     "from aviary.utils.merge_variable_metadata import merge_meta_data\n",
     "from aviary.utils.merge_hierarchies import merge_hierarchies, merge_two_hierarchies\n",
-    "# ********NOTE********\n",
+    "# ****NOTE****\n",
     "# which of these should merge_variable_hierarchies above be?"
```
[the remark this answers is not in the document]
jdgratz10: Yep, that sounds correct!

aviary/docs/user_guide/pre_mission_and_mission.ipynb
jdgratz10: @johnjasa maybe you could chime in here, I don't remember the details of the builder base for this case. However, I'm pretty sure the bus is agnostic to whether or not the variables begin with 'aircraft:'.

johnjasa: Jennifer, I agree with everything you're saying.

I just checked to confirm that `get_parameters()` is called in [`methods_for_level2.py` here](https://github.com/OpenMDAO/Aviary/blob/main/aviary/interface/methods_for_level2.py#L1071) and is used in `setup_trajectory_params` and then has logic to only add variables that have `':'` in the name (so `aircraft:` etc).

I think you could check both `get_parameters()` and `get_bus_variables()` here, but what "check" means is sort of amorphous. Carl, as you noted, those methods _themselves_ don't do much, but those methods are called in `methods_for_level2.py`. Maybe the check is something as simple as making sure that `SubsystemBuilderBase` has methods that are called `get_parameters()` and `get_bus_variables()` to make sure the methods don't change names. I don't think you'd need to do much more than that.

aviary/docs/user_guide/subsystems.ipynb
johnjasa: Yes, I think just checking that each method exists in `SubsystemBuilderBase` is all you need to do here. Let me know if that's not your question.

aviary/docs/user_guide/outputs_and_how_to_read_them.ipynb
johnjasa: Yeah, that's a good idea. You could run with one of the pre-canned example cases and set verbosity then make sure that those lists are made. You probably don't actually need to run a full case; can just do 0 max iter or similar.

aviary/docs/user_guide/propulsion.ipynb
crecine: so should I update the docs to clarify that this is not a list of the minimum set:
image

aviary/docs/user_guide/propulsion.ipynb
crecine: so for this pr, just remove these two sections:
image

aviary/docs/user_guide/pre_mission_and_mission.ipynb
crecine: Should I update the docs to reflect that it adds all variables with a `:` instead of variables that start with `aircraft:`?

aviary/docs/examples/OAS_subsystem.ipynb
```diff
@@ -30,7 +50,7 @@
     "The subsystem builder uses the Aviary external subsystem builder template to connect the analysis tool to Aviary as either a pre-mission, mission or post-mission subsystem.\n",
     "\n",
     "For this case, the analysis tool will compute a wing weight in the pre-mission portion of the Aviary analysis and return its value to Aviary to override the empirical wing weight value.\n",
-    "Fuel weight is passed in from Aviary as the only input currently, but other inputs may also be passed in through the subsystem builder, [OAS_wing_weight_builder](https://github.com/OpenMDAO/Aviary/blob/main/aviary/examples/external_subsystems/OAS_weight/OAS_wing_weight_builder.py), by the `promotes_inputs` parameter.\n",
+    "Fuel weight is passed in from Aviary as the only input currently, but other inputs may also be passed in through the subsystem builder, [OAS_wing_weight_builder](https://github.com/OpenMDAO/Aviary/blob/main/aviary/examples/external_subsystems/OAS_weight/OAS_wing_weight_builder.py), by the {glue:md}`promotes_inputs` parameter.\n",
```
Kenneth-T-Moore, line 53: I am not sure we need to glue-test the OpenMDAO API, though I suppose it doesn't hurt anything.

aviary/docs/tests/utils.py
```diff
@@ -3,6 +3,62 @@
 import tempfile
 import os
 import numpy as np
+from myst_nb import glue
+from IPython.display import Markdown
+
+
+"""
+# DocTAPE #
```
Kenneth-T-Moore, line 11: Did you come up with this acronym?

Kenneth-T-Moore (approved):

aviary/docs/tests/utils.py
crecine: maybe.....

aviary/docs/user_guide/propulsion.ipynb
jkirk5: On the actual Aviary side: you need to define something related to every bullet point on the list, but not necessarily every single variable called out. In that way, it is truly is the minimum variable list.

On the glue side: we should check all of the variables called out here exist within the code

aviary/docs/theory_guide/propulsion.ipynb
```diff
@@ -24,7 +24,7 @@
    "outputs": [],
    "source": [
     "# Testing Cell\n",
-    "# ********NOTE******** should we be using EngineModels vs EngineModel consistently here?\n"
+    "# ****NOTE**** should we be using EngineModels vs EngineModel consistently here?\n"
```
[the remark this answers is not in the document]
jkirk5: I guess to be technically correct it could be `EngineModel` and `EngineModel`s but that might look very ugly. No one is usually confused when someone refers to Python objects as plural (numpy arrays, strings, etc.) so I didn't worry much about when I wrote these sections

aviary/docs/user_guide/propulsion.ipynb
jkirk5: Yeah that is ok to remove (for now), it needs to be replaced with an entirely new section for the engine deck builder

jkirk5 (approved):

aviary/docs/user_guide/propulsion.ipynb
crecine: Should `Aircraft.Engine.DATA_FILE` be added to `required_options` in `engine_deck.py`?

aviary/docs/examples/reserve_missions.ipynb
```diff
@@ -36,14 +88,55 @@
...
+    "check_phase_info(phase_info, HEIGHT_ENERGY);\n",
+    "user_opts = phase_info['reserve_cruise']['user_options']\n",
+    "check_value(user_opts['target_duration'],(30, 'min'))\n",
+    "check_value(user_opts['reserve'],True)\n",
+    "all_keys = get_all_keys(phase_info['reserve_cruise'], track_layers=True)\n",
+    "for key in all_keys:\n",
+    "    glue_variable(key.replace('.','-'), md_code=True)\n",
+    "print(all_keys)\n",
+    "\n",
+    "# ****NOTE**** These were not removed, as the docs seem to indicate\n",
```
crecine, line 118: @Kenneth-T-Moore @ehariton

aviary/docs/user_guide/pre_mission_and_mission.ipynb
crecine: @Kenneth-T-Moore

aviary/docs/user_guide/propulsion.ipynb
crecine: @jkirk5

aviary/docs/examples/reserve_missions.ipynb
Kenneth-T-Moore: I think I would consider unexpected behavior as a bug, so it would be fine to put up a bug issue and merge this PR as is.

aviary/docs/user_guide/propulsion.ipynb
jkirk5: Yes, sorry for late reply

PR #430, opened 2024-07-05
Updated Draw Mission GUI for Aviary

jsonirobots:
### Summary

Draw Mission GUI updated to support an easier to use interface with plot coordinate readouts, tabular editing for all data points, editable axes limits, and ability to set more options within the phase info dict through the GUI. Also supports switch between light and dark mode. The phase info creation function is directly sourced from the previous GUI file, with modifications to be able to set more options through the GUI.

### New Dependencies

None - uses Tkinter and Matplotlib for graphics (used with previous GUI), importlib for opening previously made phase infos, pickle for storing persistent settings (resized window geometry and last used theme) to file.

Example Screenshots:
aaa1
aaa2


jkirk5:
@hschilling could you take a look at and see if you have any suggestions?

johnjasa (changes requested):
I like this so far! It's really nice how you can move the points around, this is definitely more full-featured than what was there.

I have some quick comments here, mostly focused on using the tool and less on the underlying code.
- could you please make the first Mach value not 0.0? Maybe something like 0.2. The underlying models in Aviary have issues with very low Mach numbers
- could you add a button on the main GUI screen that says something like "Output phase_info file" or similar? I get that I went up to "File -> Save" but that wasn't immediately intuitive and I wasn't sure if that would create the phase_info file
- there's a scrollbar between the graph view and the points on the right (see pic below); is this intentional? I see that it scrolls the view of the points on the right, though it doesn't feel strictly necessary to me. If there's not a big use case for it, I vote you remove it
scrollbar

Thanks for this PR! I love the changes. I'll give it another lookover tomorrow and dig more into the code as well

jsonirobots:
@johnjasa 
1. I updated the starting mach to be 0.3 instead of 0. 

2. I added an explicit "output phase info" button to the top left on the GUI, moving the change theme button down to the top right. Please let me know if I should modify the location/appearance of it to make it more visible. I did also edit the file menu language to be more explicit about opening/saving phase info.

aaaaaaaPRev3
aaaaaaaPRev1


3. The scrollbar is there because when the check boxes for optimize altitude/mach populate in between point entries, with enough points it is possible to overflow the window. The window can be resized to show them but my idea was to keep a scrollbar there to handle the minimum window size with any number of points. If this number of points is unreasonable, I can remove the scrollbar. 

aaaaaaaPRev2

johnjasa:
@jsonirobots Awesome, thanks for this quick and great changes! Everything you did and said makes sense.

Regarding the scrollbar, this is a nitpick, but could it be on the right side of the points then? In most applications I've seen the scrollbar is on the right side of the window it scrolls.

crecine:
> * could you please make the first Mach value not 0.0? Maybe something like 0.2. The underlying models in Aviary have issues with very low Mach numbers

I think the GASP side works fine at low speed (and might even require it) maybe it would be good to extend mach down to 0, but have a warning or maybe just gray out the low speed and include a note

crecine:
> @jsonirobots Awesome, thanks for this quick and great changes! Everything you did and said makes sense.
> 
> Regarding the scrollbar, this is a nitpick, but could it be on the right side of the points then? In most applications I've seen the scrollbar is on the right side of the window it scrolls.

I agree, having it between them gives me the impression the scrollbar is for plots, not the points

jsonirobots:
> > * could you please make the first Mach value not 0.0? Maybe something like 0.2. The underlying models in Aviary have issues with very low Mach numbers
> 
> I think the GASP side works fine at low speed (and might even require it) maybe it would be good to extend mach down to 0, but have a warning or maybe just gray out the low speed and include a note

I can add a check that triggers when the user tries to save the phase info, would the warning be "Low mach values can cause issues with FLOPS based models" ?
 
Additionally, the create_phase_info function sets mach_bounds to be a constant +/- 0.2, so if the GUI has any point at mach = 0 this bound becomes negative. Would that cause any issues? If so I can add a check to ensure bounds are always 0 or positive.

johnjasa:
> > > * could you please make the first Mach value not 0.0? Maybe something like 0.2. The underlying models in Aviary have issues with very low Mach numbers
> > 
> > 
> > I think the GASP side works fine at low speed (and might even require it) maybe it would be good to extend mach down to 0, but have a warning or maybe just gray out the low speed and include a note
> 
> I can add a check that triggers when the user tries to save the phase info, would the warning be "Low mach values can cause issues with FLOPS based models" ?
> 
> Additionally, the create_phase_info function sets mach_bounds to be a constant +/- 0.02, so if the GUI has any point at mach = 0 this bound becomes negative. Would that cause any issues? If so I can add a check to ensure bounds are always 0 or positive.

I'd suggest the default should still be Mach 0.3 instead of 0.0 as the groundroll and takeoff phases usually get up to about that speed. If someone is setting up a full mission including the detailed takeoff I'm fine with them needing to manually edit the Mach number on the right.

Regardless, you @jsonirobots you should indeed add a check for the `mach_bounds` to make sure it's not below 0.

crecine:
> I can add a check that triggers when the user tries to save the phase info, would the warning be "Low mach values can cause issues with FLOPS based models" ?

Yah, something like that is what I was thinking

> I'd suggest the default should still be Mach 0.3 instead of 0.0

Are the plot bounds editable through the "edit" or "view" menus? If so I'm fine with .3 being the default, I just wanted to make sure that there wasn't no way to set lower numbers

johnjasa:
> Are the plot bounds editable through the "edit" or "view" menus? If so I'm fine with .3 being the default, I just wanted to make sure that there wasn't no way to set lower numbers

Let me clarify; the lower Mach plot bounds should remain at 0.0 but the default starting value for the first point that a user selects should be 0.3.

aviary/interface/graphical_input.py
```diff
@@ -1,63 +1,1104 @@
...
+        self.title('Mission Design Utility')
+        self.protocol("WM_DELETE_WINDOW",self.close_window)
+        self.focus_set() # focus the window
+        # stores/retrieves persistant settings in source directory
+        self.persist_filename = os.path.join(source_directory,"persist_settings.json")
+
+        # ---------------------------------------------------------------
+        # window geometry definition, allows reuse of user's modified size/location
+        # tkinter size string format:  widthxheight+x+y ;  x,y are location
+        default_win_size = (900,500) # CHANGE THIS TO AFFECT DEFAULT SIZE
```
johnjasa, line 99: Would it make sense or be helpful to have the window size be based on the user's actual monitor resolution? Could use logic like this: https://stackoverflow.com/questions/3129322/how-do-i-get-monitor-resolution-in-python/56913005#56913005

crecine:
> > Are the plot bounds editable through the "edit" or "view" menus? If so I'm fine with .3 being the default, I just wanted to make sure that there wasn't no way to set lower numbers
> 
> Let me clarify; the lower Mach plot bounds should remain at 0.0 but the default starting value for the first point that a user selects should be 0.3.

ahh, thanks for clarifying. That is good with me.

aviary/interface/graphical_input.py
```diff
@@ -212,12 +1253,12 @@ def create_bounds(center):
...
     return phase_info
 
-
 def estimate_total_range_trapezoidal(times, mach_numbers):
+    """Source: original Aviary graphical_input.py."""
     speed_of_sound = 343  # Speed of sound in meters per second
 
     # Convert times to seconds from minutes
-    times_sec = np.array(times) * 60
+    times_sec = np.array(times)
```
johnjasa, line 1261: Change or remove this part as it's no longer doing any conversion from minutes to seconds here

johnjasa (changes requested):
Looking great! Thanks again for this PR and your quick changes to it based on comments.

Do you plan to update the associated documentation in this PR or a subsequent one? There are some places where we reference the mission drawing capabilities, especially this doc page: https://openmdao.github.io/Aviary/user_guide/drawing_and_running_simple_missions.html

aviary/interface/graphical_input.py
jsonirobots: Looking at that bit of code, I edited it so that the message that's shown to the user "Estimated range is xxx nautical miles" uses the same units for range as the units for altitude. Should I add logic to use nmi when altitude is in ft, or is it okay to stick with the same units as altitude?

aviary/interface/graphical_input.py
jsonirobots: I checked with different sizes and the minimize size has to be 800x500, otherwise the screen starts to hide information. However, I did set the default size to use either the minimum size or 50% of user's screen width/height, whichever size is larger. Is this functionality closer to what you had in mind? The user can also opt to store settings and in that case however they resize the window that's how it will open on the next run.

jsonirobots:
Latest commit features Aviary logo for the icon both on the window header and the taskbar (in Windows at least), along with a warning readout at the bottom of the table if table entries are invalid. 
PR_err1
PR_err2

jsonirobots:
> Looking great! Thanks again for this PR and your quick changes to it based on comments.
> 
> Do you plan to update the associated documentation in this PR or a subsequent one? There are some places where we reference the mission drawing capabilities, especially this doc page: https://openmdao.github.io/Aviary/user_guide/drawing_and_running_simple_missions.html

As of now I'm planning to flesh out the code itself in this PR, and once it has been merged I can create another PR to update the docs related to the GUI.

jsonirobots:
And here's the warning message when saving phase info if any mach values are below 0.25. Let me know if that number should be changed, or if anything about this warning behavior should be changed as well.
image

crecine:
This looks great, thanks for making all these changes

hschilling:
Really great GUI! Finally getting to review it. At least in my install, when I click Add New Point, it crashes 

```
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
[0]PETSC ERROR: Try option -start_in_debugger or -on_error_attach_debugger
[0]PETSC ERROR: or see https://petsc.org/release/faq/#valgrind and https://petsc.org/release/faq/
[0]PETSC ERROR: configure using --with-debugging=yes, recompile, link, and run
[0]PETSC ERROR: to get more information on the crash.
[0]PETSC ERROR: Run with -malloc_debug to check if memory corruption is causing the crash.
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI_COMM_WORLD
with errorcode 59.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
```

hschilling:
An earlier version of this GUI that I saw had a Help button with some tips. I thought that was handy. Or maybe a link to the docs for this GUI?

hschilling:
The same crash as above happens when I click on any item in the menu bar. I am on a Mac if that matters.

hschilling:
I can't come up with a repeatable way of making this happen but twice I could drag a point around but the values did not change in the table to the right. I know that doesn't help much! It wasn't like I was doing anything complicated either. I would just add one point, then another, then move the first point I added. No change in the table values

hschilling (approved):
The functionality looks good! The issues with crashing seem specific to my environment so I don't want to hold up this PR because of that.

PR #431, opened 2024-07-05
EngineModels no longer have to contain every supported output

### LINKED ISSUE #157: Engines require outputting all possible variables

jkirk5:
### Summary

Removed requirement that engines always output all supported output variables (e.g. an electroprop no longer has to report a fuel flow of zero). Variables that aren't outputs of an engine model are still defaulted to zero, at the higher propulsion subsystem mission level when all engines are muxed together, so variables like `Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL` still exist in the problem for an all-electric aircraft. This is intentional so checks to see if certain variables exist never have to happen, for example when summing total fuel burn after mission analysis.

Updated EngineDecks to follow this new paradigm.

Performed comments cleanup for many propulsion-related components

### Related Issues

- Resolves #157 

Custom EngineModels that currently have "unnecessary" outputs should be updated to remove them, however they should still work unmodified.

### New Dependencies

None


ehariton (approved):

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -55,7 +55,7 @@
 SHAFT_POWER_CORRECTED = EngineModelVariables.SHAFT_POWER_CORRECTED
 RAM_DRAG = EngineModelVariables.RAM_DRAG
 FUEL_FLOW = EngineModelVariables.FUEL_FLOW
-ELECTRIC_POWER_IN = EngineModelVariables.ELECTRIC_POWER_IN
+ELECTRIC_POWER = EngineModelVariables.ELECTRIC_POWER_IN
```
crecine, line 58: is this intentional?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -187,7 +183,7 @@ def __init__(self, name='engine_deck', options: AviaryValues = None,
         # ensure required variables are a set
         self.required_variables = {*required_variables}
 
-        self._set_variable_flags()
+        # self._set_variable_flags()
```
crecine, line 186: remove?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -466,51 +461,66 @@ def _check_data(self):
             # Check that units are the same. Variables have already been checked for valid
             # units, so it is assumed they are convertable. Prioritizes thrust units
             if engine_variables[RAM_DRAG] != engine_variables[GROSS_THRUST]:
-                self.data[RAM_DRAG] = convert_units(self.data,
-                                                    engine_variables[RAM_DRAG],
-                                                    engine_variables[GROSS_THRUST])
+                data[RAM_DRAG] = convert_units(
+                    original_data[RAM_DRAG],
+                    engine_variables[RAM_DRAG],
+                    engine_variables[GROSS_THRUST],
```
crecine, line 467: should these be renamed to `engine_variable_units`?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -962,19 +931,9 @@ def build_mission(self, num_nodes, aviary_inputs) -> om.Group:
...
-            # default value of zero at every flight condition
-            engine.add_output('thrust_net_max_unscaled',
-                              self.data[THRUST],
-                              units=units[THRUST],
-                              desc='Current max net thrust produced (unscaled)')
-
         # add created subsystems to engine_group
         outputs = []
-        if getattr(self, 'use_t4', False):
+        if self.use_t4:
```
crecine, line 936: it might be safer to keep getattr, or can we guarantee that `use_t4` will always be added?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1029,37 +988,56 @@ def build_mission(self, num_nodes, aviary_inputs) -> om.Group:
...
-        engine_group.add_subsystem('engine_scaling',
-                                   subsys=EngineScaling(num_nodes=num_nodes,
-                                                        aviary_options=self.options),
-                                   promotes_inputs=[
-                                       Aircraft.Engine.SCALE_FACTOR, Dynamic.Mission.MACH],
-                                   promotes_outputs=['*'])
+        engine_outputs = self.engine_variables.copy()
+        if SHAFT_POWER_CORRECTED in engine_outputs:
+            shp_units = engine_outputs.pop(SHAFT_POWER_CORRECTED)
+            engine_outputs[SHAFT_POWER] = shp_units
```
crecine, line 994: similar question here about renaming to `engine_output_units`

aviary/subsystems/propulsion/engine_scaling.py
```diff
@@ -2,10 +2,36 @@
...
+TEMPERATURE = EngineModelVariables.TEMPERATURE_T4
+# EXIT_AREA = EngineModelVariables.EXIT_AREA
+
+# these variables are not outputs or are variables that should not get scaled
+skip_variables = [MACH, ALTITUDE, THROTTLE, HYBRID_THROTTLE, TEMPERATURE]
+# these variables have a 'max' counterpart that should be added if they are present
+max_variables = {
+    THRUST: Dynamic.Mission.THRUST_MAX,
+    SHAFT_POWER: Dynamic.Mission.SHAFT_POWER_MAX,
+}
```
crecine, line 32: looks like these might be used in a couple places, should we define them in one place and import them in the others?

aviary/subsystems/propulsion/engine_scaling.py
```diff
@@ -130,107 +144,86 @@ def compute(self, inputs, outputs):
...
-
-        # outputs[Dynamic.Mission.EXIT_AREA] = unscaled_exit_area * scale_factor
+        # loop through all variables, singling out fuel flow to have special scaling
+        # compute 'max' counterpart of variables that have them
+        for variable in engine_variables:
+            if variable not in skip_variables:
+                if variable is FUEL_FLOW:
+                    outputs[Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE] = (
+                        -(inputs['fuel_flow_rate_unscaled'] * fuel_flow_scale_factor)
+                        - constant_fuel_flow
```
crecine, line 154: I think
`-(inputs['fuel_flow_rate_unscaled'] * fuel_flow_scale_factor + constant_fuel_flow)`
is a little clearer

aviary/subsystems/propulsion/engine_scaling.py
```diff
@@ -130,107 +144,86 @@ def compute(self, inputs, outputs):
...
+                        rows=r,
+                        cols=c,
+                        val=1.0,
+                    )
+                    self.declare_partials(
+                        Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE,
+                        'fuel_flow_rate_unscaled',
+                        rows=r,
+                        cols=r,
+                        val=1.0,
```
crecine, line 189: should this be -1?

crecine: or do we not need any of these `val=1`, since we will calculate it below?

aviary/subsystems/propulsion/motor/model/motor_mission.py
```diff
@@ -96,23 +97,34 @@ def setup(self):
...
+            ],
+        )
+
+        motor_group_max.add_subsystem(
+            'power_comp_max',
+            om.ExecComp(
+                'max_power = max_torque * pi * RPM / 30',
+                max_power={'val': np.ones(nn), 'units': 'kW'},
+                max_torque={'val': np.ones(nn), 'units': 'kN*m'},
+                RPM={'val': np.ones(nn), 'units': 'rpm'},
```
crecine, line 120: is it possible to use `rad/s` as the units and let OM handle the unit conversion?

aviary/subsystems/propulsion/motor/motor_builder.py
```diff
@@ -54,18 +54,18 @@ def build_mission(self, num_nodes, aviary_inputs):
...
+    #     Dynamic.Mission.Motor.TORQUE_CON: {
+    #         'upper': 0.0,
+    #         'type': 'path'
+    #     }
+    #     TBD Gearbox torque constraint
+    #         }
+    #     else:
+    #         constraints = {}
+
+    #     return constraints
```
crecine, line 68: remove?

aviary/subsystems/propulsion/propulsion_mission.py
```diff
@@ -94,28 +94,37 @@ def setup(self):
...
+            Dynamic.Mission.NOX_RATE, val=0, shape=(nn,), axis=1, units='lb/h'
+        )
+        perf_mux.add_var(
+            Dynamic.Mission.TEMPERATURE_T4, val=0, shape=(nn,), axis=1, units='degR'
+        )
+        perf_mux.add_var(
+            Dynamic.Mission.SHAFT_POWER, val=0, shape=(nn,), axis=1, units='hp'
+        )
+        perf_mux.add_var(
+            Dynamic.Mission.SHAFT_POWER_MAX, val=0, shape=(nn,), axis=1, units='hp'
```
crecine, line 126: should these be added based on the variables that are actually present in the engines, or are these all needed downstream of the mux (just not required from the engine model)

aviary/subsystems/propulsion/propulsion_mission.py
```diff
@@ -126,62 +135,89 @@ def setup(self):
...
+        comp_list = [self._get_subsystem(engine) for engine in engine_names]
+
+        # dictionaries of outputs for each engine in prop mission
+        output_dict = {}
+        # Dictionary of all unique inputs/outputs from all new components, keys are
+        # units for each var
+        unique_outputs = {}
+
+        # idx to be used for slicing inputs in next round of improvements
+        for idx, comp in enumerate(comp_list):
```
crecine, line 182: instead of using a list, which will require us to keep track of the order we added the engines in, would it be better to use a dictionary with the name as the key and use `_2`, `_3`... for engine models with the same name?

aviary/subsystems/propulsion/propulsion_mission.py
```diff
@@ -126,62 +135,89 @@ def setup(self):
...
+                    for key in promoted_outputs
+                ]
+            )
+
+        # add variables to the mux component and make connections to individual
+        # component outputs
+        # if num_engine_type > 1:
+        for output in unique_outputs:
+            if output in supported_outputs:
+                # self.vectorize_performance.add_var(output, units=unique_outputs[output])
```
crecine, line 210: remove?

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -104,9 +93,72 @@ def build_pre_mission(self, aviary_inputs, **kwargs) -> om.Group:
...
+                    aviary_options=aviary_inputs,
+                )
+
+        # turboprop_group.set_input_defaults(Aircraft.Engine.PROPELLER_ACTIVITY_FACTOR, val=0.0)
+        turboprop_group.set_input_default(
+            Aircraft.Engine.PROPELLER_TIP_SPEED_MAX, val=0.0, units='ft/s'
+        )
+        # turboprop_group.set_input_defaults(Dynamic.Mission.SPEED_OF_SOUND, units='ft/s', val=0.0)
+        # turboprop_group.set_input_defaults(Dynamic.Mission.VELOCITY, units='ft/s', val=0.0)
+        # turboprop_group.set_input_defaults('velocity', units='ft/s', val=0.0)
```
crecine, line 138: remove?

aviary/variable_info/variable_meta_data.py
```diff
@@ -6604,28 +6604,6 @@
     desc='The maximum possible shaft power currently producible, per engine'
 )
 
-add_meta_data(
-    Dynamic.Mission.SHAFT_POWER_CORRECTED,
```
crecine, line 6608: why were these removed?

aviary/subsystems/propulsion/propulsion_mission.py
jkirk5: I'm not sure I follow, a dictionary won't solve any ordering problems (idx will be needed for `om.slicer`).
Order shouldn't be a problem because the list of engine models passed through options will never be changed (engine model list is saved as an attribute of the propulsion builder and never gets modified once initialized).

aviary/subsystems/propulsion/engine_deck.py
jkirk5: I don't think so - the point of the variable wasn't to store units specifically, but keep track of what variables are getting used by the engine. I used a dict as a convenient way to also track what units the data file asked for.

Basically questions like "does the engine use this variable" are cleanly communicated by `if XXX in engine_variables`, and would be less clear with a name change.

aviary/subsystems/propulsion/engine_deck.py
jkirk5: I remember getattr was causing some issue but I went back and checked and it seems like it is ok now - reverted that change

aviary/subsystems/propulsion/propulsion_mission.py
jkirk5: I chose to keep all the variables present post-muxing so there won't be any problems with anything downstream asking for these values (like the fuel summation comp for an all-electric aircraft). Thrust is also definitely used by the EOMs too. For the most part I don't think most of these have to be always present, but it also doesn't hurt to just have them available and defaulted to zero just in case someone eventually does want them.

aviary/subsystems/propulsion/turboprop_model.py
jkirk5: yep, I should remove this now

aviary/subsystems/propulsion/engine_scaling.py
jkirk5: I forgot these got reused! I will move them to the propulsion utils

aviary/subsystems/propulsion/motor/model/motor_mission.py
jkirk5: Good idea, I'll try that out

aviary/subsystems/propulsion/engine_scaling.py
jkirk5: Looks like we didn't need to specify val for any of these, now removed

aviary/subsystems/propulsion/motor/motor_builder.py
jkirk5: Since we're going to eventually need this code for the separate gearbox component I'm keeping this in until that happens

aviary/subsystems/propulsion/engine_deck.py
Kenneth-T-Moore: Calling attention to this edit again. I recall that there was an earlier PR that changed the names and definitions of these variables. I have some questions about the new ones, but this might be a merge mistake.

jkirk5: I've reverted it back, should be good now

Kenneth-T-Moore (approved):

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -129,13 +129,9 @@ def build_post_mission(self, aviary_inputs, **kwargs):
...
-        # turboprop_group.set_input_defaults(Aircraft.Engine.PROPELLER_ACTIVITY_FACTOR, val=0.0)
-        turboprop_group.set_input_default(
-            Aircraft.Engine.PROPELLER_TIP_SPEED_MAX, val=0.0, units='ft/s'
-        )
-        # turboprop_group.set_input_defaults(Dynamic.Mission.SPEED_OF_SOUND, units='ft/s', val=0.0)
-        # turboprop_group.set_input_defaults(Dynamic.Mission.VELOCITY, units='ft/s', val=0.0)
-        # turboprop_group.set_input_defaults('velocity', units='ft/s', val=0.0)
+        # turboprop_group.set_input_default(
+        #     Aircraft.Engine.PROPELLER_TIP_SPEED_MAX, val=0.0, units='ft/s'
+        # )
```
crecine, line 134: remove?

crecine (approved):

PR #462, opened 2024-07-26
User Interface for Modeling Aircraft

Dawson-Manning:
### Summary

This commit adds a interface for users to easily create and edit models to use for optimization. Interface includes names, units, and descriptions of each variable and sorts them into four tabs for Aircraft, Dynamic, Mission, Setting and sorts each variable name into containers based on the subheading "wing", "engine", etc. Additionally, there is an added search feature which allows for users to quicky search a variable name to be edited.  Once finished with entering inputs, the user's inputs and remaining entries, which autofill to default values, are written to a csv file in a format that can be used by Aviary. Image of interface attached below.
Screenshot 2024-07-26 155255

All information is pulled from Aviary>aviary>variable_info>variable_meta_data.py and Aviary>aviary>variable_info>variables.py thus it automatically updates and adds variables to the interface when variables are added to Aviary's meta data.

Additional edits were made to Aviary>aviary>variable_info>variables.py to keep formatting consistent with other variables. 

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/build_model.py
```diff
@@ -0,0 +1,692 @@
...
+            entries_per_subhead[-1] += 1
+        else:
+            entries_per_subhead.append(1)
+            name_each_subhead.append(subhead)
+            headers.append(header)
+        old_subhead = subhead
+    else:
+        entries_per_subhead.append(1)
+        name_each_subhead.append(key)
+        headers.append('dynamic')
```
johnjasa, line 40: Dynamic variables (all those that are `Dynamic.Mission.<variable_name>` are not actually used in the .csv file. Because of this, I don't think they need to be included in this GUI.

There could be value in having the dynamic variables present in the GUI and you then output those values as initial guesses used in the `phase_info` object, but I think that capability is already represented by the `draw_mission` tool. So, I'd suggest removing the dynamic variables from this GUI unless I'm missing a reason (which I easily could be!)

johnjasa (commented):
Thanks for this, Dawson! I tried it out locally and it worked first try. I like the collapsible headings for each subsystem.

I added some small comments and I'll put some general ones here:

- could you please add a command mapping for this utility so that someone could do `aviary build_model` (or whatever name you choose) in the command line, like the other utilities Aviary has? There's a little bit of convoluted code to do this, but you can follow the [other utilities' logic here](https://github.com/OpenMDAO/Aviary/blob/main/aviary/interface/cmd_entry_points.py#L45) or I'd be happy to walk through this process with you
- when you hit `Submit` could you make some sort of indication that a file was saved? I hit submit and initially thought it didn't work because I didn't see a confirmation or similar message, though a file was successfully created. This notification could be a message in the command-line window, a pop-up in the GUI, or something else. I think it'd be useful to let the user know that a file was made explicitly

Feel free to disregard any of these suggestions, I just wanted to weigh in. Thanks for the nice work!

Dawson-Manning:
With new commits, the gui can now be ran using aviary command <aviary build_model>. Additionally, all Dynamic.Mission.**vars were removed along with the associated tab. The gui will now indicate to the user via command prompt when a file has been saved with "Summit", "File->Save", and "File->Save as...", The gui will now indicate to the user via command prompt when a file has been opened with "File->Open" and "File->Open & Display".

aviary/interface/build_model.py
```diff
@@ -0,0 +1,692 @@
+import csv
+import math
+from tkinter import *
+import tkinter as tk
+import tkinter.ttk as ttk
+from tkinter import messagebox
+from tkinter import filedialog
```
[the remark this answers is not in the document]
jkirk5: I think the cleanest is just to have the 2 aliased imports (tk, ttk), and use those to call all needed functions. The two functions specifically imported here don't seem to be used more frequently than other tkinter functions, so to me there isn't a clear reason to specially import those.

I'll weigh in that we shouldn't globally import here. That's something we should only do if there is a very good reason to specifically do that.

johnjasa (approved):

aviary/interface/cmd_entry_points.py
```diff
@@ -97,10 +100,11 @@ def aviary_cmd():
     cmdargs = [a for a in sys.argv[1:] if a not in ('-h',)]
 
     if len(args) == 1 and len(user_args) == 0:
-        if args[0] != 'draw_mission' and args[0] != 'run_mission':
+        if args[0] != 'draw_mission' and args[0] != 'run_mission' and args[0] != 'build_model':
```
crecine, line 103:
```suggestion
        if args[0] not in ('draw_mission', 'run_mission', 'build_model'):
```

aviary/variable_info/variables.py
```diff
@@ -155,7 +155,7 @@ class Design:
         FINENESS = 'aircraft:design:fineness'
         FIXED_EQUIPMENT_MASS = 'aircraft:design:fixed_equipment_mass'
         FIXED_USEFUL_LOAD = 'aircraft:design:fixed_useful_load'
-        IJEFF = 'ijeff'
+        IJEFF = 'aircraft:design:ijeff'
```
crecine, line 158: This one is intentional
```suggestion
        IJEFF = 'ijeff'
```

crecine:
There are some variables that have different defaults for GASP vs FLOPS (see utils/legacy_code_data). It might be a good idea to allow the user to load the defaults from one of those before they start making changes.

aviary/variable_info/variables.py
Dawson-Manning: @crecine 
Without this modification "ieff" becomes a subheading and since this variable is in the middle of the "design" section it takes the rest of the design variables into its container. Would it be better to move this variable to later in the variable list or is there a better option?

Dawson-Manning:
> There are some variables that have different defaults for GASP vs FLOPS (see utils/legacy_code_data). It might be a good idea to allow the user to load the defaults from one of those before they start making changes.
@crecine 
I am able to do this, however since there are far more default values for flops the resulting csv file would be a mixture of FLOPs and GASP default values as this interface prints all lines from the variable meta data except dynamic variables. Is this an issue???

aviary/variable_info/variables.py
crecine: if we need this change to prevent that issue, that's fine. But since we don't need to set this value, excluding it from the GUI would also work

crecine:
> @crecine
> I am able to do this, however since there are far more default values for flops the resulting csv file would be a mixture of FLOPs and GASP default values as this interface prints all lines from the variable meta data except dynamic variables. Is this an issue???

I meant adding a button (or something else) to allow the user to load one of the legacy defaults after it reads the variable_meta_data

aviary/interface/build_model.py
xjjiang, line 1: This file needs docstring.

hschilling:
Thanks, Dawson!

In the search tab, if I put something in the search box, do a search, and then press Clear, the results go away as they should but the search box still has what I typed into it. Should it? I think most search interfaces clear the search text box also.

hschilling:
When you do a Submit, the text in the terminal has a typo

```
Model successfully Seved to Aircraft_Model.csv
```

hschilling:
This field looks like it might be hard to fill out. Can a popup menu be used for values that have a limited number of options?

image

hschilling:
Not sure what is different from when I tried the Search and commented about it, but now when I do searches, nothing shows up. For example, searching for "wing" results in no results. 

image

Issue #294, opened 2024-05-29
Replace ParamPort with VariableIn


johnjasa:
### Desired capability or behavior.

The GASPy-based code currently uses a somewhat ad hoc mechanism for statically declaring the variables that need to be passed either from the user or from static analysis into the "phases" (referring to actual dymos phases and the analytic flight segments). It then functions as an input-only component in each ODE for other components to connect to via promotion and provides methods to resolve conflicting defaults and add them as parameters to the dymos trajectories.

Talked this through a bit with @crecine and @ehariton; we like the idea of naming the VariablesIn components when they're in the model to something like _dummy_component_for_inputs or something equally obvious what's going on.

Additionally, these components should move as far down in the model as possible to not create unnecessary backwards coupling from appearing in the n2.

jkirk5:
The chosen path forward on this issue is for VariablesIn and ParamPort to be completely removed

PR #478, opened 2024-08-07
Paramport refactor part 1: Remove need for VariablesIn for height-energy mission.

Kenneth-T-Moore:
### Summary

This PR removes the `VariablesIn` component from the height-energy mission. This component was formerly needed because we were adding parameters from a curated list.  The new approach moves the responsibility for declaring those parameters into the aerodynamics builder, because those variables are introduced by components in the default aero subsystem.

1. Aero builder now declares all parameters it uses in height-energy.
2. `VariablesIn` component has been removed.

This caused some challenges in models that use the level 3 interface, and existing models will need to be updated. Anyone running level 3 may need to manually set_input_defaults on some variables in their top level script.  A new helper function is introduced to aid in this.

3. New method `set_aviary_input_defaults` for level 3.  Requires a list of aviary variables as input.  Replaces the old `set_aviary_initial_values` and the "input_sink" `VariablesIn`, which are no longer needed.
4. New method `set_aviary_initial_values` (repurposing the old name, which is more accurate), which is called after setup and loops over the aviary_inputs, setting all the values. This is needed because we don't loop over the dummy "input_sink" anymore.
5. Docs updated to show these.

TODO: Next up, need to apply this to the gasp counterpart `ParamPort`.

### Related Issues

- Resolves # 294

None

### New Dependencies

None


xjjiang:
In Issue #294, John pointed out the desire of replacing `ParamPort` with `VariableIn`. Is it still a desire?

Kenneth-T-Moore:
> In Issue #294, John pointed out the desire of replacing `ParamPort` with `VariableIn`. Is it still a desire?

I think it is even more desirable to remove the paramport entirely, but we weren't sure it was possible then.  GIven that it works on the FLOPS side, I think it should also work on the gasp side.

aviary/mission/flops_based/phases/detailed_landing_phases.py
```diff
@@ -1248,7 +1251,33 @@ def build_trajectory(
...
+            # They mostly come from the low-speed aero subsystem.
+
+            aero = CoreAerodynamicsBuilder('core_aerodynamics',
+                                           BaseMetaData,
+                                           LegacyCode('FLOPS'))
+
+            phase_info = {}
+            phase_info['subsystem_options'] = {}
+            phase_info['subsystem_options']['core_aerodynamics'] = {}
+            phase_info['subsystem_options']['core_aerodynamics']['method'] = 'low_speed'
```
crecine, line 1265:
```suggestion
            phase_info = {'subsystem_options':{'core_aerodynamics':{'method':'low_speed'}}}
```

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -339,14 +340,106 @@ def get_parameters(self, aviary_inputs=None, phase_info=None):
...
             if method == 'computed':
-                param_vars = [Aircraft.Nacelle.CHARACTERISTIC_LENGTH,
-                              Aircraft.Nacelle.FINENESS,
-                              Aircraft.Nacelle.LAMINAR_FLOW_LOWER,
-                              Aircraft.Nacelle.LAMINAR_FLOW_UPPER,
-                              Aircraft.Nacelle.WETTED_AREA]
-                for var in param_vars:
+
+                # Parameters for drag computation.
+                core_inputs = [
```
crecine, line 345: would it make sense to define these lists further up/out, so they can be imported by other files?

crecine (approved):

aviary/variable_info/functions.py
```diff
@@ -166,9 +165,9 @@ def setup_trajectory_params(
     are being used in the trajectory, and for the variables which are
     not options it adds them as a parameter of the trajectory.
     """
-    # TODO: variables_to_add might be an unused option.
+    # TODO: variables_to_add is required, so should be an arg, not a kwarg.
```
jkirk5, line 168: Is it guaranteed to always be required, even in the case of all custom subsystems? Also, can we default it to an empty list and eliminate this code block?

jkirk5 (approved):

aviary/variable_info/functions.py
Kenneth-T-Moore: It's more that every time we call it right now in Aviary, we give it that argument.

PR #487, opened 2024-08-15
Aviary plot_drag_polar GUI using a csv input file

tahaitani:
### Summary

Summary of PR.
This commit adds a new feature to the aviary command window which allows the user to plot drag polar data. The user can  select the x-axis and y-axis values to be either : CL,CD, alpha or CL/CD and fix either mach or altitude to a specific value using data from the inputted drag polar data file. A sample polar data file based on the n3cc model was also added in this PR. 

The user has to run: "aviary plot_drag_polar "

This will prompt the "file open" feature and the user will have to select a csv file with drag polar data. After a properly formatted file is selected, the plotting window will open and the user can select their axes and fixed variable and value. Below is an image of a CL vs CD graph for a Mach of 0.2 and varying altitude based on data obtained from the n3cc model.
GUI_sample

- Resolves #

None

### New Dependencies

None


aviary/interface/plot_drag_polar.py
```diff
@@ -0,0 +1,206 @@
...
+            fix_value_var.set(str(altitude_values[0]))
+
+    set_x_var = StringVar(value='CD')
+
+    set_x_label = Label(master=window, text="x-axis")
+    set_x_label.pack()
+    set_x_combobox = Combobox(master=window, textvariable=set_x_var, values=[
+        'CD', 'CL', 'Alpha', 'CL/CD'])
+
+    set_x_combobox.pack()
```
crecine, line 149: could you use `side = LEFT` for all of these so that these all fit on one line instead of stacking vertically?
This will give more space for the plot

aviary/interface/plot_drag_polar.py
```diff
@@ -0,0 +1,206 @@
...
+    fix_variable_combobox.current(0)
+    fix_value_label = Label(master=window, text="Fix Value:")
+    fix_value_label.pack()
+
+    fix_value_combobox = Combobox(master=window, textvariable=fix_value_var)
+    fix_value_combobox.pack()
+    update_fix_value_combobox()
+
+    plot_button = Button(master=window, text="Plot", command=update_plot)
+    plot_button.pack()
```
crecine, line 174: Maybe use `side = RIGHT` for this one

aviary/interface/plot_drag_polar.py
tahaitani: Fixed and added spacing between comboboxes

aviary/interface/plot_drag_polar.py
tahaitani: Fixed.

Kenneth-T-Moore (approved):
I think this works pretty well. It has some odd behavior on the mac that we will have to address later.

aviary/interface/plot_drag_polar.py
```diff
@@ -0,0 +1,207 @@
...
+from matplotlib.backends.backend_tkagg import (FigureCanvasTkAgg, NavigationToolbar2Tk)
+import tkinter as tk
+from aviary.utils.functions import get_path
+
+
+def plot_drag_polar(input_file=None):
+    if input_file is None:
+        input_file = filedialog.askopenfilename(filetypes=[("CSV files", "*.csv")])
+        if not input_file:
+            messagebox.showerror(
```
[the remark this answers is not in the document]
ehariton: Do you want me to pull out the example file and make this error exit() ?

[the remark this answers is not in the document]
jkirk5: Yeah that sounds good, exit w/ error message

jkirk5 (approved):

PR #488, opened 2024-08-15
Level 2 Off-Design

kaushikponnapalli:
### Summary

Created a level 2 interface to create and run fallout and alternate missions automatically.

### Related Issues

- Resolves #

None

### New Dependencies

json


xjjiang:
Is it better to change `save_json` to something like `save_aviary_to_json`? It is more meaningful and we may have a more general function to save json files.

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
 
         self.problem_ran_successfully = not failed
 
+    def run_alternate_mission(self, json_filename='sizing_problem.json',
+                              payload_mass=None, mission_range=None,
+                              phase_info=None, optimizer=None, max_iter=50,
+                              verbosity=Verbosity.BRIEF):
+        """
+        This function runs a fallout mission based on a sizing mission output.
```
Kenneth-T-Moore, line 2335: The docstring is cut-and-paste from sizing.

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
...
+        prob_alternate.link_phases()
+        prob_alternate.add_driver(optimizer, max_iter=max_iter, verbosity=verbosity)
+        prob_alternate.add_design_variables()
+        prob_alternate.add_objective()
+        prob_alternate.setup()
+        prob_alternate.set_initial_guesses()
+        prob_alternate.run_aviary_problem(record_filename='alternate_problem_history.db')
+        return prob_alternate
+
+    def run_fallout_mission(self, json_filename='sizing_problem.json',
```
Kenneth-T-Moore, line 2385: General api thoughts -- do we want a separate function for each offdesign mission type, or do we want a single function with an mission type attribute? Asking mostly with the consideration of future types (if there are any others).

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
...
+        prob_fallout.add_pre_mission_systems()
+        prob_fallout.add_phases()
+        prob_fallout.add_post_mission_systems()
+        prob_fallout.link_phases()
+        prob_fallout.add_driver(optimizer, max_iter=max_iter, verbosity=verbosity)
+        prob_fallout.add_design_variables()
+        prob_fallout.add_objective()
+        prob_fallout.setup()
+        prob_fallout.set_initial_guesses()
+        prob_fallout.run_aviary_problem(record_filename='fallout_problem_history.db')
```
Kenneth-T-Moore, line 2436: I wonder if we should split this into 2 functions, where the first one "sets up" the model, doing everything up to, but not including, run_aviary_problem.  And then a second function to run the problem.  This is to allow the user to do any extra sets or to change optimizer settings. 

Note, I don't have a specific usecase for this -- just thinking that maybe some other variables might be different in off-design, particularly when we include external subsystems.

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
...
+            Aviary problem object optimized for the aircraft design/sizing mission.
+            Assumed to contain aviary_inputs and Mission.Summary.GROSS_MASS
+        json_filename:   string
+            User specified name and relative path of json file to save the data into.
+        """
+
+        aviary_input_list = []
+        with open(json_filename, 'w') as jsonfile:
+            # Loop through aviary input datastructure and create a list
+            for data in self.aviary_inputs:
```
Kenneth-T-Moore, line 2455: Since this iterates on aviary_inputs, we might need to double check that we aren't using `pop` on any keys while building the level2 model. I think we have gotten rid of most of them.

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
...
+                (name, (value, units)) = data
+                type_value = type(value)
+
+                # Get the gross mass value from the sizing problem and add it to input list
+                if name == Mission.Summary.GROSS_MASS or name == Mission.Design.GROSS_MASS:
+                    Mission_Summary_GROSS_MASS_val = self.get_val(
```
Kenneth-T-Moore, line 2461: So, the gross mass is the only variable we take from the model. Is there going to be support (in the future maybe) for more design variables in the design run? I think you might be able to curate a list from the design variables assigned in the original design problem, after removing any dynamic variables added for the trajectories.

jkirk5 (approved):
Small suggestions but nothing blocking approval

xjjiang:
Will there be document update?

aviary/interface/methods_for_level2.py
Kenneth-T-Moore: We have to be careful of copying the aviary_inputs, as it might contain things like VSP models for external subsystems.

aviary/interface/methods_for_level2.py
```diff
@@ -2325,6 +2327,169 @@ def run_aviary_problem(self,
...
+            Max number of iterations allowed for the optimizer.
+        verbosity : Verbosity or list, optional
+            If Verbosity.DEBUG, debug print options ['desvars','ln_cons','nl_cons','objs'] will be set.
+            If a list is provided, it will be used as the debug print options.
+        """
+        if phase_info is None:
+            phase_info = self.phase_info
+        if mission_range is None:
+            design_range = self.get_val(Mission.Design.RANGE)
+        if payload_mass is None:
```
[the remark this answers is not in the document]
cmbenne3: I like the idea of this check, but I think we need to check against the Specified Maximum Payload (i.e. PAYLMX in FLOPS), which is not necessarily the payload mass carried on the sizing mission?

Kenneth-T-Moore (approved):
I think my comments have at least been discussed, and we are looking forward to more evaluation from MBSAE.

aviary/interface/methods_for_level2.py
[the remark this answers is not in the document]
jkirk5: Yeah that sounds right. As we develop the feature further we can think about adding more checks

PR #492, opened 2024-08-20
GASP engine deck Convertion modification

### LINKED ISSUE #345: Reading GASP engine data with 3 Mach numbers

If you run `aviary convert_engine -f GASP CFM_56_7B27.eng`, you will get `RuntimeWarning: divide by zero encountered`. This is because this GASP engine file has the following table:

3   11       0.0
              0.0000    0.2000    0.3500
    3.0000  119.3320   14.3682  -95.7382
    4.0000 5986.5659 4426.9868 4101.6514
...

Aviary uses `3d-lagrange3 interpolation` which requires at least four points.

We can either automatically switch to linear interpolation or have a pre-process to `CFM_56_7B27.eng` by adding another point (say Mach 0.5) to it along with each T4/T2 value. 

BTW the format  for the data is F10.4 to read off the values. a lot of the more recent engine decks use data to one decimal point for the data. Really does not make a lot of sense to know the values of the thrust to the fourth decimal point.

_No response_

### Associated Bug Report

_No response_

xjjiang:
### Summary

When users convert a GASP engine deck to Aviary format, Aviary uses `lagrange3` to mimic GASP, but some engine decks may not have enough points per dimension. If there are only three points, switch to `lagrange2`. If there are only two points, switch to `slinear`. 

For `GASP` engine deck, try to provide at least 4 Mach numbers.
For `GASP_TP` engine deck, try to provide at least 4 advanced ratios

### Related Issues

- Resolves #345 

None

### New Dependencies

None


aviary/utils/engine_deck_conversion.py
```diff
@@ -501,6 +501,13 @@ def _make_structured_grid(data, method="lagrange3", fields=["thrust", "fuelflow"
 
             # would explicitly use lagrange3 here to mimic GASP, but some engine
             # decks may not have enough points per dimension
+            # For GASP engine deck, try to provide at least 4 Mach numbers.
+            # For GASP_TP engine deck, try to provide at least 4 advanced ratios
```
jkirk5, line 505:
```suggestion
            # For GASP_TP engine deck, try to provide at least 4 Mach numbers
```

advance ratio isn't a variable in the GASP turboshaft decks, we're still checking Mach for number of points here

jkirk5 (approved):

aviary/utils/engine_deck_conversion.py
xjjiang: What I meant was the `GASP_TP` engine deck before it was converted to Aviary engine deck. In GASP, it is `advanced ratio`, not `Mach number`.

Kenneth-T-Moore (approved):

PR #510, opened 2024-08-29
Wetted area got a huge number because multiplication factor is 4000

xjjiang:
### Summary
set `Aircraft.Fuselage.WETTED_AREA_FACTOR = 1.0` because `Aircraft.Fuselage.PROVIDE_SURFACE_AREA = False`. It was set to `4000`.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):
Probably makes sense to have reasonable values in the tests. The original test writer may have forgotten to toggle the PROVIDE_SURFACE_AREA flag.

jkirk5 (changes requested):
Can we change the scale factor to a value that isn't 1.0? That way the test is also making sure we're actually multiplying the scaler correctly

xjjiang:
Changed one of `Aircraft.Fuselage.WETTED_AREA_FACTOR` to 0.5 from 1.0 to verify it works.

xjjiang:
This feature will be covered by another PR. So, I am closing it.

Issue #351, opened 2024-06-26
Regional turboprop model completion


jkirk5:
Create a model for the large regional turboprop that contains the following features:

1. Validated results against base GASP case. Weights and aerodynamics should match unless reasonable deviations from baseline model made
2. Includes the turboprop propulsion model using Hamilton Standard propeller performance modeling
3. Models a design and off-design mission, optionally with a payload-range diagram
4. Create a hybrid-electric variant that utilizes electroprops and a battery

PR #514, opened 2024-09-03
Turboprop with gearbox integration

jkirk5:
### Summary

Overhauls the turboprop model to natively accept a gearbox model. All connections and promotions within the turboprop are dynamically set up in the `configure` step with many logic paths for various combinations of with/without gearbox, whether that variable should be promoted or not based on presence of gearbox variables that get "updated" (e.g. shaft power in and shaft power out). This should work for any combination of built-in and custom motor, gearbox, and propeller models - all the right connections will be dynamically made and the right set of outputs will be promoted outside the turboprop group so other components can find them.

Gearbox variables got genericized to a `X_in` and `X_out` format for clarity and to be more intuitive when the gearbox is used in other applications. Variables like `Dynamic.Mission.X_GEARBOX` got removed from the hierarchy due to difficulty handling them in a generic way in the configure call - I am currently adding "_gearbox" to ALL input variables in a gearbox model that get changed (using the "X_in" and "X_out pattern). It might be possible to individually check if variable names exist in the hierarchy, and if so, instead alias to the gearbox version using the hierarchy itself, but right now I couldn't get that working.

The turboprop freighter model was included as a test case actually using the turboprop model in a full benchmark. Many additional fixes to various components were made to address errors that model encountered.

### Related Issues

Progress towards #351 

None

### New Dependencies

None


aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -286,37 +337,260 @@ def setup(self):
...
+        Correctly connect variables between shaft power model, gearbox, and propeller,
+        aliasing names if they are present in both sets of connections.
+
+        If a gearbox is present, inputs to the gearbox are usually done via connection,
+        while outputs from the gearbox are promoted. This prevents intermediate values
+        from "leaking" out of the model and getting incorrectly connected to outside
+        components. It is assumed only the gearbox has variables like this.
+
+        Set up fixed RPM value if requested by user, which overrides any RPM defined by
+        shaft powerm model
```
crecine, line 358:
```suggestion
        shaft power model
```

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -286,37 +337,260 @@ def setup(self):
...
+            shp_output_list.remove(Dynamic.Mission.THRUST_MAX)
+
+        # Gearbox connections
+        if has_gearbox:
+            for var in shp_output_list.copy():
+                # Check for case: var is output from shp_model, connects to gearbox, then
+                #   gets updated by gearbox
+                # RPM has special handling, so skip it here
+                if (
+                    var in shp_output_list
```
crecine, line 471: this might be unnecessary, since we are using `for var in shp_output_list`

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -286,37 +337,260 @@ def setup(self):
...
+        # GEARBOX MODEL CONNECTIONS #
+        #############################
+        if has_gearbox:
+            # Promote all inputs which don't come from shp model (those got connected),
+            #   don't promote ones in skip list
+            for var in gearbox_input_list.copy():
+                if var not in skipped_inputs:
+                    gearbox_inputs.append(var)
+                # DO NOT promote inputs in skip list - always skip
+                gearbox_input_list.remove(var)
```
crecine, line 543: should this be in an `else`? it is currently applied to every variable in gearbox_input_list

aviary/subsystems/propulsion/gearbox/model/gearbox_mission.py
```diff
@@ -21,65 +23,103 @@ def initialize(self):
...
+            'rpm_comp',
+            om.ExecComp(
+                'rpm_out = rpm_in / gear_ratio',
+                rpm_out={'val': np.ones(n), 'units': 'rpm'},
+                gear_ratio={'val': 1.0, 'units': 'unitless'},
+                rpm_in={'val': np.ones(n), 'units': 'rpm'},
+                has_diag_partials=True,
+            ),
+            promotes_inputs=[
+                ('rpm_in', Dynamic.Mission.RPM + '_in'),
```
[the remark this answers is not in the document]
crecine: I think doing this will also allow you to simplify a lot of the aliasing and special handling you do below.

aviary/utils/engine_deck_conversion.py
```diff
@@ -256,26 +256,26 @@ def EngineDeckConverter(input_file, output_file, data_format: EngineDeckType):
...
 
-            # remove all points above T4max
-            # TODO save these points as commented out?
-            valid_idx = np.where(data[THROTTLE] <= 1.0)
-            data[MACH] = data[MACH][valid_idx]
-            data[ALTITUDE] = data[ALTITUDE][valid_idx]
-            data[THROTTLE] = data[THROTTLE][valid_idx]
-            data[FUEL_FLOW] = data[FUEL_FLOW][valid_idx]
+        else:
+            data[THROTTLE] = normalize(T4T2, minimum=0.0, maximum=t4max)
```
crecine, line 264: is t4max the right value to use as the maximum here? Is it actually the max t4/t2 if we are in this branch of the the `if`?

aviary/variable_info/variable_meta_data.py
```diff
@@ -2116,12 +2127,13 @@
...
-                     "LEAPS1": None
-                     },
+    historical_name={
+        "GASP": None,  # TODO this needs verification
+        "FLOPS": None,
+        "LEAPS1": None,
+    },
     units='unitless',
-    desc='maximum allowable Mach number at propeller tip (based on helical speed)',
+    desc='maximum allowable Mach number at propeller tip (based on helical airspeed)',
```
crecine, line 2136: I believe this should just be "speed", because the helical speed is a combination of the airspeed and the rotational speed

aviary/variable_info/variable_meta_data.py
```diff
@@ -2414,14 +2426,15 @@
 add_meta_data(
     Aircraft.Engine.Gearbox.SHAFT_POWER_DESIGN,
     meta_data=_MetaData,
-    historical_name={"GASP": 'INPROP.HPMSLS',  # max sea level static horsepower, hp
-                     "FLOPS": None,
-                     "LEAPS1": None,
-                     },
+    historical_name={
+        "GASP": 'INPROP.HPMSLS',  # max sea level static horsepower, hp
```
crecine, line 2430: the gasp variable doesn't have the same units as the aviary variable I think this will cause problems in the fortran to aviary converter

aviary/subsystems/propulsion/propeller/propeller_performance.py
```diff
@@ -425,11 +428,11 @@ class PropellerPerformance(om.Group):
 
     def initialize(self):
         self.options.declare(
-            'num_nodes', types=int, default=1,
-            desc='Number of nodes to be evaluated in the RHS')
-        self.options.declare(
-            'input_rpm', types=bool, default=False,
-            desc='If True, the input is RPM, otherwise RPM is set by propeller limits')
```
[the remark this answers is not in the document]
jkirk5: Eliot and I are completely reworking how RPM is handled - the propeller model no longer picks an RPM, so essentially we are always using a version of `input_rpm`

RPM is always an input to the prop model. It either comes from the engine/gearbox/somewhere else in the model, is set to a fixed value by the user, or the optimizer picks a value (this part is WIP, will need to be a follow-on)

aviary/variable_info/variable_meta_data.py
jkirk5: Will update for clarity

aviary/subsystems/propulsion/turboprop_model.py
jkirk5: Yep I think you're right - removing

aviary/subsystems/propulsion/turboprop_model.py
jkirk5: I was taking a shortcut here, to be more verbose what I'm actually doing is this:
```
if var not in skipped_inputs:
    gearbox_inputs.append(var)
    gearbox_input_list.remove(var)
else:
    gearbox_input_list.remove(var)
```
 but figured that was redundant. The `gearbox_input_list` is tracking all the inputs I have yet to handle, either through promotion or connection. I end up handling every variable in the list here (either marking for promotion, or making sure it is entirely skipped), so deleting everything from `gearbox_input_list` was intended.

aviary/utils/engine_deck_conversion.py
jkirk5: What's happening here is if we are not calculating T4, that means that the T4T2 values are actually throttles. And based on the GASP engine decks we have in Aviary, in that scenario the t4max values correspond to max throttle (like in `turbofan_24k_2.deck`, the recovered t4max is 50.0 which perfectly lines up with max throttle)

I guess the normalization here is actually optional (Aviary will do it anyway after loading the data), so you can make the call if we want converted engines to maintain the original throttle values for easier comparison with GASP decks, or convert for 0-1 for easier comparisons with Aviary mission results

aviary/mission/gasp_based/phases/landing_components.py
```diff
@@ -28,7 +28,9 @@ def compute_partials(self, inputs, J):
 
 class GlideConditionComponent(om.ExplicitComponent):
     def setup(self):
-        self.add_input("rho_app", val=0.0, units="slug/ft**3", desc="air density")
+        self.add_input(
+            Dynamic.Mission.DENSITY, val=0.0, units="slug/ft**3", desc="air density"
```
ehariton, line 32: I was surprised to see the 'rho_app' rename in this PR.

jkirk5: There was actually an issue where the turboprop needs density (and the old engine models did not). Because this component renames most of the atmosphere outputs, downstream components actually can't access any of them. There appears to be no actual need for these aliases, and it caused problems, so out it goes

jkirk5: Can't just alias propulsion inputs because engine decks don't use density, so that would throw an error for those cases

aviary/models/large_turboprop_freighter/phase_info.py
```diff
@@ -0,0 +1,362 @@
...
+#             "initial_bounds": ((100, 361.5), "min"),
+#             "duration_bounds": ((29.0, 87.0), "min"),
+#         },
+#     },
+#     "post_mission": {
+#         "include_landing": False,
+#         "constrain_range": True,
+#         "target_range": (2_020., "nmi"),
+#     },
+# }
```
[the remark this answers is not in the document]
jkirk5: That's the energy method phase info for the mission, I'm keeping it for now so we can swap to it if needed

aviary/models/large_turboprop_freighter/large_turboprop_freighter.csv
```diff
@@ -0,0 +1,249 @@
...
+aircraft:wing:surface_ctrl_mass_coefficient, 0.588, unitless
+aircraft:wing:sweep, 0, deg
+aircraft:wing:taper_ratio, 0.52, unitless
+aircraft:wing:thickness_to_chord_root, 0.18, unitless
+aircraft:wing:thickness_to_chord_tip, 0.12, unitless
+aircraft:wing:zero_lift_angle, -1, deg
+mission:design:cruise_altitude, 21000, ft
+mission:design:gross_mass, 155000, lbm
+mission:design:mach, 0.475, unitless
+mission:design:range, 0, NM
```
ehariton, line 149: with design range = 0 this mission isn't going to fly anywhere

aviary/subsystems/propulsion/gearbox/model/gearbox_premission.py
```diff
@@ -44,18 +44,25 @@ def setup(self):
...
-        self.add_subsystem('mass_comp',
-                           om.ExecComp('gearbox_mass = torque_max / specific_torque',
-                                       gearbox_mass={'val': 0.0, 'units': 'kg'},
-                                       torque_max={'val': 0.0, 'units': 'N*m'},
-                                       specific_torque={'val': 0.0, 'units': 'N*m/kg'},
-                                       has_diag_partials=True),
-                           promotes_inputs=['torque_max',
-                                            ('specific_torque', Aircraft.Engine.Gearbox.SPECIFIC_TORQUE)],
-                           promotes_outputs=[('gearbox_mass', Aircraft.Engine.Gearbox.MASS)])
+        if self.options["simple_mass"]:
```
ehariton, line 47: It does make sense to create this as an option. Thanks for putting that in.

aviary/subsystems/propulsion/gearbox/test/test_gearbox.py
```diff
@@ -53,19 +53,26 @@ def test_gearbox_mission(self):
 
         prob.setup(force_alloc_complex=True)
 
-        prob.set_val(av.Aircraft.Engine.RPM_DESIGN, [5000, 6195, 6195], units='rpm')
-        prob.set_val(av.Dynamic.Mission.SHAFT_POWER, [100, 200, 375], units='hp')
-        prob.set_val(av.Dynamic.Mission.SHAFT_POWER_MAX, [375, 300, 375], units='hp')
+        prob.set_val(av.Dynamic.Mission.RPM + '_in', [5000, 6195, 6195], units='rpm')
```
ehariton, line 56: For me, the _in and _out variables aren't helping me remember what is what. A graphical representation would help me with this.

aviary/subsystems/propulsion/test/test_turboprop_model.py
```diff
@@ -176,7 +181,9 @@ def test_case_1(self):
 
         self.prob.run_model()
         results = self.get_results()
-        assert_near_equal(results, truth_vals)
+        assert_near_equal(results[0], truth_vals[0], tolerance=1.5e-12)
```
ehariton, line 184: These tol values are exceedingly small. Are you sure a 1e-4 wouldn't suffice?

aviary/subsystems/propulsion/gearbox/model/gearbox_premission.py
```diff
@@ -44,18 +44,25 @@ def setup(self):
...
+            # Simple gearbox mass will always produce positive values for mass based on a fixed specific torque
+            self.add_subsystem(
+                'mass_comp',
```
ehariton, line 50: If include_gearbox = false probably want to set this mass to zero if you have not done that elsewhere already.

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -79,6 +107,16 @@ def build_pre_mission(self, aviary_inputs, **kwargs) -> om.Group:
                     promotes=['*']
                 )
 
+        gearbox_model_pre_mission = gearbox_model.build_pre_mission(
+            aviary_inputs, **kwargs
+        )
+        if gearbox_model_pre_mission is not None:
```
ehariton, line 113: Will this leave mass.gearbox at zero if pre-mission isn't initialized?

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -286,37 +337,256 @@ def setup(self):
...
+            promotes_outputs=[('turboprop_thrust_max', Dynamic.Mission.THRUST_MAX)],
         )
 
     def configure(self):
-        # configure step to alias thrust output from shaft power model if present
+        """
+        Correctly connect variables between shaft power model, gearbox, and propeller,
+        aliasing names if they are present in both sets of connections.
+
+        If a gearbox is present, inputs to the gearbox are usually done via connection,
```
ehariton, line 352: So essentially without gearbox: Motor -> promoted -> Prop
with gearbox: Motor -> connection -> Gearbox -> promotion -> Prop
?

ehariton (approved):

aviary/subsystems/propulsion/turboprop_model.py
jkirk5: That's my understanding. If a custom gearbox model has no pre-mission and the user doesn't override a mass value in the input file, then mass _should_ stay at the default of zero, which is hopefully expected/intuitive behavior. I don't have a test of that though

aviary/subsystems/propulsion/gearbox/test/test_gearbox.py
jkirk5: I'll do that when documentation finally gets added

aviary/models/large_turboprop_freighter/large_turboprop_freighter.csv
jkirk5: No worries, this isn't the final model, just the raw converted GASP file. I'll double check my current version to make sure that got updated

crecine: in GASP, design range=0 means don't resize the aircraft, just fly the off design

aviary/subsystems/propulsion/propeller/hamilton_standard.py
```diff
@@ -457,6 +470,8 @@ def _biquad(T, i, xi, yi):
     .525, .540, .565, .615, .670, .710, .745, .790, .825, .860, .880, .895,  # X = 0.20
     .225, .260, .320, .375, .430, .495, .550, .610, .660, .710, .740, .775,  # X = 0.30
 ])
+# autopep8: off
+# fmt: off
```
jkirk5, line 474:
```suggestion
# autopep8: on
# fmt: on
```

aviary/subsystems/propulsion/turboprop_model.py
crecine: my preference would be:
motor/engine deck/engine model output: `power_source_rpm`
gearbox input: `power_source_rpm`
gearbox output: `prop_rpm`
propeller input: `prop_rpm`

If the gearbox exists, all connections are made automatically by promotions
If there is no gearbox, then a connection or alias is used.

I think that will be significantly more clear to users and could also simplify the code here

crecine (approved):

aviary/subsystems/propulsion/gearbox/test/test_gearbox.py
crecine: see https://github.com/OpenMDAO/Aviary/pull/514#discussion_r1823530694

aviary/subsystems/propulsion/gearbox/model/gearbox_mission.py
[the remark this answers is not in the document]
crecine: see https://github.com/OpenMDAO/Aviary/pull/514#discussion_r1823530694

aviary/models/large_turboprop_freighter/large_turboprop_freighter.csv
ehariton: Sounds like we have to set the mission_type = Alternate flag in that case and remove the mission.design.range from the csv. because mission.design.range of zero will brick things with the new multi-mission update.

aviary/subsystems/propulsion/gearbox/test/test_gearbox.py
ehariton: Those descriptions in #514 make sense to me.

PR #517, opened 2024-09-04
Override wetted area gasp_based and docstring update

### LINKED ISSUE #366: Simplify provide_surface_area option

The option Aircraft.Fuselage.PROVIDE_SURFACE_AREA is now obsolete with the addition of overrides. It should be removed and override functionality implemented

**_Erik:_**
There are currently four variables related to fuselage wetted area:
```
Aircraft.Fuselage.PROVIDE_SURFACE_AREA
Aircraft.Fuselage.WETTED_AREA
Aircraft.Fuselage.WETTED_AREA_FACTOR
Aircraft.Fuselage.WETTED_AREA_SCALER
```
In GASP geometry calculations, if `PROVIDE_SURFACE_AREA` is True, `WETTED_AREA_FACTOR` is used to override the actual fuselage wetted area. Otherwise, wetted area is calculated and scaled by `WETTED_AREA_FACTOR`.
In FLOPS geometry, if `WETTED_AREA` is input it is used to override the calculated fuselage wetted area. Otherwise, wetted area is calculated and scaled by `WETTED_AREA_SCALER`.

The GASP calculations should be modified to use the same process as the FLOPS calcs. `PROVIDE_SURFACE_AREA` and `WETTED_AREA_FACTOR` then should be eliminated.

**_Ken:_**
Some thoughts on this:

I agree with Erik that we should (wherever we can) make the interface between the 2 sides similar. We can use WETTED_AREA_SCALER on the flops side, so would be ideal to use it on the gasp side too.

For PROVIDE_SURFACE_AREA to work, we need overrides to be supported on the gasp side. I am not confident that they are implemented there, so getting that to work would be a pre-requisite for this issue.

_No response_

### Associated Bug Report

_No response_

xjjiang:
### Summary

- Override is a feature in Aviary that is available on both FLOPS based and GASP based variables. On FLOPS side, it has been verified. This PR confirms that it works on GASP side too. A unit test is added for this purpose.
- `Aircraft.Fuselage.WETTED_AREA_FACTOR` is replaced by `Aircraft.Fuselage.WETTED_AREA_SCALER` and `Aircraft.Fuselage.WETTED_AREA` is qualified for override. As a result, `Aircraft.Fuselage.Aircraft.Fuselage.PROVIDE_SURFACE_AREA` is not needed and is removed.
- Some of the unit tests have to be modified because override does not work for a single component.
- `fortran_to_aviary.py is updated for the change.
- `utils/set_mass_defaults.py` is not used. I suggest that we remove this file.
- `interface/default_phase_info/two_dof_fiti_deprecated.py` is not used. I removed it in this PR.
- Some of the docstring work is included in this PR. Somehow, I mixed both work into one branch. They should be easy read.

### Related Issues

- Resolves #366 

None

### New Dependencies

None


xjjiang:
two PRs are mixed. I will close this PR and submit two separate PRs.

PR #524, opened 2024-09-09
set initial guesses when providing an AviaryValues object to 'load_inputs()'

### LINKED ISSUE #313: Ability to run 2DOF-mission problem using level 2 interface without .csv

There is currently no way to provide initial guesses through the python interface, which prevents running 2DOF mission problems without a .csv file.

'load_inputs()' in `methods_for_level2.py` has the functionality to provide an AviaryValues object instead of a .csv. However, for 2DOF-mission problems initial guesses are strictly required, usually provided in the .csv as additional inputs. When trying to run a 2DOF-mission problem without a .csv, an error is triggered because initial guesses are defaulted to an empty dict.

### Associated Bug Report

_No response_

crecine:
### Summary

Added the ability to set initial guesses when providing an AviaryValues object to 'load_inputs()' in `methods_for_level2.py`

### Related Issues

- Resolves #313 

None

### New Dependencies

None


aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
```diff
@@ -156,14 +156,14 @@ mission:taxi:duration,0.1677,h
 settings:equations_of_motion,2DOF
 settings:mass_method,GASP
 
-# Initial Guesses
+# Initialization Guesses
```
Kenneth-T-Moore, line 159: I think "initial guesses" sounds better than "initialization guesses". (and is close to the math term for PDEs "initial conditions")

What the motivation for the change?

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
```diff
@@ -156,14 +156,14 @@ mission:taxi:duration,0.1677,h
...
 actual_takeoff_mass,0
 climb_range,0
 cruise_mass_final,0
 flight_duration,0
-fuel_burn_per_passenger_mile,0.1
```
Kenneth-T-Moore, line 164: Is there any ramification to zeroing out these initial conditions? Seems they are in all of the main benchmark-test GASP missions.

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
crecine: the intitial_guesses in phase_info are the initial conditions used throughout the trajectory for the optimizer.
These are variables used to initialize the values of other variables.

I wanted to disambiguate the two of them

aviary/models/large_single_aisle_1/large_single_aisle_1_GwGm.csv
crecine: no, if these variables are unset (or set to 0) they will be replaced with default values when the csv is read in. The values here (.1 and .99) are the default values that they will be replaced with, basically the defaults were specified twice for these two values, so I made them consistent with the rest of the variables here

jkirk5 (approved):

Kenneth-T-Moore (approved):

PR #534, opened 2024-09-11
Connect flight phase mass to gross mass

### LINKED ISSUE #325: Gross mass is not properly tied to the mission mass
### Description

@kaushikponnapalli and I have been working through an issue that we noticed when examining on/off-design capabilities.

The default `phase_info` for `height_energy` currently has `optimize_mass=True` _and_ `fix_initial=True`.
This means that the mass state in the Dymos mission is fixed to be whatever the initial value is, set by the initial guessing.
This is not forced to be the same as `Aircraft.Design.GROSS_MASS`, so those values can vary.
They should not be allowed to vary; this is a bug.
This only occurs when `include_takeoff=False`; if that is True, then the initial mass is set correctly.

### Example

See description; run a basic Aviary example with height energy

### Aviary Version

0.9.4-dev

_No response_

kaushikponnapalli:
### Summary
In the height energy formulation, when takeoff is not included in the phase info, the initial mass of the first flight phase is not connected to Mission.Summary.GROSS_MASS. When this is coupled with the option fix_initial=True it results in cases where the payload mass has no impact on the amount of fuel burned.
A constraint is now added in cases where takeoff is not included and the initial mass of the first phase is constrained to equal the gross mass.
If fix_initial is set to true, that is changed for the mass

### Related Issues

- Resolves #325 

None

### New Dependencies

None


Kenneth-T-Moore (approved):
Bump up the ref and i think this is good.

aviary/interface/methods_for_level2.py
```diff
@@ -1443,6 +1451,21 @@ def link_phases(self):
...
                                    src_indices=[-1], flat_src_indices=True)
 
+                if not self.pre_mission_info['include_takeoff']:
+                    first_flight_phase_name = list(self.phase_info.keys())[0]
+                    eq = self.model.add_subsystem(f'link_{first_flight_phase_name}_mass',
+                                                  om.EQConstraintComp(),
+                                                  promotes_inputs=[('rhs:mass',
+                                                                    Mission.Summary.GROSS_MASS)])
+                    eq.add_eq_output('mass', eq_units='lbm', normalize=False,
+                                     ref=10000., add_constraint=True)
```
jkirk5, line 1461:
```suggestion
                                     ref=100_000., add_constraint=True)
```

jkirk5 (approved):

PR #536, opened 2024-09-16
Add support for constraint aliases to the phase info.

### LINKED ISSUE #264: Multiple constraints for the same variable in the `phase_info` object are not respected

If you have two constraints in a `phase_info['constraints']` dict with the same variable name, only the last one is used. That's because within a dictionary you can only have unique keys. You might need multiple constraints if you want both initial and final boundary constraints on the variable.

We could rework the constraints handling so it's a list of dicts instead of a dict of dicts. This will be backwards incompatible. There might be a way to make constraints handle either a list of dicts _and_ a dict of dicts and that would remove the backwards incompatibility.

_No response_

### Associated Bug Report

_No response_

Kenneth-T-Moore:
### Summary

Added the ability to specify two constraints on the same dynamic variable in the phase info file. Note that the use case is a little contrived, as we already have phase_info arguments to constraint most of the dynamic variables. The api utilizes OpenMDAO's constraint alias feature, which was created to address this same problem. 

### Related Issues

- Resolves #264

None

### New Dependencies

None


aviary/interface/test/test_height_energy_mission.py
```diff
@@ -252,6 +253,48 @@ def test_custom_phase_builder_error(self):
                 optimizer='SLSQP',
             )
 
+    def test_support_constraint_aliases(self):
+        # Test specification of multiple constraints on a single variable.
```
xjjiang, line 257: Can you write this comment using the standard format: *** docstring ***?

aviary/interface/test/test_height_energy_mission.py
```diff
@@ -252,6 +253,48 @@ def test_custom_phase_builder_error(self):
...
+
+        prob.run_model()
+
+        prob_vars = prob.list_problem_vars()
+        cons = {key: val for (key, val) in prob_vars['constraints']}
+        con1 = cons['traj.phases.climb->initial_boundary_constraint->throttle_1']
+        con2 = cons['traj.phases.climb->final_boundary_constraint->throttle_2']
+
+        self.assertEqual(con1['name'], 'timeseries.throttle_1')
+        self.assertEqual(con2['name'], 'timeseries.throttle_2')
```
jkirk5, line 296: So this checks that both constraints were added to the problem by checking the name? Can we add an assert that the actual throttle values at start and end of phase match the constraints to make sure they are both active?

jkirk5 (approved):

aviary/interface/test/test_height_energy_mission.py
```diff
@@ -252,6 +253,48 @@ def test_custom_phase_builder_error(self):
                 optimizer='SLSQP',
             )
 
+    def test_support_constraint_aliases(self):
+        # Test specification of multiple constraints on a single variable.
+        modified_phase_info = self.phase_info.copy()
```
crecine, line 258: do we want to use a deep copy here? I think that's what we have done most of the time we have made a copy of the phase_info

aviary/mission/phase_builder_base.py
```diff
@@ -419,7 +419,15 @@ def _add_user_defined_constraints(self, phase, constraints):
...
                 kwargs.pop('type')
-                phase.add_boundary_constraint(constraint_name, **kwargs)
+
+                if 'target' in kwargs:
+                    # Support for constraint aliases.
+                    target = kwargs.pop('target')
+                    kwargs['constraint_name'] = constraint_name
+                    phase.add_boundary_constraint(target, **kwargs)
+                else:
+                    phase.add_boundary_constraint(constraint_name, **kwargs)
```
crecine, line 429: would we ever want this alias for the non-boundary constraints?

crecine (approved):
minor question

aviary/mission/phase_builder_base.py
Kenneth-T-Moore: A non-boundary constraint is just a path constraint, where you specify upper or lower bounds that the whole phase shouldn't exceed. We don't support a general constraint yet (i.e., specify values at will at specific points in the trajectory), but it might be good for future work.

aviary/interface/test/test_height_energy_mission.py
Kenneth-T-Moore: Yeah, it wouldn't hurt to be careful.

aviary/interface/test/test_height_energy_mission.py
Kenneth-T-Moore: I initially tried to do that, but I couldn't find a case that didn't end up overconstrained when I added these on the throttle. We just haven't come up with a useful solvable case where we need this.

PR #529, opened 2024-09-11
Multi mission opt

ehariton:
### Summary

The Goal was to create a large_single_aisle example of a multi-mission monolithic optimization to allow an aircraft to be optimized for two missions, a mission where the aircraft was full of passengers, and one where the aircraft had no passengers. The implementation contains two pre-mission, missions, and post-mission analysis followed by summing the objective functions from both analysis into a single objective.

A number of updates to the way that aircraft are described were needed to make this PR work because there was previously no notion of "Design" passenger quantities vs. "As-flown" passenger quantities. In order to make a mission that could simulate both full and empty passenger seats, a clear division was created between these two values:

**New Design Terms Added:**
Design.NUM_PASSENGERS
Design.NUM_BUSINESS_CLASS
Design.NUM_TOURIST_CLASS
Design.NUM_FIRST_CLASS
Mission.Design.RANGE  vs. target_range

Weights for all subsystems in pre-mission are sized based on the Design.pax values. This required a number of updates throughout different subsystems. For example, the weight of the air conditioning system should not be affected by the number of passengers on the plane. Rather it is affected by how many passengers the plane was designed for. That is because we don't remove the air conditioning if there are fewer people on the plane!

A number of checks had to be added to help users if they never specified Design.pax values. Those tests, run in preprocessor.py, will assist the user in filling out Design.pax values if they are not included in the csv files, enabling backwards compatibility. 

Tested with FLOPS weights and mission models. 

The `N3CC_data.py` configuration file has been revised. This file previously called `preprocess_options()`. This would trigger whenever `import get_flops_inputs` was called causing aviary to be unable to detect what aviary_values had been set by the user and what had been left blank. `preprocess_options()` fills in those blank areas. Removing this call from `N3CC_data.py` exposed that in many of the N3CC related tests, `preprocess_options()` was never called so it had to be added.

All the aircraft .csv definitions were changed from specifying the as-flow passenger values to Design.pax values. This was required to make checks for the FLOPS and GASP converters work successfully. Those converters translate `INGASP.PAX` and `'aircraft.outputs.L0_crew_and_payload.passenger_count'` to `Aircraft.CrewPayload.Design.NUM_PASSENGERS`, reflecting GASP and FLOPS intent to set design passengers not as-flown passengers. There are no equivalents for historical_names for as-flown passengers so those are left blank.

### Related Issues

- Resolves #

### Backwards incompatibilities

None

### New Dependencies

None

PR #538, opened 2024-09-19
Preprocessors update

cmbenne3:
### Summary

This PR makes a small change to preprocessors to remove a block of code that was never used and improve the intelligence of the aviary code base. Methods for level 2 has been changed to utilize the preprocess_options method rather than calling propulsion and crewpayload methods separately.

#### Before this Change:
Aviary checked if 'Aircraft.CrewPayload.NUM_PASSENGERS' was in aviary_options. 
If not then it would be calculated as the sum of First + Business + Tourist Passengers.

This logic was never invoked since before this check happened the aviary_options object gets initialized with default values (and the default values for all passenger numbers is 0). 
If a user specified First + Business + Tourist passenger numbers without specifying a total NUM_PASSENGERS they would get an error that the total (0) didn't match the sum, which is annoying!

**Function call order:**
- prob.load_inputs() [methods_for_level2.py]
    - create_vehicle() [process_input_decks.py]
    -  get_option_defaults() [options.py]
       (this sets all 4 num passenger parameters to 0 before updating inputs from input file or aviary_inputs object.)
- prob.check_and_preprocess_inputs() [methods_for_level2.py] 
  - preprocess_options() [preprocessors.py]  
  
#### After this Change:
Aviary will now calculate NUM_PASSENGERS if this is set to 0 and there are non-zero number of First + Business + Tourist passengers specified. Aviary will throw an error if the input values are inconsistent. The exception is a user is allowed to only specify the total without specifying which classes those passengers sit in.

The the checks for crew remain unchanged since they no not have default values specified in variable_meta_data and so the logic is valid.

### Related Issues

N/A

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -496,8 +496,7 @@ class AviaryProblem(om.Problem):
 
         # PREPROCESSORS #
         # Fill in anything missing in the options with computed defaults.
-        preprocess_propulsion(aviary_inputs, self.engine_builders)
-        preprocess_crewpayload(aviary_inputs)
+        preprocess_options(aviary_inputs, engine_models=self.engine_builders)
 
         mission_method = aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
         mass_method = aviary_inputs.get_val(Settings.MASS_METHOD)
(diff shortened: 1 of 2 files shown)


cmbenne3:
Changes merged with Multi mission opt #529. This PR is no longer required.
https://github.com/OpenMDAO/Aviary/pull/529

PR #543, opened 2024-09-24
Update of gross mass design variable bounds

cmbenne3:
### Summary

Update of Gross mass upper and lower bounds and reference values for consistency across mission_methods and to allow for modelling of aircraft up to 900000lbm gross mass.

Before this change the max gross mass that aviary could model using height energy or 2DOF missions was 400000lbm.

These 2 aircraft would still be outside the capabilities of Aviary - should we increase the limit to e.g. 1.5million lbm?:
* A380 max gross weight of 1.2million lbm
* AN225 max gross weight of 1.4 million lbm

Other option would be to set dynamically as a fraction (e.g. 0.7/1.3) of the user input gross mass value.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore:
Can you take a couple of the IPOPT benchmarks and run them on your branch and on main, and compare the itercount? I am curious if moving the lower bound from 10k to 10 impacts that optimizer.

cmbenne3:
FwFm benchmark with IPOPT comparison of command Line Output from main branch (left) and des_var_bounds branch with upper bounds set = None:

image

I believe this shows that there is no difference in the output, and the optimizer runs for the same number of iterations in both cases.

cmbenne3:
GwGm benchmark with IPOPT comparison of Command Line Output from the main branch (left) and the des_var_bounds branch with upper_bounds set = None:

image

Difference in number of variables with only lower vs lower and upper is expected (red ticks)

IPOPT seems to be taking slightly less time and fewer iterations with the upper_bound set = None, which is interesting.

It looks like it might be exiting before successfully completing the optimization (line 120) Please can someone confirm? This is the same for both the main branch and the one with modifications.

Small differences in the overall output beyond 5th decimal place - these seem insignificant.

cmbenne3:
@Kenneth-T-Moore - I've Run FwFm and GwGm, which I think covers Height Energy and Two Degrees of Freedom mission methods - do we have a SOLVED_2DOF benchmark too, I see a .csv file for it, but I don't see it in any of the benchmark test scripts...

Kenneth-T-Moore (approved):
ok, looks like the benchmarks working fine with the new bounds. I don't know why the one benchmark actually improves, but i am happy with that.

jkirk5 (approved):

PR #550, opened 2024-09-30
Issue233

### LINKED ISSUE #233: Make initial guessing consistent in terms of (initial, final) value pairs or (initial, duration)

Migrated from old repo:

"I thought that, for initial guesses on dymos phases, we provide an initial and final value. Here it looks like an initial value and a delta, which is a little inconsistent."

Originally posted by @Kenneth-T-Moore 

John's response: I agree with you it's a bit inconsistent. This is true because Dymos is expecting initial-final pairs for some things and initial-duration pairs for others (like time or in this case, mass). We can remove that complexity from an Aviary user though with some quick math. I'll make an issue based on your comment for this.

**This will be a breaking non-backwards-compatible change, must increment version accordingly.**

_No response_

### Associated Bug Report

_No response_

cmbenne3:
### Summary

This PR makes the initial guessing in phase info consistent for the user and requires initial guesses for all phases to be specified as [initial, final] pairs.

The initial guessing functions are also tidied up to try and reduce the number of separate logic blocks in the code.
At some point in the future it would be nice to get the lists of control and state variables from within the phase definitions, rather than having hardcoded lists that are dependent on phase_name and mission_method.
The initial guesses for the integration variable (usually time, sometimes mass) is now specified by checking the dymos phase for the 'name' of 'time', which is a more robust implementation, however the need for checking phase_name and mission_method has not been totally eliminated.

### Related Issues

- Resolves #233 

All phase info dictionaries with initial guesses will need to be updated to conform with the [initial, final] specification.

### New Dependencies

None


xjjiang:
Will you make a more meaningful title?

PR #558, opened 2024-10-05
Moving gasp based taxi, landing, and breguet components

xjjiang:
### Summary

- Moved `mission/gasp_based/phases/landing_components.py` to `mission/gasp_based/ode/landing_eom.py`
- Moved `mission/gasp_based/phases/landing_group.py` to `mission/gasp_based/ode/landing_ode.py`
- Moved `mission/gasp_based/phases/test/test_landing_components.py` to `mission/gasp_based/ode/test/test_landing_eom.py`
- Moved `mission/gasp_based/phases/taxi_component.py` to `mission/gasp_based/ode/taxi_eom.py`
- Moved `mission/gasp_based/phases/taxi_group.py` to `mission/gasp_based/ode/taxi_ode.py`
- Moved `mission/gasp_based/phases/test/test_taxi_component.py` to `mission/gasp_based/ode/test/test_taxi_eom.py`
- Moved `mission/gasp_based/phases/breguet.py` to  `mission/gasp_based/ode/breguet_cruise_eom.py`
- Vectorized `taxi_eom.py`.
- Modified `test_taxi_eom.py` to test vectorization.
- Switch `event_xform` and `taxi`, switch `h_fit` and `landing`. Rename `link_taxi_groundroll` to `taxi_groundroll_mass_constraint`.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/mission/gasp_based/ode/landing_eom.py
```diff
@@ -11,13 +11,22 @@ class LandingAltitudeComponent(om.ExplicitComponent):
...
+    def initialize(self):
+        self.options.declare('num_nodes', default=1, types=int)
+
     def setup(self):
-        add_aviary_input(self, Mission.Landing.OBSTACLE_HEIGHT, val=50.0)
-        add_aviary_input(self, Mission.Landing.AIRPORT_ALTITUDE, val=0.0)
+        nn = self.options['num_nodes']
+
+        add_aviary_input(self, Mission.Landing.OBSTACLE_HEIGHT, val=np.ones(nn)*50.0)
+        add_aviary_input(self, Mission.Landing.AIRPORT_ALTITUDE, val=np.zeros(nn))
```
Kenneth-T-Moore, line 21: When we vectorize a group of components for use in the ODE, we need to think about what it is going to be doing, and what values it will take on. If we think about landing, it computes a bunch of different quantities, but really only has a small subset variables that we track across the trajectory. These are the ones that change with time throughout the entire mission, and likewise do so in landing too.

Mass - initial comes from previous phase; final is computed (though I don't see this in landing, perhaps very little fuel is burned.
Velocity - initial comes from previous phase; final is 0.
Distance - initial comes from previous phase, final is computed by summing several of the computed distances like flare or decision.
Probably altitude - initial comes from previous phase, final is airport altitude.
Possibly throttle

Additionally, for each of those variables, there are only certain values for num_nodes that make sense. We could do 2 (start end), or 3 (start, touchdown, end), since we compute values for those, or maybe 4 (start, flare, touchdown, end).  You should look through the calculations and figure out what level of time-discretization makes the most sense, and then require num_nodes to be fixed to that amount. Any other value would break this, because it is not inherently vectorizable in time.

Also, most of the variables you vectorized should not be vectorized. Airport altitude is not changing as a function of time. What you did would work in a multi-mission or a fleet context, but we need discretization along the mission.

xjjiang: Now, I rolled back the vectorization in landing EOM.

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -11,39 +12,47 @@ class TaxiFuelComponent(om.ExplicitComponent):
...
-                Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL])
+                Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL],
+            rows=arange, cols=arange)
         self.declare_partials(
-            Dynamic.Mission.MASS, Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL)
+            Dynamic.Mission.MASS, Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL,
+            rows=arange, cols=arange)
         self.declare_partials(
-            Dynamic.Mission.MASS, Mission.Summary.GROSS_MASS, val=1)
+            Dynamic.Mission.MASS, Mission.Summary.GROSS_MASS)
```
Kenneth-T-Moore, line 55: `val=1`  would become `val=np.ones(num_nodes)` here (and then you can remove it from compute_partials.)

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -52,8 +61,14 @@ def compute(self, inputs, outputs):
         outputs[Dynamic.Mission.MASS] = takeoff_mass - outputs["taxi_fuel_consumed"]
 
     def compute_partials(self, inputs, J):
+        nn = self.options['num_nodes']
+
         dt_taxi = self.options['aviary_options'].get_val(Mission.Taxi.DURATION, 's')
 
-        J["taxi_fuel_consumed", Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL] = -dt_taxi
+        J["taxi_fuel_consumed", Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL] = \
+            -np.ones(nn) * dt_taxi
```
Kenneth-T-Moore, line 69: You can actually change this to:
    
    J["taxi_fuel_consumed", Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL][:] = -dt_taxi

to save some math operations by taking advantage of numpy's ability to broadcast a scalar into slices.

aviary/mission/gasp_based/ode/test/test_taxi_eom.py
```diff
@@ -17,21 +17,25 @@ class TaxiFuelComponentTestCase(unittest.TestCase):
...
         self.prob.model.add_subsystem('taxi', TaxiFuelComponent(
-            aviary_options=aviary_options), promotes=['*'])
+            num_nodes=num_nodes, aviary_options=aviary_options), promotes=['*'])
 
     def test_fuel_consumed(self):
         self.prob.setup(force_alloc_complex=True)
 
-        self.prob.set_val(
-            Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL, -1512, units="lbm/h")
+        self.prob.set_val(Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL,
```
Kenneth-T-Moore, line 31: I think I would use different values for the 2 elements in this test. I know this is a simple component, and will work fine, but it is helpful in catching more errors in general.

aviary/mission/gasp_based/ode/taxi_eom.py
xjjiang: I interpret that you want to keep `val=1`. I just did.

aviary/mission/gasp_based/ode/test/test_taxi_eom.py
xjjiang: Done.

aviary/mission/gasp_based/ode/taxi_eom.py
xjjiang: Got it. Done.

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -11,37 +12,45 @@ class TaxiFuelComponent(om.ExplicitComponent):
...
             "taxi_fuel_consumed", [
-                Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL])
+                Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL],
+            rows=arange, cols=arange)
         self.declare_partials(
-            Dynamic.Mission.MASS, Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL)
+            Dynamic.Mission.MASS, Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL,
+            rows=arange, cols=arange)
         self.declare_partials(
             Dynamic.Mission.MASS, Mission.Summary.GROSS_MASS, val=1)
```
jkirk5, line 55:
```suggestion
            Dynamic.Mission.MASS, Mission.Summary.GROSS_MASS, val=np.ones(nn))
```

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -11,37 +12,45 @@ class TaxiFuelComponent(om.ExplicitComponent):
...
         self.add_output(
             Dynamic.Mission.MASS,
-            val=175000.0,
+            val=np.ones(nn)*175000.0,
             units="lbm",
             desc="mass after taxi",
         )
 
+    def setup_partials(self):
+        arange = np.arange(self.options['num_nodes'])
```
jkirk5, line 45:
```suggestion
        nn = self.options['num_nodes']
        arange = np.arange(nn)
```

jkirk5: Doing this so nn is available later for another change

aviary/mission/gasp_based/ode/taxi_eom.py
jkirk5: I think this is what Ken meant? Then deleting line 72 since we already defined it here

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -52,8 +61,12 @@ def compute(self, inputs, outputs):
...
     def compute_partials(self, inputs, J):
+        nn = self.options['num_nodes']
+
         dt_taxi = self.options['aviary_options'].get_val(Mission.Taxi.DURATION, 's')
 
         J["taxi_fuel_consumed", Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL] = -dt_taxi
 
         J[Dynamic.Mission.MASS, Dynamic.Mission.FUEL_FLOW_RATE_NEGATIVE_TOTAL] = dt_taxi
+
+        J[Dynamic.Mission.MASS, Mission.Summary.GROSS_MASS] = np.ones(nn)
```
jkirk5, line 72:
```suggestion
```

jkirk5: Removed since the value can be defined during `declare_partials()`

jkirk5 (changes requested):

aviary/mission/gasp_based/ode/taxi_eom.py
xjjiang: Right now, `Mission.Summary.GROSS_MASS` is not a vector. Is it wrong?

aviary/mission/gasp_based/ode/taxi_eom.py
xjjiang: Agree. Done.

jkirk5 (approved):

aviary/mission/gasp_based/ode/taxi_eom.py
```diff
@@ -52,6 +62,8 @@ def compute(self, inputs, outputs):
         outputs[Dynamic.Mission.MASS] = takeoff_mass - outputs["taxi_fuel_consumed"]
```
Kenneth-T-Moore, line 62: I think we should abandon the vectorization of the taxi_eom, since we aren't really discretizing it. We really just have a starting mass, a delta_t, and a fuelflow rate.  All we really compute is the ending mass, which is start mass - fuelflow*delta_t. With the new changes, only fuelflow is a vector. To make this work, delta_t in this calculation would need to be an array of time points (something like np.linspace(0, 1, num_nodes) * dt_taxi instead of dt_taxi.) It just doesn't seem worth it, because we aren't doing a meaningful calculation here,  Fuelflow is always assumed constant over the taxi.  If it ever becomes time-varying (i.e., we are letting the optimizer pick taxi throttle speeds), then it might become relevant.

xjjiang: done.

Kenneth-T-Moore (approved):

PR #567, opened 2024-10-11
Change atmosphere comp defaults from geopotential to Geodetic

### LINKED ISSUE #466: Atmosphere comp defaults to geopotential alt
### Description

The USatm1976Comp from dymos we are using actually defaults to using geopotential altitude - this may be inconsistent with how we are treating altitude across different subsystems including mission analysis.

Since there is a conversion factor between them, we should be consistent with which altitude we use. Possible further discussion is needed on the merits of geopotential altitude for conceptual design and if we want the option to pick which to use, or globally hardcode Aviary to a specific altitude type.

Components such as table loading (such as in propulsion and aero) need to have the capability to convert altitude type. Propulsion uses the FLOPS equations for geopotential -> geometric altitude, which is inconsistent with the 1976 US atmosphere standard. All components should use the conversion method in the USatm1976Comp for consistency with a published method.

### Example

No specific example - any place `USatm1976Comp` is used in Aviary, the options `h_def` is not set, which defaults it to `geopotential`

### Aviary Version

0.9.4-dev

_No response_

xjjiang:
### Summary

In this PR, the default option `h_def` in `atmosphere.py` is switched to `geodetic` from `geopotential`. As a result, some validation data are updated. Currently there is no interface to alter the default value but in the future, we will have an atmosphere builder to allow users to alter this option.

Some default values in `aviary/subsystems/aerodynamics/gasp_based/gaspaero.py` are changed in order to avoid some divide-by-zero warning messages.

In `aviary/subsystems/propulsion/test/test_custom_engine_model.py`, I have to modify the `phase_info` by setting `num_segments = 3` instead of `num_segments = 2` in order to avoid a unit test failure on GitHub. The unit test does not fail locally but only on GitHub. I had to try many different options to get it to work. This phenomenon happened before. I had to adjust tolerance a little bit. But this time it seemed that a constraint was violated in OpenMDAO and a constraint variable was reset. That might have resulted in an `inf` or `NaN` somewhere such that the unit test failed.

`Aircraft.Engine.GEOPOTENTIAL_ALT` is not changed.

We don't have a doc page on atmosphere.

### Related Issues

- Resolves #466 

None

### New Dependencies

None


aviary/validation_cases/benchmark_tests/test_bench_multiengine.py
```diff
@@ -124,8 +124,8 @@ def test_multiengine_static(self):
         alloc_cruise = prob.get_val('traj.cruise.parameter_vals:throttle_allocations')
         alloc_descent = prob.get_val('traj.descent.parameter_vals:throttle_allocations')
 
-        assert_near_equal(alloc_climb[0], 0.5, tolerance=1e-2)
-        assert_near_equal(alloc_cruise[0], 0.64, tolerance=1e-2)
+        assert_near_equal(alloc_climb[0], 0.5137, tolerance=1e-2)
```
Kenneth-T-Moore, line 127: Can you set these to their original values and leave it failing? This regression is due to a different bug, related to the design range: #569

xjjiang: Done.

Kenneth-T-Moore (approved):

PR #577, opened 2024-10-18
Headers for converted input files

jkirk5:
### Summary

Tiny PR to add header creation back to fortran_to_aviary converter and remove a warning about invalid escape characters

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/utils/test/test_fortran_to_aviary.py
+++ b/aviary/utils/test/test_fortran_to_aviary.py
@@ -41,7 +41,7 @@ class TestFortranToAviary(unittest.TestCase):
         # Execute the conversion
         _exec_F2A(args, None)
 
-    def compare_files(self, filepath, skip_list=[]):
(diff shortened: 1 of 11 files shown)


aviary/utils/test/test_fortran_to_aviary.py
```diff
@@ -57,6 +58,11 @@ def compare_files(self, filepath, skip_list=[]):
             for line in f_in:
                 if any(s in line for s in skip_list):
                     break
+                # skip first header line, as it changes based on time of execution and
+                # username
+                timestamp = datetime.now().strftime('%m/%d/%y at')
+                if f'# created {timestamp}' in line:
+                    break
```
crecine, line 65: instead of adding this, could we just use the existing skip_list, like in `test_engine_deck_conversion.py`:
`self.compare_files(filename, skip_list=['# created'])`

Kenneth-T-Moore (approved):

PR #578, opened 2024-10-21
Smoothen minimum function in propeller performance

xjjiang:
### Summary

Currently, there are two uses of `min()` function in `InstallLoss` class. That results in the discontinuity of partial derivatives. These two min() functions are replaced by a smoothed minimum function using the log-sum-exp trick. The new function along with its derivative are added to `aviary/subsystems/propulsion/propeller/propeller_performance.py` because it is not used anywhere else. The smoothing factor `alpha` is set to `100.0`. It is tested that this factor should be limited to between 75 and 275. Otherwise, a few unit tests will fail.

`aviary/subsystems/propulsion/propeller/hamilton_standard.py` is updated at a few places to avoid a few `numpy` warnings and to fix a bug.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/propulsion/propeller/propeller_performance.py
+++ b/aviary/subsystems/propulsion/propeller/propeller_performance.py
@@ -32,7 +32,7 @@ def smooth_min(x, b, alpha=100.0):
 
 def d_smooth_min(x, b, alpha=100.0):
     """
-    Derivative of function smooth_min
+    Derivative of function smooth_min(x)
 
     Parameters:
     x (float or array-like): First value.
--- a/aviary/subsystems/propulsion/propeller/propeller_performance.py
+++ b/aviary/subsystems/propulsion/propeller/propeller_performance.py
@@ -391,8 +391,8 @@ class AreaSquareRatio(om.ExplicitComponent):
         if smooth:
             alpha = self.options['alpha']
             sqa = smooth_min(sqa, 0.50, alpha)
-            dSQA_dNacDiam = d_smooth_min(sqa, alpha) * dSQA_dNacDiam
-            dSQA_dPropDiam = d_smooth_min(sqa, alpha) * dSQA_dPropDiam
+            dSQA_dNacDiam = d_smooth_min(sqa,  0.50, alpha) * dSQA_dNacDiam
+            dSQA_dPropDiam = d_smooth_min(sqa,  0.50, alpha) * dSQA_dPropDiam
 
         partials['sqa_array', "DiamNac"] = np.ones(nn) * dSQA_dNacDiam
         partials['sqa_array', "DiamProp"] = np.ones(nn) * dSQA_dPropDiam
@@ -454,9 +454,9 @@ class AdvanceRatio(om.ExplicitComponent):
         smooth = self.options["smooth_zje"]
         if smooth:
             alpha = self.options["alpha"]
-            djze_dsqa = d_smooth_min(sqa_array, alpha) * djze_dsqa
-            djze_dvktas = d_smooth_min(sqa_array, alpha) * djze_dvktas
-            djze_dtipspd = d_smooth_min(sqa_array, alpha) * djze_dtipspd
+            djze_dsqa = d_smooth_min(sqa_array,  0.50, alpha) * djze_dsqa
+            djze_dvktas = d_smooth_min(sqa_array,  0.50, alpha) * djze_dvktas
+            djze_dtipspd = d_smooth_min(sqa_array,  0.50, alpha) * djze_dtipspd
 
         partials["equiv_adv_ratio", "sqa_array"] = djze_dsqa
         partials["equiv_adv_ratio", "vktas"] = djze_dvktas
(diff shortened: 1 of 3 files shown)


jkirk5 (approved):

aviary/subsystems/propulsion/propeller/propeller_performance.py
```diff
@@ -305,6 +339,130 @@ def compute_partials(self, inputs, J):
...
+
+        dSQA_dNacDiam = 2 * diamNac / diamProp**2
+        dSQA_dPropDiam = -2 * diamNac**2 / diamProp**3
+
+        smooth = self.options["smooth_sqa"]
+        if smooth:
+            alpha = self.options['alpha']
+            sqa = smooth_min(sqa, 0.50, alpha)
+            dSQA_dNacDiam = d_smooth_min(sqa, alpha) * dSQA_dNacDiam
+            dSQA_dPropDiam = d_smooth_min(sqa, alpha) * dSQA_dPropDiam
```
crecine, line 395: here (and potentially elsewhere) `d_smooth_min(sqa, alpha)` means that when `def d_smooth_min(x, b, alpha=100.0)` is called, sqa will be used as x, alpha will be used as b. I think this needs to be `d_smooth_min(sqa, .5, alpha)`

aviary/subsystems/propulsion/test/test_propeller_performance.py
```diff
@@ -507,7 +508,7 @@ def test_mach(self):
             minimum_step=1e-12, abs_err_tol=5.0E-4, rel_err_tol=5.0E-5)
         assert_check_partials(partial_data, atol=1e-4, rtol=1e-4)
 
-    def test_tip_mach(self):
+    def tstest_tip_mach(self):
```
crecine, line 511: is this renaming intentional? I think this means that testflo won't recognize it as a test any more

aviary/subsystems/propulsion/test/test_propeller_performance.py
```diff
@@ -531,7 +532,7 @@ def test_tip_mach(self):
         assert_check_partials(partial_data, atol=1e-4, rtol=1e-4)
 
 
-class TipSpeedLimitTest(unittest.TestCase):
+class TipSpeedLimitest(unittest.TestCase):
```
crecine, line 535: is this renaming intentional? It looks like it was an accident?

aviary/subsystems/propulsion/test/test_propeller_performance.py
```diff
@@ -574,5 +575,95 @@ def test_tipspeed(self):
...
+            promotes=["*"],
+        )
+        prob.setup(force_alloc_complex=True)
+        prob.set_val("DiamNac", val=2.8875, units='ft')
+        prob.set_val("DiamProp", val=10.0, units='ft')
+        prob.run_model()
+
+        sqa_ratio = prob.get_val("sqa_array", units='unitless')
+        assert_near_equal(sqa_ratio, [
+            0.08337656, 0.08337656, 0.08337656], tolerance=1e-5)
```
crecine, line 597: it looks like the sqa array is not dependent on time, is there a reason it is an array instead of a scalar?

aviary/subsystems/propulsion/test/test_propeller_performance.py
```diff
@@ -574,5 +575,95 @@ def test_tipspeed(self):
...
+        )
+        prob.setup(force_alloc_complex=True)
+        prob.set_val("vktas", val=[0.1, 125., 300.], units='knot')
+        prob.set_val("tipspd", val=[800., 800., 750.], units='ft/s')
+        prob.set_val("sqa_array", val=[0.0756, 0.0756, 0.0756], units='unitless')
+        prob.run_model()
+
+        equiv_adv_ratio = prob.get_val("equiv_adv_ratio", units='unitless')
+        assert_near_equal(equiv_adv_ratio, [
+            0.000650881807, 0.813602259, 2.08282178], tolerance=1e-5)
```
crecine, line 642: for these two tests, could you add a fourth element that will trigger the limit?
Doesn't have to be "realistic" adding a 4th vktas that is 1000 would probably do it.

aviary/subsystems/propulsion/test/test_propeller_performance.py
xjjiang: I don't know what happened. It is named back now.

aviary/subsystems/propulsion/test/test_propeller_performance.py
xjjiang: I did a global replacement. Know it is named back.

aviary/subsystems/propulsion/test/test_propeller_performance.py
xjjiang: It is true that sqa array is not dependent on time, but a later component needs a sqa_array. In current main, there is a `convert_sqa` component in `propeller_performance.py`.

aviary/subsystems/propulsion/test/test_propeller_performance.py
xjjiang: Done adding unit tests that trigger the limits.

aviary/subsystems/propulsion/test/test_propeller_performance.py
```diff
@@ -532,7 +533,7 @@ def tstest_tip_mach(self):
         assert_check_partials(partial_data, atol=1e-4, rtol=1e-4)
 
 
-class TipSpeedLimitest(unittest.TestCase):
+class TipSpeedLimittest(unittest.TestCase):
```
crecine, line 536:
```suggestion
class TipSpeedLimitTest(unittest.TestCase):
```

crecine (approved):

PR #581, opened 2024-10-22
adding unit test for json

crecine:
### Summary

Added tests to check functionality of reading/writing `.json` files for off-design mission analysis

### Related Issues

- Resolves #

None

### New Dependencies

None


cmbenne3:
Thanks for adding this test @crecine. I think we should add (either to here) or perhaps to the benchmark tests some variations where we vary payload and range. This would catch the error we currently have where variation in payload accidentally varies the design of the aircraft for gasp-based missions... Obviously I would need to fix that problem before we add to the tests.

aviary/interface/methods_for_level2.py
```diff
@@ -2516,12 +2516,12 @@ def save_sizing_to_json(self, json_filename='sizing_problem.json'):
 
                     # Lists are fine except if they contain enums
                     if type_value == list:
-                        if type(type(value[0])) == enum.EnumType:
+                        if type(type(value[0])) == type(enum.Enum):
```
jkirk5, line 2519:
```suggestion
                        if isinstance(value[0], enum.Enum):
```
Should get you want you want but a little easier to read - `isinstance` will handle inheritance so you don't need to nest the `type()`s

aviary/interface/methods_for_level2.py
```diff
@@ -2516,12 +2516,12 @@ def save_sizing_to_json(self, json_filename='sizing_problem.json'):
...
                             for i in range(len(value)):
                                 value[i] = str([value[i]])
 
                     # Enums need converting to a string
-                    if type(type(value)) == enum.EnumType:
+                    if type(type(value)) == type(enum.Enum):
```
jkirk5, line 2524:
```suggestion
                    if isinstance(value, enum.Enum):
```

aviary/interface/test/test_json.py
```diff
@@ -0,0 +1,58 @@
...
+
+from aviary.utils.functions import get_aviary_resource_path
+from openmdao.utils.testing_utils import require_pyoptsparse, use_tempdirs
+import aviary.api as av
+from aviary.interface.default_phase_info.height_energy import phase_info, phase_info_parameterization
+
+
+@use_tempdirs
+@require_pyoptsparse
+class TestJson(unittest.TestCase):
```
jkirk5, line 12: From what I understand this is just checking that the json files _can_ be saved or loaded, and doesn't check that the files were properly created or that the off-design mission ran correctly - can we add comments explaining this, along with TODOs or an issue created to "finish" these tests and add a 2DOF test as well?

jkirk5 (approved):
Minor requests

aviary/interface/methods_for_level2.py
crecine: for some reason, isinstance was not triggering properly from some of the enums, but this did. I can experiment some more

aviary/interface/test/test_json.py
crecine: this was mainly intended to catch the problem that EnumType wasn't added until 3.13, so these functions didn't work with the minimum versions we list, but since they weren't being tested anywhere, CI didn't catch that

aviary/examples/run_off_design_example.py
```diff
@@ -117,7 +118,10 @@
 
 # Link phases and variables
 prob.link_phases()
-prob.add_driver('SNOPT', max_iter=100)
+if env_truthy("TESTFLO_RUNNING"):
+    prob.add_driver('SLSQP', max_iter=100)
+else:
+    prob.add_driver('SNOPT', max_iter=100)
```
ehariton, line 124: What happens if a user doesn't have SNOPT installed?

ehariton (approved):

aviary/examples/run_off_design_example.py
crecine: this example was previously set up to use SNOPT, all this PR does it add it to CI and switch the optimizer to SLSQP if it is running in CI, so that SNOPT isn't required when it is run as a test instead of an example

cmbenne3 (approved):

PR #582, opened 2024-10-22
Addition of global_throttle flag to variable hierarchy

### LINKED ISSUE #554: Add flag for local/global throttle normalization

There is a hidden (i.e. hardcoded) option to toggle which normalization scheme is used by `engine_deck.py` on line 185-186. These should be exposed as an option that can be passed/defined when setting up an Aviary (`Aircraft.Engine.GLOBAL_NORMALIZATION` or something similar?).

Some discussion on defaults might be needed - local normalization is pretty much universally desired except when replicating results for GASP-translated engine decks, which need global right now. The fix is to add a T4 constraint to each individual mission segment to set throttle to matche what the engine ran at when used in GASP, but that may be out of scope for this version update. If we don't do that, then 2DOF benchmark tests may provide slightly different values, which is ok, just may not match quite as closely to legacy results.

If we do local normalization as a default, then a new test that ensures the normalization worked properly is required (with a little dummy data set designed solely to test this feature)

_No response_

### Associated Bug Report

_No response_

jkirk5:
### Summary

Exposes the previously hidden option to toggle if EngineDecks used a global throttle range across the entire data set, or disable it to normalize the throttle per flight condition (the default).

GASP-converted input files have global_throttle set to True, as GASP-based engine decks have varied throttle ranges and do not produce matching results if throttle is normalized per flight condition.

Will route this to 0.9.6.dev branch once reviews are in and it is still passing testing

### Related Issues

- Resolves #554 

None

### New Dependencies

None


aviary/subsystems/mass/gasp_based/test/test_mass_summation.py
ehariton, line 1: Did you mean to commit these changes in this PR?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -180,12 +184,20 @@ def __init__(self, name='engine_deck', options: AviaryValues = None,
         # Create dict for variables present in engine data with associated units
         self.engine_variables = {}
 
-        # TODO make this an option - disabling global throttle ranges is better to
-        #      prevent unintended extrapolation, but breaks missions using GASP-based
-        #      engines that have uneven throttle ranges (need t4 constraint on mission
-        #      to truly fix)
-        self.global_throttle = True
-        self.global_hybrid_throttle = True
+        if Aircraft.Engine.GLOBAL_THROTTLE in options:
```
ehariton, line 187: So it looks like these options are exclusive, they could both be true or both be false?

ehariton (approved):

aviary/subsystems/mass/gasp_based/test/test_mass_summation.py
jkirk5: Looks like I was debugging this test at some point

aviary/subsystems/mass/gasp_based/test/test_mass_summation.py
```diff
@@ -3318,4 +3318,7 @@ def test_case1(self):
 
 
 if __name__ == "__main__":
-    unittest.main()
+    # unittest.main()
+    test = MassSummationTestCase1()
+    test.setUp()
+    test.test_case1()
```
jkirk5, line 3324:
```suggestion
    unittest.main()
```

aviary/subsystems/propulsion/engine_deck.py
jkirk5: Yeah you can have different values for hybrid vs. regular throttle

PR #595, opened 2024-10-30
Changing were dashboard looks for input files based on recent changes in how OpenMDAO saves reports

### LINKED ISSUE #586: Change the dashboard code to use the new locations for the reports and other output files in OpenMDAO
### Description

The latest version of OpenMDAO puts reports into a new directory structure. The current version of the dashboard in Aviary assumes the old directory structure.

Update the dashboard code to use the new OpenMDAO reports structure. Merge with 0.9.6-dev branch of Aviary

### Example

All runs of Aviary are affected. For example

aviary run_mission ./aviary/models/test_aircraft/aircraft_for_bench_FwGm.csv

### Aviary Version

0.9.6-dev

_No response_

hschilling:
### Summary

OpenMDAO recently changed the directory structure of the reports it generates. The Aviary dashboard gets its inputs from files in the reports directory. The code needed to be updated to look in the correct location. 

Also made greater use of the pathlib library when creating paths. 

Changed one tab so that it reports any problems in its generation in the tab rather than a warning to the command line.

### Related Issues

- Resolves #586 

Dashboard will not work if the reports are still in the old location of the reports in previous versions of OpenMDAO.

### New Dependencies

None

### DIFF
--- a/aviary/visualization/dashboard.py
+++ b/aviary/visualization/dashboard.py
@@ -174,18 +174,9 @@ def _dashboard_cmd(options, user_args):
 
     # check to see if options.script_name is a zip file
     # if yes, then unzip into reports directory and run dashboard on it
-    # TODO fix!!!
-    
-    # TODO need to save .db files in the zip file. 
-    
-
-    
     if zipfile.is_zipfile(options.script_name):
-        # report_dir_name = Path(options.script_name).stem
         report_dir_name = Path(options.script_name).stem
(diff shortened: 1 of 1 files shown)


crecine (commented):
Would it be possible to have this check both paths? We could prioritize the new method, but if we don't support the old openMDAO path, we will have to increment the minimum version.

Kenneth-T-Moore (approved):

jkirk5 (approved):

PR #597, opened 2024-11-04
Test geom builders

xjjiang:
### Summary

- Created two unit tests: the GASP based geom subsystem builder and the FLOPS based geom subsystem builder.
- Changed the default value of `Aircraft.HorizontalTail.ASPECT_RATIO` to 4.75 in `subsystems/geometry/flops_based/characteristic_lengths.py`
- Changed the default value of `Aircraft.Wing.FOLDED_SPAN` to 25 in `subsystems/geometry/flops_based/non_dimensional_conversion.py`
- Changed the default value of `Aircraft.HorizontalTail.TAPER_RATIO` to 0.352 in `subsystems/mass/flops_based/horizontal_tail.py`
- In `get_parameters()` method of `subsystems/geometry/geometry_builder.py`, excluded `__dict__` from `Aircraft.Nacelle.__dict__` because its value is Nacelle object which can not be handled by mappingproxy.
- Created two unit test files: `test_flops_mass_builder.py` and `test_gasp_mass_builder.py` in `subsystems/geometry/test`.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):
I'm not too big a fan of changing the openmdao defaults values in some geo components just so the unit test can be run without calling set_input_defaults manually. Still, this is only a problem because some components currently use those defaults. We should come up with a more uniform policy for setting them -- maybe  all 0 or 1, to make it obvious if they were forgotten in the csv.  Until then, I will approve the PR.

xjjiang:
> I'm not too big a fan of changing the openmdao defaults values in some geo components just so the unit test can be run without calling set_input_defaults manually. Still, this is only a problem because some components currently use those defaults. We should come up with a more uniform policy for setting them -- maybe all 0 or 1, to make it obvious if they were forgotten in the csv. Until then, I will approve the PR.

Agree. 
- In testing subsystem builders, there is no set_input_defaults available. Otherwise, I would have used it.
- Software development has a basic principle that code and data should be separated. But We have data everywhere. Since I am sure why we have those non-zero default values, I was worried that some non-zero default values are used somewhere and zero default values somewhere else. This inconsistency is not accepted by the builders.
- I agree that we should be consistent to set all 0 or 1. 
- I will change the defaults of the variables that I touched (including the ones in mass) to zeros and see if It breaks anything.

aviary/subsystems/geometry/gasp_based/non_dimensional_conversion.py
```diff
@@ -79,7 +79,7 @@ def setup(self):
         add_aviary_input(self, Aircraft.Wing.SPAN, val=0)
 
         if self.options["aviary_options"].get_val(Aircraft.Wing.FOLD_DIMENSIONAL_LOCATION_SPECIFIED, units='unitless'):
-            add_aviary_input(self, Aircraft.Wing.FOLDED_SPAN, val=0)
+            add_aviary_input(self, Aircraft.Wing.FOLDED_SPAN, val=25)
```
crecine, line 82: should probably add units here

aviary/subsystems/geometry/test/test_flops_geom_builder.py
```diff
@@ -0,0 +1,86 @@
...
+    That class inherits from TestSubsystemBuilder. So all the test functions are
+    within that inherited class. The setUp() method prepares the class and is run
+    before the test methods; then the test methods are run.
+    """
+
+    def setUp(self):
+        self.subsystem_builder = CoreGeometryBuilder(
+            'core_geometry',
+            BaseMetaData,
+            use_both_geometries=True,
```
crecine, line 28: should this be False for the non hybrid one?

aviary/subsystems/geometry/test/test_gasp_geom_builder.py
```diff
@@ -0,0 +1,86 @@
...
+    That class inherits from TestSubsystemBuilder. So all the test functions are
+    within that inherited class. The setUp() method prepares the class and is run
+    before the test methods; then the test methods are run.
+    """
+
+    def setUp(self):
+        self.subsystem_builder = CoreGeometryBuilder(
+            'core_geometry',
+            BaseMetaData,
+            use_both_geometries=True,
```
crecine, line 28: Should this be false?

crecine (approved):

crecine:
> * In testing subsystem builders, there is no set_input_defaults available. Otherwise, I would have used it.

can we use set_value instead of set_input_default for testing the subsystem builders?

aviary/subsystems/geometry/test/test_flops_geom_builder.py
xjjiang: Done. Theoretically, we should test all the scenarios of all the booleans on and off. But each case add twenty more tests. Should we do that?

aviary/subsystems/geometry/test/test_gasp_geom_builder.py
xjjiang: Changed to use_both_geometries=False. Should we test both?

xjjiang:
> > * In testing subsystem builders, there is no set_input_defaults available. Otherwise, I would have used it.
> 
> can we use set_value instead of set_input_default for testing the subsystem builders?

I tried. It does not work.

aviary/subsystems/geometry/test/test_flops_geom_builder.py
crecine: Maybe I'm misunderstanding what these tests are for, but I thought TestFLOPSGeomBuilder would be testing the pure FLOPS based geometry, TestGASPGeomBuilder would test the pure GASP based geometry, and then TestFLOPSGeomBuilderHybrid and TestGASPGeomBuilderHybrid would be testing when both are used.

xjjiang: I tried it but all unit test fail if `use_both_geometries=False`. I rolled it back.

crecine: when use_both_geometries is false, `code_origin` needs to be set to either LegacyCode.GASP or LegacyCode.FLOPS

aviary/subsystems/geometry/test/test_flops_geom_builder.py
```diff
@@ -0,0 +1,86 @@
...
+            code_origin_to_prioritize=FLOPS)
```
crecine, line 29:
```suggestion
        self.subsystem_builder = CoreGeometryBuilder(
            'core_geometry',
            BaseMetaData,
            use_both_geometries=False,
            code_origin=FLOPS,
            code_origin_to_prioritize=FLOPS)
```

PR #617, opened 2024-12-03
Improvements to DocTAPE

crecine:
### Summary

Added support for objects like the Aircraft and Mission hierarchies
to get_all_keys, get_value, and glue_keys.
Fixed a bug where setting display=False for glue_variable would display
<IPython.core.display.Markdown object> instead of nothing
Added support for including parentheses in the arguments for get_variable_name

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/utils/doctape.py
```diff
@@ -89,6 +89,7 @@ def get_previous_line(n=1) -> str:
 def get_variable_name(*variables) -> str:
     """
     returns the name of the variable passed to the function as a string
+    # NOTE: You cannot call this function multiple times on one line
```
crecine, line 92: In python 3.11 [inspect](https://docs.python.org/3.11/library/inspect.html#inspect.FrameInfo) adds `positions` as an attribute to `FrameInfo` that will allow for this

xjjiang:
Desired features:
- check if a blue variable exist;
- glue a numpy table.

crecine:
> * check if a blue variable exist;

Could you elaborate on that?
> * glue a numpy table.

do you mean a pandas dataframe? or are you asking for a feature that will automatically format a numpy array into a table?

xjjiang:
> Could you elaborate on that?

Yes, a pandas table. I have an example here:

https://github.com/xjjiang/om-Aviary/blob/glue_getting_started/aviary/docs/getting_started/onboarding_level2.ipynb

jkirk5 (approved):

ehariton (approved):

PR #618, opened 2024-12-03
Fix for failing multi-engine tests.

Kenneth-T-Moore:
### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/utils/preprocessors.py
```diff
@@ -397,6 +397,21 @@ def preprocess_propulsion(aviary_options: AviaryValues, engine_models: list = No
...
+        keys = [
+            Aircraft.Nacelle.LAMINAR_FLOW_LOWER,
+            Aircraft.Nacelle.LAMINAR_FLOW_UPPER
+        ]
+
+        for var in keys:
+            try:
+                aviary_options.get_val(var)
+            except KeyError:
+                aviary_options.set_val(var, np.zeros(num_engine_type))
```
crecine, line 413: I think this might be more clear?
```suggestion
            if key not in aviary_options:
                aviary_options.set_val(var, np.zeros(num_engine_type))
```

Kenneth-T-Moore:
Hotfixed by #622 for now. New solution will involve working with the defaults so that they are broadcast out to the correct size.

PR #623, opened 2024-12-09
Off Design Payload

cmbenne3:
### Summary

This PR has 3 parts:
1. Fixes the bug with GASP-based off design missions where variations in payload affect the design of the aircraft.

2. Enables GASP based missions to carry cargo whilst keeping backwards compatibility with the current Aviary functionality.

3. Updates the off design functions to handle the above changes.

**New cargo variables are created:**
Aircraft.CrewPayload.Design.CARGO_MASS - this is similar to the design passenger number variables and refers to the cargo carried on the aircraft for the design mission.
Aircraft.CrewPayload.MAX_CARGO_MASS - this is the maximum cargo the aircraft is designed to be capable of carrying. It is used in sizing the fuselage fuel body tank.

Aircraft.CrewPayload.CARGO_MASS - this is the current variable and is now repurposed as the 'as-flown' cargo mass for GASP-Based missions. This variable is used as a summary variable for FLOPS-Based missions as the sum of wing_cargo + misc_cargo.

preprocessors.py has been updated with logic to detect which of the above variables has been input by the user. If only CARGO_MASS is specified by the user then the analysis is assumed to be legacy Aviary and for backwards compatibility it is assumed that the user intends to set MAX_CARGO_MASS with zero cargo flown on as-flown or design missions.

The cargo variables for FLOPS-based missions remain unchanged - so currently Max and Design cargo mass are not used by flops based missions.

Due to the limitations of the passenger count logic in preprocessors it is not possible to set passenger counts to zero for off design missions.

### Comments from previous PR moved to here:
**Chris:**
Currently aviary only treats Aircraft.CrewPayload.CARGO_MASS as part of the design of the aircraft, and does not apply any of this mass to the gross mass of the aircraft when the mission is actually flown. If this is desired functionality then renaming this to 'Aircraft.CrewPayload.Design.CARGO_MASS' would seem sensible, but because this mass is not added to the aircraft for flight it would be different to the design passenger numbers (where that number of passengers is actually flown on the aircraft for that mission).

**Eliot:**
It looks like Aircraft.CrewPayload.CARGO_MASS has an effect on calculations for:
Aircraft.CrewPayload.CARGO_CONTAINER_MASS and payload_mass_max

Aircraft.CrewPayload.CARGO_CONTAINER_MASS goes into OperatingMass() to calculate Aircraft.Design.OPERATING_MASS. payload_mass_max is input to FuelMass() which outputs fuel_mass_min, which is then input to BodyTankCalculations() along with Aircraft.Design.OPERATING_MASS to calculatewingfuel_mass_min which is then input to FuelSysAndFullFuselageMass() to calculate fus_mass_full() which is input to FuselageAndStructureMass() to calculate Aircraft.Fuselage.MASS.

This large string of connections, and the resulting impact on Aircraft.Fuselage.MASS indicate that CARGO_MASS does have an impact on fuselage mass (at least in cases where fuel.py is used which I believe are GASP cases of mass-sizing). This would indicate that separating a design.CARGO_MASS and an as-flow cargo mass would be the correct approach.

Caution would need to be taken when importing GASP and FLOPS files if this change were implemented as the historical names which can be viewed in variable_meta_data.py for these values (i.e. INGASP.WCARGO) should be considered both the design values, with the as-flown values left blank. We could then assume that if as-flown is left blank, it indicates design is the same as as-flown. This is similar to how back-wards compatibility for design.passengers is handled as neither GASP nor FLOPS had concepts in their output files for the differentiation between as-flown and design passengers.

**Chris:**
I confirmed with @crecine that the current variable: Aircraft.CrewPayload.CARGO_MASS is implemented as a maximum cargo parameter, and this mass is NOT added to the aircraft gross mass for mission analysis. This copies one type of GASP analysis.
GASP does have the ability to fly missions with cargo, but this is not currently implemented in Aviary.
I plan to:
- Create a new "Aircraft.CrewPayload.Design.CARGO_MASS" parameter, which will exactly replace the current CARGO_MASS variable. Other option would be to name this "MAX_CARGO_MASS"
- Re-purpose "Aircraft.CrewPayload.CARGO_MASS" as an 'as-flown' cargo variable.
- If the user wants to replicate the current GASP implementation they would set Design.CARGO_MASS as they do currently for CARGO_MASS, and set re-purposed CARGO_MASS = 0 (i.e. don't fly the mission with any cargo).
- If they want to add cargo to a mission (design or off-design) then they would update the value of CARGO_MASS (and I will make some modifications to link this through to the mission analysis in the correct way).
- I will add some checks to ensure CARGO_MASS < design.CARGO_MASS

**Carl:**
Make sure that you have historical_name={"GASP": 'INGASP.WCARGO'} for the new design variable and historical_name={"GASP": None,} for the new as flown variable

I don't think you need to check that CARGO_MASS is less than Design.CARGO_MASS, I think it would be better to check that CARGO_MASS + PASSENGER_MASS is less than Design.CARGO_MASS + Design.PASSENGER_MASS, since it is possible to offload passengers in order to load more cargo


aviary/variable_info/variables.py
```diff
@@ -135,6 +136,7 @@ class Design:
             NUM_FIRST_CLASS = 'aircraft:crew_and_payload:design:num_first_class'
             NUM_TOURIST_CLASS = 'aircraft:crew_and_payload:design:num_tourist_class'
             NUM_PASSENGERS = 'aircraft:crew_and_payload:design:num_passengers'
+            CARGO_MASS = 'aircraft:crew_and_payload:design:cargo_mass'
```
Kenneth-T-Moore, line 139: I think these are supposed to be in alphabetical order.

aviary/utils/preprocessors.py
```diff
@@ -162,6 +163,120 @@ def preprocess_crewpayload(aviary_options: AviaryValues):
         raise om.AnalysisError(
             f"ERROR: In preprocesssors.py: NUM_PASSENGERS ({aviary_options.get_val(Aircraft.CrewPayload.NUM_PASSENGERS)}) is larger than the number of seats set by Design.NUM_PASSENGERS ({aviary_options.get_val(Aircraft.CrewPayload.Design.NUM_PASSENGERS)}) .")
 
+    # Check and process cargo variables from FLOPS
+    if Settings.MASS_METHOD in aviary_options:
+        mass_method = aviary_options.get_val(Settings.MASS_METHOD)
+    else:
+        print('ERROR! mass_method not specified cannot preprocess cargo inputs')
```
Kenneth-T-Moore, line 170: missing punct  between specified and cannot

aviary/utils/preprocessors.py
```diff
@@ -162,6 +163,120 @@ def preprocess_crewpayload(aviary_options: AviaryValues):
...
+        input_cargo = Aircraft.CrewPayload.CARGO_MASS in aviary_options
+        input_max_cargo = Aircraft.CrewPayload.MAX_CARGO_MASS in aviary_options
+        input_des_cargo = Aircraft.CrewPayload.Design.CARGO_MASS in aviary_options
+
+        if Settings.PROBLEM_TYPE in aviary_options:
+            problem_type = aviary_options.get_val(Settings.PROBLEM_TYPE)
+        else:
+            problem_type = ProblemType.SIZING
+
+        if input_cargo:
```
Kenneth-T-Moore, line 182: What would the value of input_cargo be if they don't set it? This expression will be true for both 0.0 or None, and I think you might want to be explicit about which one you want, unless the behavior is the same for unspecified, and specified at 0. This goes for all of the other conditionals as well.

aviary/utils/preprocessors.py
```diff
@@ -162,6 +163,120 @@ def preprocess_crewpayload(aviary_options: AviaryValues):
...
+
+        else:
+            # user has input no cargo information
+            cargo = max_cargo = des_cargo = 0
+            print(
+                'No CARGO variables detected, assume CARGO_MASS = MAX_CARGO_MASS = Design.CARGO_MASS = 0')
+
+        # check for potential cargo errors:
+        if cargo > des_cargo:
+            print('WARNING! as flown cargo > design cargo')
```
Kenneth-T-Moore, line 244: Some of these message might be candidates for filtering with Aviary's verbosity levels.

aviary/utils/preprocessors.py
```diff
@@ -162,6 +163,120 @@ def preprocess_crewpayload(aviary_options: AviaryValues):
...
+                Aircraft.CrewPayload.PASSENGER_MASS_WITH_BAGS, 'lbm')
+
+        # calculate and check total payload NOTE this is only used for error messaging the calculations for analysis are subsystems/mass/gasp_based:
+        design_passenger_payload_mass = design_num_pax * pax_mass_with_bag
+        des_payload = design_passenger_payload_mass + des_cargo
+        num_pax = aviary_options.get_val(Aircraft.CrewPayload.NUM_PASSENGERS)
+        as_flown_passenger_payload_mass = num_pax * pax_mass_with_bag
+        as_flown_payload = as_flown_passenger_payload_mass + cargo
+        if as_flown_payload > des_payload:
+            print('WARNING! as flown payload > design payload! Please re-design the aircraft!')
```
Kenneth-T-Moore, line 271: Yeah, i agree with just warning on this.

aviary/subsystems/mass/gasp_based/fuel.py
```diff
@@ -1317,7 +1317,7 @@ def compute(self, inputs, outputs):
             - control_wt
             - fixed_equip_wt
             - useful_wt
-            - payload_wt
+            - payload_wt_des
```
Kenneth-T-Moore, line 1320: There should be a corresponding change in the `compute_partials`.

FUEL_MASS_REQUIRED wrt payload_wt should become FUEL_MASS_REQUIRED wrt payload_wt_des

aviary/variable_info/variables.py
cmbenne3: Updated in correct order.
Maybe the test that checks for alphabetical order doesn't check the 4th level of the variable naming hierarchy.

aviary/utils/preprocessors.py
cmbenne3: Good spot - I have also updated this 'print' with raising an om.AnalysisError so the code stops executing.

aviary/subsystems/mass/gasp_based/fuel.py
cmbenne3: @Kenneth-T-Moore - I'm not 100% sure on what change I should make to the compute_partials section. There only seem to be partials computed for one of the 4 outputs (Mission.Design.FUEL_MASS) with respect to each of it's inputs. In an ideal world should we have partials for the other 3 outputs (Aircraft.Propulsion.MASS, Mission.Design.FUEL_MASS_REQUIRED and fuel_mass_min) for all of their inputs?

aviary/interface/methods_for_level2.py
```diff
@@ -2563,7 +2560,8 @@ def run_aviary_problem(self, record_filename="problem_history.db",
 
     def alternate_mission(self, run_mission=True,
                           json_filename='sizing_problem.json',
-                          payload_mass=None, mission_range=None,
+                          num_first=None, num_business=None, num_tourist=None, num_pax=None,
+                          wing_cargo=None, misc_cargo=None, cargo_mass=None, mission_range=None,
```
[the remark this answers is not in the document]
cmbenne3: This function is designed to run at level 2 after an aircraft has already been 'sized' by a design mission. The 'inputs' are loaded from the json file, and so the user doesn't have an opportunity to alter them by modifying a .csv. In this setup the user can setup the desired 'off design' mission by specifying arguments in this function. I think a further development might be to allow a level 1 user to run off design, by specifying 'off design' payload range variables in the .csv input, but this hasn't been built yet.

aviary/utils/preprocessors.py
cmbenne3: I have added a check for verbosity >=1 to suppress the print messages when user has set verbosity to quiet

aviary/subsystems/mass/gasp_based/fuel.py
Kenneth-T-Moore: I just noticed that the deriv is defined in declare_partials as -1, so this is already taken care of. Disregard.

Kenneth-T-Moore (approved):

aviary/variable_info/variable_meta_data.py
```diff
@@ -838,6 +843,23 @@
     default_value=165.0,
 )
 
+add_meta_data(
+    Aircraft.CrewPayload.MAX_CARGO_MASS,
```
jkirk5, line 847: Max cargo mass sounds like it should be in Design to me - the maximum cargo capacity can't change from mission to mission and is set during the aircraft sizing correct? If I'm wrong then I think we should add more clarification to the description to make it clearer

jkirk5 (approved):

aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -269,21 +274,27 @@ def setup(self):
...
+                              Aircraft.CrewPayload.CARGO_MASS], val=1.0)
+        self.declare_partials("payload_mass_des", [
+                              Aircraft.CrewPayload.Design.CARGO_MASS], val=1.0)
+        self.declare_partials("payload_mass_max", [
+                              Aircraft.CrewPayload.MAX_CARGO_MASS], val=1.0)
 
     def compute(self, inputs, outputs):
         pax_mass, _ = self.options[Aircraft.CrewPayload.PASSENGER_MASS_WITH_BAGS]
         PAX = self.options[Aircraft.CrewPayload.NUM_PASSENGERS]
+        PAX_des = self.options[Aircraft.CrewPayload.Design.NUM_PASSENGERS]
```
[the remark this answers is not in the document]
cmbenne3: Done!

aviary/variable_info/variable_meta_data.py
cmbenne3: it certainly doesn't make sense to have an 'as-flown' max_cargo_mass... so you're correct that this should probably live under Aircraft.CrewPayload.Design.MAX_CARGO_MASS instead. I have updated this in all locations in the aviary code.

jkirk5 (approved):

PR #625, opened 2024-12-09
Fix 0 pax bug for multi-mission

jkirk5:
### Summary

Fixes a bug in the crew & payload preprocessor where providing a passenger count of 0 for a single mission was confused with not providing that particular passenger count variable at all, leading to overwriting mission pax count with design pax count.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/utils/preprocessors.py
+++ b/aviary/utils/preprocessors.py
@@ -6,7 +6,7 @@ import openmdao.api as om
 from aviary.utils.aviary_values import AviaryValues
 from aviary.utils.named_values import get_keys
 from aviary.variable_info.variable_meta_data import _MetaData
-from aviary.variable_info.variables import Aircraft, Mission
+from aviary.variable_info.variables import Aircraft, Mission, Settings
 from aviary.utils.test_utils.variable_test import get_names_from_hierarchy
 from aviary.variable_info.variable_meta_data import _MetaData as BaseMetaData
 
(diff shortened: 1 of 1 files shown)


ehariton:
test_ben_FwFm.py fails on this PR.

This code detects:
pax_provided True
design_pax_provided True

even though only Design information is provided in `aircraft_for_bench_FwFm.csv`

ehariton (changes requested):
test_ben_FwFm.py fails on this PR.

This code detects:
pax_provided True
design_pax_provided True

even though only Design information is provided in aircraft_for_bench_FwFm.csv

jkirk5:
I will need to update docs once I have tests working

aviary/utils/preprocessors.py
```diff
@@ -35,142 +35,201 @@ def preprocess_crewpayload(aviary_options: AviaryValues):
...
+                    "passenger classes for the flown mission."
+                )
+            aviary_options.set_val(Aircraft.CrewPayload.NUM_PASSENGERS, mission_sum)
+
+    # Check cases where mission passengers are greater than design passengers
+    design_pax = aviary_options.get_val(Aircraft.CrewPayload.Design.NUM_PASSENGERS)
+    mission_pax = aviary_options.get_val(Aircraft.CrewPayload.NUM_PASSENGERS)
+    if mission_pax > design_pax:
+        if verbosity >= 1:
+            # TODO Should this be an error?
```
Kenneth-T-Moore, line 166: yeah, i would just warn on this, just in case the user wants to run a mission overloaded with passengers (though where do they sit?)

Kenneth-T-Moore (approved):

aviary/utils/preprocessors.py
jkirk5: https://www.guinnessworldrecords.com/world-records/most-passengers-on-an-aircraft
Over 1k people were crammed on a 747 once! So we'd be able to model that flight now

jkirk5:
> Examining `aviary_options` what we find is that `'aircraft:crew_and_payload:num_business_class': (0, 'unitless'),` has been set before getting to the following code lines in preprocessor.py:

It appears it is an issue stemming from the way `load_vehicle()` works, which is to grab every single variable from the metadata with a default and apply it to the aviary inputs. I think this is really bad because it has the potential to accidentally set default values as overrides which 100% isn't intended behavior. It also is really overkill, we only need to set values for inputs the user wants - every other default should come directly from the metadata when adding component inputs/outputs.

This PR may need to stay on hold until we clean up how defaults work in the rest of Aviary, this approach should work with the new paradigm (all defaults come from metadata rather than using user inputs as a vehicle to get those values into the problem)

aviary/docs/examples/off_design_example.ipynb
```diff
@@ -25,37 +25,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The methods used to run the off-design missions accept the mission parameters `payload_mass`, `mission_range`, `mission_mass`, and `phase_info` as inputs depending on the type of mission being run.\n",
+    "The methods used to run the off-design missions accept a variety of arguments including `phase_info`, some mission configuration variables such as `equations_of_motion`, and inputs that depend on the type of mission being run such as payload, passenger count, and range.\n",
```
ehariton, line 28: Isn't this `mission_info` now? or is mission info different than `phase_info` ?

aviary/docs/user_guide/payload_range_functionality.ipynb
```diff
@@ -44,199 +46,63 @@
...
+    "A payload-range diagram is an aircraft performance visualization that illustrates the trade-off between how much mass the aircraft can carry in terms of passengers and cargo (payload) vs how far it can fly (range).\n",
     "The payload-range shows 3 different points:\n",
-    "- **Maximum Payload Point**: The maximum structural payload the aircraft can carry with sufficient fuel to achieve its design range, this is the 'sizing' mission.\n",
-    "- **Maximum Fuel Point**: The maximum fuel the aircraft can hold with additional payload up to the aircraft's max takeoff weight. \n",
-    "- **Ferry Range Point**: The maximum fuel the aircraft can carry with zero payload to achieve maximum possible range. \n",
+    "- **Maximum Payload Point (\"Design Range\")**: The range the aircraft can fly with maximum payload mass and fuel mass up to the maximum gross mass. This range is typically the design point of the aircraft (the \"sizing\" mission).\n",
+    "- **Maximum Fuel Point (\"Max Economic Range\")**: The range the aircraft can fly at maximum fuel capacity and payload mass up to maximum gross mass. This is the furthest range the aircraft can fly at its maximum gross mass.\n",
```
ehariton, line 54: I thought this was called "max fuel" point not "max economic range" ? The wording here doesn't make me feel great and I mis interpreted it the first time I read it. I might go for something like:
The range the aircraft can fly at where the fuel tanks are completely full and payload mass is added until the aircraft reaches it's maximum gross mass.

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1244,513 @@ def run_aviary_problem(
...
-        num_pax=None,
-        wing_cargo=None,
-        misc_cargo=None,
-        cargo_mass=None,
-        mission_mass=None,
-        phase_info=None,
-        verbosity=None,
-    ):
+        off_design_prob.check_and_preprocess_inputs(verbosity=verbosity)
+        off_design_prob.add_pre_mission_systems(verbosity=verbosity)
```
ehariton: Remember you can use `off_design_prob.build_model()` to execute add_pre, add_phase, add_post, and link_phases!

aviary/docs/user_guide/payload_range_functionality.ipynb
jkirk5: This change is actually already in main from the off-design overhaul, but I'll comment here.

I'm not sure I follow, the point is still directly named "Maximum Fuel Point" here. Each point has a name, and mentions the particular range that point corresponds to (e.g. flying the max payload point falls out design range) as additional information. The only other name I've ever heard to describe this range is "harmonic range", which to me is super unhelpful. I chose "Max Economic Range" based on the fact off-design missions that describe actual operating cases are often referred to as "economic missions". "Max Economic Range" is meant to refer to the maximum range of the aircraft at full utilization (and therefore maximizing profitability at that range). I'm open to a better name for this specific range since there doesn't seem to be a good official one.

And the sentence I have seems to me to say the exact thing you suggested? "fuel tanks completely full" == "at maximum fuel capacity", "payload mass added until reaches maximum gross mass" == "payload mass up to maximum gross mass"

aviary/interface/methods_for_level2.py
jkirk5: Right! Didn't exist when I first wrote this code, I've streamlined it now!

aviary/interface/methods_for_level2.py
```diff
@@ -1456,11 +1467,13 @@ def run_off_design_mission(
                 Mission.Summary.GROSS_MASS, mission_gross_mass, units='lbm'
             )
 
-        off_design_prob.check_and_preprocess_inputs(verbosity=verbosity)
-        off_design_prob.add_pre_mission_systems(verbosity=verbosity)
-        off_design_prob.add_phases(verbosity=verbosity)
-        off_design_prob.add_post_mission_systems(verbosity=verbosity)
-        off_design_prob.link_phases(verbosity=verbosity)
+        # off_design_prob.check_and_preprocess_inputs(verbosity=verbosity)
```
Kenneth-T-Moore: `check_and_preprocess_inputs` is not part of `build_model`, so we need to keep it here.

jkirk5: Oh right, didn't check too carefully when I made that change. Reverted that

aviary/interface/methods_for_level2.py
```diff
@@ -1563,13 +1576,10 @@ def run_payload_range(self, verbosity=None):
...
 
-        # Off-design missions do not currently work for GASP masses or missions.
+        # Off-design missions are not tested with 2DOF missions.
         mass_method = self.model.aviary_inputs.get_val(Settings.MASS_METHOD)
         equations_of_motion = self.model.aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
-        if (
-            mass_method == LegacyCode.FLOPS
-            and equations_of_motion is EquationsOfMotion.HEIGHT_ENERGY
-        ):
+        if equations_of_motion is EquationsOfMotion.HEIGHT_ENERGY:
```
ehariton, line 1582: @Kenneth-T-Moore, considering the simplification here we may want to push all the below into the EOM builder.

ehariton (approved):

PR #630, opened 2024-12-12
Update value of Mission.Summary.RANGE for fallout missions

cmbenne3:
### Summary

This PR makes some changes to methods for level 2 to remove the variable "actual_range" and replace it with Mission.Summary.RANGE

This is particularly helpful for fallout missions where this value was not being updated as expected.

### Related Issues

None

None

### New Dependencies

None


cmbenne3:
@jkirk5 @Kenneth-T-Moore:
I ran a comparison of prob.model.list_vars for 'main' compared to this branch 'fallout_summary_range' and think that the value used for the near equal check of Mission.Summary.RANGE needs to be updated.

image

The check that fails is:
image

This is because in 'main' the Mission.Summary.Range value is simply overwritten with the 'Target Range' value, which is just incorrect it doesn't actually tell you anything about the values output from the optimizer on how far that aircraft has actually flown . I don't know much about the shooting method and so I'm not sure why the range residual hasn't converged to zero in either case. This probably needs to be investigated further, but perhaps as a separate issue. For now what are your thoughts on updating the assert near equal value to 3765.48?

ehariton (approved):

PR #632, opened 2024-12-13
Resolve duplicate N2 report generation

### LINKED ISSUE #587: Double creation of n2.html file
### Description

When you run a script like

`aviary run_mission ./aviary/models/test_aircraft/aircraft_for_bench_FwGm.csv`
 
two n2.html files were created.
 
`output/aircraft_for_bench_FwGm/n2/html`
 
`aircraft_for_bench_FwGm_out/reports/n2.html`

But there should only be one. 

A theory on what is happening: a step was added to generate the N2 after model run so it is populated with the actual values from optimization - this is supposed to overwrite the N2 created before model run, but with the new openmdao version that means they are getting put in different places 
 
The code that generates the updated N2 is made in `methods_for_level_1.py` on line 143. Not exactly sure where `outdir` is pointing to, it is modified and adjusted in a few places. So it seems like there is Aviary stuff that wasn't using the reports system, so it is putting stuff in the "wrong" place. Maybe a better long-term solution is to move that to an actual report so it always gets put in the same place as the other stuff
 

### Example

Run this script

`aviary run_mission ./aviary/models/test_aircraft/aircraft_for_bench_FwGm.csv`

### Aviary Version

0.9.4-dev

_No response_

jkirk5:
### Summary

Rework L1 interface to always route n2 generation to openmdao report directory. This means the `outdir` argument for `run_mission` is now removed, as it only affected where this post-mission generated n2 goes

### Related Issues

- Resolves #587

None

### New Dependencies

None


Kenneth-T-Moore (approved):
Approved, but you can remove the commented-out argument.

PR #635, opened 2024-12-23
Set FLOPS mass default

xjjiang:
### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/variable_info/functions.py
+++ b/aviary/variable_info/functions.py
@@ -34,17 +34,21 @@ def add_aviary_input(comp, varname, val=None, units=None, desc=None, shape_by_co
     else:
         input_desc = meta['desc']
     if val is None:
-        val = meta['default_value']
-        if val is None:
-            if shape is None:
+        if shape is None:
+            val = meta['default_value']
+            if val is None:
                 val = 0.0
-            else:
+        else:
+            val = meta['default_value']
+            if val is None:
                 val = np.zeros(shape)
+            else:
+                val = np.ones(shape) * val
     comp.add_input(varname, val=val, units=input_units,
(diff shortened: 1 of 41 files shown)


aviary/variable_info/functions.py
```diff
@@ -33,12 +34,21 @@ def add_aviary_input(comp, varname, val=None, units=None, desc=None, shape_by_co
     else:
         input_desc = meta['desc']
     if val is None:
-        val = meta['default_value']
+        if shape is None:
+            val = meta['default_value']
+            if val is None:
```
Kenneth-T-Moore, line 39: Need to rework the logic here, because val is always None because the entire block is under another `if val is None` check.

aviary/variable_info/functions.py
```diff
@@ -58,6 +68,17 @@ def add_aviary_output(comp, varname, val, units=None, desc=None, shape_by_conn=F
         output_desc = desc
     else:
         output_desc = meta['desc']
+    if val is None:
+        if shape is None:
+            val = meta['default_value']
+            if val is None:
```
Kenneth-T-Moore, line 74: Rework here too, double val is None check.

aviary/variable_info/functions.py
xjjiang: The second `if val is None` is after `val = meta['default_value` and val may not be `None` anymore.

Kenneth-T-Moore: Oh, I see, that makes sense.

Kenneth-T-Moore (approved):

PR #648, opened 2025-01-22
Optional subsystems

jkirk5:
### Summary

Extends the capability for a subsystem to be "skipped" via the `external` method flag in `phase_info` to geometry and mass. The `external` method also now skips both pre-mission and mission.

None

### New Dependencies

None


aviary/docs/user_guide/mass.ipynb
```diff
@@ -44,7 +44,10 @@
     "Which variables are used depends on which mass estimation subsystem you're using, such as `aircraft:crew_and_payload:mass_per_passenger`, `aircraft:engine:additional_mass_fraction`, etc.\n",
     "\n",
     "Aviary allows for extensive customization and extensions within the mass subsystem.\n",
-    "Users can develop alternative components and integrate them as subsystems with the existing platform."
+    "Users can develop alternative components and integrate them as subsystems with the existing platform.\n",
+    "\n",
+    "## External Mass\n",
+    "Specifying the `external` mass method disables Aviary's core mass group. This allows for external subsystems to completely replace these calculations."
```
Kenneth-T-Moore, line 50: Maybe we should stress that, yeah, this replaces all mass calculations, so you don't want to use it for subsystem mass.

Also, it might be worth mentioning that, to replace the mass subsystem for the sizing problem, your new mass group needs to compute the Aircraft.Design.OPERATING_MASS, which feeds into the downstream constraints.

jkirk5: We'll need to do that for every subsystem, as well as provide lists of their required top-level inputs and primary outputs - I'm going to add that to our documentation to-do list outside the scope of this PR

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -100,14 +100,20 @@ def __init__(self, name=None, meta_data=None, code_origin=None):
...
         code_origin = self.code_origin
+        try:
+            method = kwargs.pop('method')
+        except KeyError:
+            method = None
+        aero_group = None
 
-        if code_origin is GASP:
-            aero_group = PreMissionAero()
+        if method != 'external':
```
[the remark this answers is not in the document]
jkirk5: done

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -98,91 +98,110 @@ def __init__(self, name=None, meta_data=None, code_origin=None):
...
+                if (
+                    any(
+                        key in kwargs
+                        for key in [
+                            'free_aero_data',
+                            'free_flaps_data',
+                            'free_ground_data',
+                        ]
+                    )
+                    in kwargs
```
Kenneth-T-Moore, line 187: Doesn't this extra "in kwargs" negate the entire logic?

    z = ['a', 'b', 'c']
    any(key in z for key in ['a', 'e']) 
    True
    any(key in z for key in ['a', 'e']) in z
    False

ehariton (approved):

aviary/subsystems/aerodynamics/aerodynamics_builder.py
jkirk5: Agreed that this logic seems wrong. I've removed the extra "in kwargs"

Kenneth-T-Moore (approved):

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -98,91 +98,112 @@ def __init__(self, name=None, meta_data=None, code_origin=None):
...
     def build_mission(self, num_nodes, aviary_inputs, **kwargs):
         try:
             method = kwargs.pop('method')
         except KeyError:
             method = None
+
+        aero_group = None
+
+        if method is 'external':
+            return None
```
jkirk5, line 126:
```suggestion
        if method is 'external':
            return None
        
        aero_group = None
```

PR #651, opened 2025-01-24
Methods for level 2 revamp part 1.

Kenneth-T-Moore:
### Summary

1. Moved PreMissiongroup, PostMissionGroup and AviaryGroup into their own files.
2. Created a new object called a Problem Builder which builds parts of an aviary problem that are unique to a specific `mission_method`. 

This is a first pass at some refactoring that will allow for more customization. 

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/interface/methods_for_level2.py
```diff
@@ -274,24 +115,40 @@ def load_inputs(
...
+        elif mission_method is CUSTOM:
+            if problem_builder:
+                self.builder = problem_builder()
+                # TODO: make draft / example custom builder
+            else:
+                raise ValueError(
+                    f'When using "settings:equations_of_motion,custom", a problem_builder must be specified in load_inputs().')
+        else:
+            raise ValueError(
+                f'settings:equations_of_motion must be one of: height_energy, 2DOF, or custom')
```
ehariton, line 147: SOLVED_2DOF should be added here per line 136

aviary/mission/problem_builder_height_energy.py
```diff
@@ -0,0 +1,443 @@
...
+
+        try:
+            fix_duration = user_options.get_val('fix_duration')
+        except KeyError:
+            fix_duration = False
+
+        # The rest of the phases includes all Height Energy method phases
+        # and any 2DOF phases that don't fall into the naming patterns
+        # above.
+        input_initial = False
```
ehariton, line 118: consolidate with line 139 which also sets `input_initial`

aviary/mission/problem_builder_height_energy.py
```diff
@@ -0,0 +1,443 @@
...
+            phase.set_time_options(
+                fix_initial=fix_initial, fix_duration=fix_duration, units=time_units,
+                duration_bounds=user_options.get_val("duration_bounds", time_units),
+                duration_ref=user_options.get_val("duration_ref", time_units),
+                initial_ref=initial_ref,
+            )
+        elif phase_name == 'descent':  # TODO: generalize this logic for all phases
+            phase.set_time_options(
+                fix_initial=False, fix_duration=False, units=time_units,
+                duration_bounds=user_options.get_val("duration_bounds", time_units),
```
ehariton, line 161: The only difference between this elif and the code before it is that if the phase_name is 'descent' then fix_duration is set to False, no matter what the user input. That is a check that should be happening on the phase inputs, not here. That would allow us to delete lines 158 through 165.

Honestly a bit strange to me that we specifically need fix_duration = False in descent.

aviary/mission/problem_builder_height_energy.py
```diff
@@ -0,0 +1,443 @@
...
+                initial_ref = user_options.get_val("initial_ref", time_units)
+            else:
+                # Redundant on a fixed input; raises a warning if specified.
+                initial_ref = None
+
+            phase.set_time_options(
+                fix_initial=fix_initial, fix_duration=fix_duration, units=time_units,
+                duration_bounds=user_options.get_val("duration_bounds", time_units),
+                duration_ref=user_options.get_val("duration_ref", time_units),
+                initial_ref=initial_ref,
```
ehariton, line 156: why are we not applying the initial_bounds that the user supplied here??

aviary/interface/methods_for_level2.py
Kenneth-T-Moore: Added

aviary/mission/problem_builder_height_energy.py
Kenneth-T-Moore: Consolidated into `input_initial = phase_idx > 0`

aviary/mission/problem_builder_height_energy.py
Kenneth-T-Moore: Yeah, that is strange too, plus it is the only example where we have to adhere to a naming convention in HE.  It looks like fix_duration is always false in our phase_infos, and the ones in MBSAE.  If the user sets `target_duration` in a phase, then fix_duration is automatically set to True (we have so much interaction between these attributes), but in those cases, the phase is named something like "reserve_hold" instead of "descent".  I am going to remove this "if" branch and see what happens.

aviary/mission/problem_builder_height_energy.py
Kenneth-T-Moore: It is potentially an oversight. We should treat initial_bounds just like initial_ref -- none unless mpi.

aviary/interface/methods_for_level2.py
```diff
@@ -551,7 +297,8 @@ def check_and_preprocess_inputs(self):
         elif mission_method is TWO_DEGREES_OF_FREEDOM:
```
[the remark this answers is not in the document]
Kenneth-T-Moore: Yeah, this is one of the areas that is still pretty messy. It makes me wonder if we should move all of the core pre-mission builder setup into the problem builders.  Concepts of FLOPS and GASP maybe shouldn't exist at this level.

jkirk5 (approved):
Additional sugestions:
Change name to ProblemConfigurator (suggested file names, `two_dof_problem_configurator.py`, `energy_problem_configurator.py`)
Create a base class for the configurators to define the interface

aviary/core/__init__.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.9.7-dev"
```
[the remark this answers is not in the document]
Kenneth-T-Moore: Good catch. This was definitely accidental.

Kenneth-T-Moore:
Latest updates fixed the naming of the new classes, bringing them in line with standards. Also added a base class.

aviary/interface/methods_for_level2.py
```diff
@@ -346,127 +195,22 @@ def load_inputs(
...
-                    phase_info['post_mission']['target_range'], 'NM')
-            else:
-                self.require_range_residual = False
-                # still instantiate target_range because it is used for default guesses
-                # for phase comps
-                self.target_range = aviary_inputs.get_val(
-                    Mission.Design.RANGE, units='NM')
+        self.builder.initial_guesses(self)
+        # This function sets all the following defaults if they were not already set
+        # self.engine_builders, self.mass_method, self.pre_mission_info, self_post_mission_info
```
ehariton, line 209: Looks like mass_method is not dealt with in here after all, we deal with it further up

aviary/docs/examples/modified_aircraft.csv
```diff
@@ -49,6 +49,7 @@ aircraft:engine:num_fuselage_engines,0,unitless
 aircraft:engine:num_wing_engines,2,unitless
 aircraft:engine:reference_mass,7400,lbm
 aircraft:engine:reference_sls_thrust,28928.1,lbf
+aircraft:engine:scale_factor,1.0,unitless
```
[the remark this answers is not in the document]
Kenneth-T-Moore: This is just a fix related to a change that Jason made a while ago, where engine scale_factor is a required input now instead of an output. I must have fixed it when i was debugging one of the tests.

aviary/interface/methods_for_level2.py
Kenneth-T-Moore: Thanks, I have removed it.

aviary/mission/height_energy_problem_configurator.py
```diff
@@ -0,0 +1,625 @@
...
+    use_new_dymos_syntax = True
+
+
+class HeightEnergyProblemConfigurator(ProblemConfiguratorBase):
+    """
+    A Height-Energy specific builder that customizes AviaryProblem() for use with
+    height energy phases.
+    """
+
+    def initial_guesses(self, prob):
```
ehariton, line 30: We have both `initial_guesses()` (Set any initial guesses for variables in the aviary problem) and `add_guesses()` (Adds the initial guesses for each variable of a given phase to the problem.). I'm wondering if we can consolidate? I'm sure that the reason we have two is because we setup guesses in two places in M4L2

ehariton (approved):

aviary/mission/height_energy_problem_configurator.py
Kenneth-T-Moore: It is kind of weird, and something we didn't really notice until we pulled the code out. I think untangling all of these strands is going to be difficult, but will result in more maintainable code.

PR #653, opened 2025-01-25
Set aero default again

xjjiang:
### Summary

This is a draft.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):
This is a good cleanup. I think one thing I would do to be on the safe side is to take the models in `test_bench_GwGm.py` and `test_bench_FwFm.py`, and do a run_model (instead of the run_aviary_problem) followed by a `prob.model.list_vars(units=True, print_arrays=True)`, and save that output for `main` and for this branch, and do a diff. This is just to verify that we aren't missing a value in the csv file for something that used to have a default value.  I don't expect there will be any differences, but it is still good to check.

EDIT - given the test failures, maybe there is something missing, and this is a good way to find it.

PR #663, opened 2025-01-30
Aircraft.Wing.AIRFOIL_TECHNOLOGY `types` is inconsistent with all other numeric vars.

Kenneth-T-Moore:
### Summary

Aircraft.Wing.AIRFOIL_TECHNOLOGY is the only numeric variable that we explicitly declared as float. This causes some problems if you set it to 1.  We should allow both floats and ints, since we allow it for all other vars (e.g. Aircraft.Wing.WEIGHT_SCALER=1). Removing the `types` declaration brings it in line with the others.

Note that it is a separate issue that our metadata dict might be enforcing type too loosely with numeric vars, though I think the csv reader and the AviaryValues evaluators were written with this in mind. Any future refactor needs to consider all of these together.

### Related Issues

- Resolves #

None

### New Dependencies

None


ehariton:
searching through variable_meta_data.py I found that `Mission.Design.CRUISE_ALTITUDE` and `Aircraft.Engine.ADDITIONAL_MASS_FRACTION` were also declared as floats.l

Kenneth-T-Moore:
> `CRUISE_ALTITUDE`

Thanks. I've removed the explicit typing from Cruise Altitude. I kept it in the Additional Mass Fraction because it is one of those weird propulsion variables that can be a list.

jkirk5 (approved):

xjjiang:
My PR #676 will allow `types` stay. `Aircraft.Wing.WEIGHT_SCALER` should be `Aircraft.Wing.MASS_SCALER`. All of the variables that are supposed to be float can be described as `types=float` after PR #676.

PR #666, opened 2025-02-06
Set aero defaults and units

xjjiang:
### Summary

Setting most variables to their `default_value`s as in meta data. Keep showing units in components.

Two exceptions:

1. In `subsystems/aerodynamics/flops_based/mux_component.py`:
      - `Aircraft.Wing.CHARACTERISTIC_LENGTH`
      - `Aircraft.HorizontalTail.CHARACTERISTIC_LENGTH`
      - `Aircraft.VerticalTail.CHARACTERISTIC_LENGTH`
      - `Aircraft.Fuselage.CHARACTERISTIC_LENGTH`

If I set them to default values 0.0, `docs/getting_started/onboarding_level3.ipynb` fails.

2. In `variable_info/variable_meta_data.py`:
      - `Aircraft.Wing.FORM_FACTOR`
      - `Aircraft.VerticalTail.FORM_FACTOR`
      - `Aircraft.HorizontalTail.FORM_FACTOR`

If I set their `default_value`s to 0.0 instead of 1.25, `validation_cases/benchmark_tests/test_bench_GwGm.py:test_bench_GwGm_shooting()` fails. This needs further investigation.

In `validation_cases/benchmark_tests/test_bench_GwGm.py`, I updated the desired `Mission.Summary.TOTAL_FUEL_MASS` to `42109.1488` from `42529.0` in `test_bench_GwGm_SNOPT()` and `test_bench_GwGm_SNOPT_lbm_s()`. This is because the outputs from main are approximately `42109.1488`. My outputs are `42094.97734613` and `42089.4709307` are close enough to the new desired value. But when compared to `42529.0`, they are a little bit off for the relative tolerance 0.01.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):

aviary/mission/gasp_based/test/test_idle_descent_estimation.py
```diff
@@ -70,6 +74,7 @@ def test_subproblem(self):
         prob.model.promotes('idle_descent_estimation', inputs=['parameters:*'])
 
         prob.setup()
+        prob.set_val(Aircraft.Wing.FORM_FACTOR, 1.25)
```
ehariton, line 77: Why are you setting this twice? You already set it in line 33?

xjjiang: You are right. I just removed it.

aviary/subsystems/aerodynamics/flops_based/test/test_tabular_aero_group.py
```diff
@@ -194,6 +197,9 @@ def setUp(self):
 
         setup_model_options(self.prob, aviary_options)
 
+        self.prob.model.set_input_defaults(Dynamic.Atmosphere.MACH,
+                                           val=0.3876, units='unitless')
```
ehariton, line 201: did you mean to set these values here? This doesn't seem to be covered in the original intent of the PR.

aviary/subsystems/aerodynamics/gasp_based/test/test_table_based.py
```diff
@@ -194,6 +197,7 @@ def test_case(self):
             promotes_inputs=['*'],
             promotes_outputs=['*'],
         )
+        # prob.model.set_input_defaults(Aircraft.Wing.AREA, val=1370.3)
```
ehariton, line 200: Do we need this comment?

ehariton (approved):

aviary/subsystems/aerodynamics/flops_based/test/test_tabular_aero_group.py
xjjiang: Yes, they need to be set here. Otherwise, I will get the following error:

```<model> <class Group>: The following inputs, ['aero.CD0_interp.mach', 'aero.CDI_interp.mach', 'aero.drag.simple_CD.mach'], promoted to 'mach', are connected but their metadata entries ['val'] differ. Call <group>.set_input_defaults('mach', val=?), where <group> is the Group named 'aero' to remove the ambiguity.```

This is because I removed the default values from individual components. 

I removed the line `self.prob.set_val(Dynamic.Atmosphere.MACH, val=0.3876, units='unitless')` that follows.

jkirk5 (changes requested):
Putting this in as a merge hold, I'd like to investigate the changed benchmark values first

xjjiang:
> Putting this in as a merge hold, I'd like to investigate the changed benchmark values first

Agree. I can investigate the source in my code that changes the number, but the result from main is already too far from 42529.0: 

(42529. - 42109.1488)/42529 = 0.009872.

Our tolerance 0.01 is probably too large.

jkirk5:
> > Putting this in as a merge hold, I'd like to investigate the changed benchmark values first
> 
> Agree. I can investigate the source in my code that changes the number, but the result from main is already too far from 42529.0:
> 
> (42529. - 42109.1488)/42529 = 0.009872.
> 
> Our tolerance 0.01 is probably too large.

Agreed. I am currently trying to double check if your code closely matches main, and if it does then this change is ok!

jkirk5 (approved):

jkirk5:
I think I've fixed all the exceptions you've found, as well as update the GwGm bench values

PR #668, opened 2025-02-11
Universal ODE class

jkirk5:
### Summary

Creates a universal ODE class for all ODEs in Aviary (BaseODE), renaming the base ODEs for energy and 2DOF methods and having them inherit this new class.

Added external subsystems support to all ODEs through BaseODE interface, with new unittest verifying this in 2DOF cruise phase (mirroring energy method test)

Cleanup and remove redundant code in all ODEs and phase builders

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/mission/flops_based/phases/detailed_takeoff_phases.py
```diff
@@ -139,15 +139,10 @@ class attribute: derived type customization point; the default value
     validate_options
     assign_default_options
     '''
-    __slots__ = ()
```
Kenneth-T-Moore, line 142: Not sure why you removed the slots here and kept them in landing. I know these are supposed to be more efficient that self.dict lookup, though we rarely think about using them.

Kenneth-T-Moore (approved):
This looks pretty clean.

aviary/mission/flops_based/phases/detailed_takeoff_phases.py
jkirk5: I think I meant to leave them in here, will fix. Someday we'll have to do a polish run for at least our biggest classes to uniformly implement features like slots and the like

xjjiang:
If you want a universal base ODE, why don't you modify other FLOPS based ODEs?

jkirk5:
> If you want a universal base ODE, why don't you modify other FLOPS based ODEs?

Can you clarify? Which ODEs are you referring to?

PR #669, opened 2025-02-12
Fixed a bug where 2 variables were not being promoted as parameters from the mission ode.

Kenneth-T-Moore:
This fixes an issue that Xun found on his PR that removes defaults for aero components. It turns out that a couple of variable promotions were missing from the get_parameters in the core aero builder. Our tests masked this because they tested using the same aircraft as the default values.

### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/subsystems/aerodynamics/test/test_gasp_drag.py
```diff
@@ -0,0 +1,55 @@
+"""
+Tests the aero builder with the 2dof (gasp-based) drag.
```
jkirk5, line 2: It's not clear that's what this test is doing without the context of this PR - it's just a standard Aviary problem using a standard bench csv and full mission profile, but the name and class imply it's specifically an aero test. Can you elaborate more on what specifically is getting tested here and why?

jkirk5 (approved):

xjjiang:
I tested my code. This PR works for the changes that I need.

PR #675, opened 2025-02-24
Split run_examples into separate tests.

Kenneth-T-Moore:
### Summary

This splits the run_examples tests into separate tests for each example, which will improve the testflo runtime for the tests, and allow us to find and address specific issues.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/examples/test/test_all_examples.py
+++ b/aviary/examples/test/test_all_examples.py
@@ -35,6 +35,7 @@ def find_examples():
                 run_files.append(Path(root) / file)
     return run_files
 
+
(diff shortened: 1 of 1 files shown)


xjjiang:
Several unit tests are skipped. Is it possible not to skip them?

Kenneth-T-Moore:
> Several unit tests are skipped. Is it possible not to skip them?

I think the skipped ones are all because of SNOPT. Seems like we might be able to install SNOPT on these workflows.

PR #679, opened 2025-02-27
Added subsystems_options combability to propulsion builder

jkirk5:
### Summary

Adds `**kwargs` support to `build_mission()`, `build_premission()` so `subsystem_options` in `phase_info` can be passed down into individual engines

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/propulsion/propulsion_mission.py
+++ b/aviary/subsystems/propulsion/propulsion_mission.py
@@ -28,10 +28,19 @@ class PropulsionMission(om.Group):
             'engine_models', types=list, desc='list of EngineModels on aircraft'
         )
 
+        # engine options is optional
+        self.options.declare(
+            'engine_options',
+            types=dict,
+            default={},
+            desc='dictionary of options for each EngineModel'
+        )
+
     def setup(self):
         nn = self.options['num_nodes']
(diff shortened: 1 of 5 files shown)


aviary/subsystems/propulsion/propulsion_mission.py
```diff
@@ -114,19 +130,25 @@ def setup(self):
...
+                engine.name,
+                subsys=engine.build_mission(
+                    num_nodes=nn,
+                    aviary_inputs=options,
+                    **options
+                ),
+                promotes_inputs=['*'],
+            )
 
-            for i, engine in enumerate(engine_models):
```
Kenneth-T-Moore, line 118: Where did the loop go?

jkirk5: I found some redundant code that seemed to come from different updates. First `engine = engine_models[0]`, because this is a branch of the logic tree where `engine_models` is length 1. Then we loop through `engine_models`, overwriting the value of `engine` we know we will be using. So I streamlined it and removed the unecessary loop (we don't need index "i" in this section either

Kenneth-T-Moore (approved):

PR #686, opened 2025-03-03
Electric Breguet cruise EOM/ODE with unit testing

xjjiang:
### Summary

This PR creates Electric Breguet cruise EOM/ODE with unit testing.

A fake table engine is created for testing. A more realistic engine is needed for further testing.

### Related Issues

- Resolves #

None

### New Dependencies

None


jkirk5 (approved):

aviary/mission/gasp_based/ode/breguet_cruise_eom.py
```diff
@@ -5,6 +5,9 @@
 from aviary.variable_info.variables import Dynamic
 
 
+ft_to_nm = 6076.1
```
crecine, line 8: instead of using this constant, could we just update the units of TAS_cruise to be "Nm/s"?

xjjiang: This will change in both `RangeComp` and `E_RangeComp`. Should we do it? @jkirk5 @Kenneth-T-Moore

ehariton: I'm onboard with this change. You're doing all the other inputs in NM, might as well go for NM/S

aviary/mission/gasp_based/ode/test/test_data/turbofan_23k_electrified.deck
ehariton, line 1: I don't think this is the correct place for an engine deck. All the other models are in
`aviary/models/engines/`

ehariton (changes requested):

aviary/mission/gasp_based/ode/test/test_data/turbofan_23k_electrified.deck
xjjiang: This engine is totally fake and is for testing purpose only.

aviary/mission/gasp_based/ode/breguet_cruise_eom.py
xjjiang: Now, I've switched the units of `TAS_cruise` to `NM/s` from `ft/s`.

PR #690, opened 2025-03-07
Phase info revamp part 2: Use modified OptionsDictionary for type-checking the phase_info.

Kenneth-T-Moore:
### Summary

1. Defined a new `AviaryOptionsDictionary` from OpenMDAO's `OptionsDictionary`.
 - Variables with Units are stored as tuples `(val, unit_string)`
 - Bounds variables like `((lower, upper), unit_string)` are also supported.
 - New `units` argument when declaring an option.
 - Unit conversion on set handled in the class.
 - Method `get_val(name, units=new_unit` added to support conversion on get.

2. Modified all phase builders to store all supported phase_info options in the local user_options, which is an AviaryOptionsDictionary instead of a python dict.
 - In some cases, more restrictive typing was used, since it is supported.

TODO:
Follow-on work includes a complete overhaul of the phase_info keys to make them more uniform, more flexible, and more understandable.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/mission/gasp_based/phases/descent_phase.py
```diff
@@ -1,11 +1,243 @@
...
+
+        self.declare(
+            name='EAS_limit',
+            default=0.0,
+            units='kn',
+            desc='Value for maximum speed constraint in this phase.'
+        )
+
+        self.declare(
+            name='mach_cruise',
```
ehariton, line 69: I can't tell in the code if this is just a guess for mach_cruise. Since this is descent, it probably only uses this value for the first point of mach at t_0 for descent.

ehariton (approved):
Nice job Ken.

aviary/interface/test/test_phase_info.py
```diff
@@ -33,7 +33,7 @@ def _test_phase_info_dict(self, phase_info_dict, name):
         _phase_builder: PhaseBuilder = phase_info_to_builder(*_climb_info)
 
         # Convert back the phase builder to phase info
-        _phase_builder_info = _phase_builder.to_phase_info()
+        _phase_builder_info = _phase_builder.to_phase_info(legacy=True)
```
jkirk5, line 36: This seems to be the only place the legacy flag is used? If so then we should just rework the test rather than adding a workaround to the API to handle a test case.

aviary/utils/aviary_options_dict.py
```diff
@@ -0,0 +1,292 @@
...
+            return
+
+        # Loop over all user_options and set them.
+        for name, val in data.items():
+
+            # Support for legacy format (unitless)
+            if (isinstance(val, tuple) and
+                self._dict[name]['set_function'] is None and
+                    val[1] == "unitless"):
+                val = val[0]
```
jkirk5, line 146: Should we have a `DeprecationWarning` here? I assume we don't want to support the legacy format indefinitely

aviary/utils/aviary_options_dict.py
```diff
@@ -0,0 +1,292 @@
...
+        Returns
+        -------
+        dict
+            Equivalent phase_info.
+        """
+        phase_info = {}
+        for name, val in self.items():
+
+            meta = self._dict[name]
+            return_units = legacy or 'units' in meta
```
jkirk5, line 284: Similar `DeprecationWarning` here?

jkirk5 (approved):
A couple questions about future plans (the `legacy` flag) but otherwise good to go!

aviary/interface/test/test_phase_info.py
Kenneth-T-Moore: Yeah, that is a better approach. I have removed the legacy flag and just reworked the testing.

aviary/utils/aviary_options_dict.py
Kenneth-T-Moore: I was thinking about doing that in the second part of this, where we reimagine all of the phase_info keys.

aviary/mission/gasp_based/phases/descent_phase.py
Kenneth-T-Moore: It's an actual boundary constraint, but this is definitely one of those cases where we have to ask, do we really need both "mach_constraint" and "target_mach"? The latter is a boolean switch to activate the former.

Issue #706, opened 2025-03-18
Aerodynamics Code Re-Use


jkirk5:
### Desired capability or behavior.

The various aerodynamics methods have many steps in common that can likely be consolidated. The `aero_common.py` file is an example of a place this is happening, but it is very short (and the `DynamicPressure`) component should probably get moved to the Atmosphere subsystem).

A incomplete list of code that can likely be consolidated include:
1. Scaling of aero coefficients using all of the various scaling factors provided (and using this component for all methods)
2. Calculation of aircraft-level forces (i.e. total lift and drag) from coefficients
3. Tabular aero (probably not all of the code, but hopefully large portions of it)

jkirk5:
Combining all tabular aero methods will probably require updating our data table format to allow for specifying independent/dependent variables somehow

PR #708, opened 2025-03-20
Groundwork for aero builder overhauls

jkirk5:
### Summary

Adds the `AERODYNAMICS_METHOD` flag to `Settings`
Streamlines the geometry builder to remove the `both_geom` flag, and instead provide a tuple of (`FLOPS`, `GASP`)

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -126,6 +126,9 @@ class AviaryProblem(om.Problem):
         self.mission_method = mission_method = aviary_inputs.get_val(
             Settings.EQUATIONS_OF_MOTION)
         self.mass_method = mass_method = aviary_inputs.get_val(Settings.MASS_METHOD)
+        self.aero_method = aero_method = aviary_inputs.get_val(
+            Settings.AERODYNAMICS_METHOD
+        )
 
         # Create engine_builder
         self.engine_builders = engine_builders
@@ -149,7 +152,7 @@ class AviaryProblem(om.Problem):
                 f'settings:equations_of_motion must be one of: height_energy, 2DOF, solved_2DOF, or custom')
 
         # TODO this should be a preprocessor step if it is required here
-        if mass_method is GASP:
+        if mass_method is GASP or aero_method is GASP:
             aviary_inputs = update_GASP_options(aviary_inputs)
 
         ## LOAD PHASE_INFO ###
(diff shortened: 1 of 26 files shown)


aviary/interface/methods_for_level2.py
```diff
@@ -126,6 +126,9 @@ def load_inputs(
         self.mission_method = mission_method = aviary_inputs.get_val(
             Settings.EQUATIONS_OF_MOTION)
         self.mass_method = mass_method = aviary_inputs.get_val(Settings.MASS_METHOD)
+        self.aero_method = aero_method = aviary_inputs.get_val(
```
Kenneth-T-Moore, line 129: So presently, the user chooses the "mass method" and the "equations of motion". When we started to allow the gasp-flops crossover versions, we determined that the aero method was coupled to the equations_of_motion. But now that we consider it a builder-style subsystem that just provides the lift and drag forces, cutting the connection is feasible. It does add some additional complexity for the logic to decide what geometry we need to include. That decision is currently driven by EoM, but really, the aero_method is what it should be.

I think there is a further consideration on what to do with the solved aero method in FLOPS, which is really a gasp-style polar, but without the empirical part (and thus, wouldn't need any gasp geometry.)

Finally, when a user provides a custom aero, what value should AERODYNAMICS_METHOD take?

jkirk5: Part 2 will address the solved aero method and moving it to the GASP side.

For custom aero, you have to select the aero_method that matches the table style you are providing. If your table has AoA, you will need to select GASP, and otherwise you want FLOPS. Eventually I'd like to combine all tabular aero into a single method (#706), which should be possible if we update our table format to specify independent/dependent variables. Then tabular aero is independent of code origin, and might become optional in that case.

Kenneth-T-Moore: So, for current MBSA&E users who switch between external aero or a gasp-style table that doesn't require gasp geometry, they should use Flops for the aerodynamics method? (and also, what is the expected behavior if they don't specify an aero method?)

Kenneth-T-Moore (approved):
My main comments are currently on part 2 of the aero overhaul, so I think this one can go in.

aviary/interface/methods_for_level2.py
jkirk5: Current implementation in the follow-on is that if you want to use a GASP-style table, you use the GASP aero method, and similar for FLOPS. And if you want to use GASP aero with energy method you'll need a `solve_aero` flag (maybe we could automate that though). If you don't provide an aero method, you get an exception. If you're using external aero and don't want Aviary aero to be used, then you should be using the existing 'external' flag.

PR #747, opened 2025-04-22
Engine deck conversion update

jkirk5:
### Summary

Testing of engine conversion was broken, never actually checking data. Converter utility as it currently existed produced different results than what was actually present (different number of generated points).

- Reverted some converter settings to their original values, which can reproduce our original Aviary engine decks (number of generated points in those decks was intended behavior)
- Added in optional rounding of engine deck values when converting. Engine decks inside models were either re-converted if old data still available, or just re-formatted & rounded without otherwise changing data for better human readability.
- Updated tests that changed with these new rounded engine decks. Typically changes are extremely minor and only break very tight tolerance tests

### Related Issues

- Resolves #

Will very slightly change results for any model that uses the built-in engine decks. This PR may want to wait for 0.10.1+

### New Dependencies

None


aviary/utils/engine_deck_conversion.py
```diff
@@ -58,22 +59,52 @@ def __str__(self):
...
+    GROSS_THRUST: 7,
+    RAM_DRAG: 6,
+    FUEL_FLOW: 6,
+    NOX_RATE: 5,
+    TEMPERATURE: 6,
+    SHAFT_POWER_CORRECTED: 6,
+    TAILPIPE_THRUST: 6,
+}
+
+# number of sig figs to round each header to, if requested
```
Kenneth-T-Moore, line 91: This code is duplicated here.

jkirk5: fixed!

PR #748, opened 2025-04-23
Fixed a couple of regressions on the nightly build.

Kenneth-T-Moore:
### Summary

This fixes 2 failures from the nightly build:

1. OpenMDAO error message for `get_val` has recently changed. Make test work with both.
2. The 2dof aero was missing some promotes that formerly had non-zero default values.  This fixes the broken GwGm test and bench, which now optimizes to a nice 0,1 in 21 iters.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/test/test_premission.py
+++ b/aviary/subsystems/test/test_premission.py
@@ -214,7 +214,9 @@ class PreMissionTestCase(unittest.TestCase):
         )
         with self.assertRaises(KeyError) as cm:
             self.prob.get_val(Aircraft.Fuel.AUXILIARY_FUEL_CAPACITY)
-        self.assertTrue(err_text1 in str(cm.exception) or err_text2 in str(cm.exception))
+
+        actual_text = str(cm.exception)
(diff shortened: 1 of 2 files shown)


aviary/subsystems/test/test_premission.py
```diff
@@ -205,12 +205,15 @@ def test_GASP_mass_FLOPS_everything_else(self):
 
         # This is not in the model because it has been overridden, but is not an
         # input to any other component in the GASP premission model.
-        err_text = (
-            '\'<model> <class Group>: Variable "aircraft:fuel:auxiliary_fuel_capacity" not found.\''
-        )
+        # Note: newer version of OpenMDAO has changed the exception text
```
jkirk5, line 208: This section is super odd to me. We're testing that an exact string is printed when a variable is not found? That seems to be testing the OM interface for a specific behavior (printout) rather than an Aviary behavior (that some edge case isn't being violated where a variable is unintendedly added to the problem?)

jkirk5 (approved):

PR #761, opened 2025-05-09
Fix bug that broke preprocessing of crew options.

### LINKED ISSUE #738: Default values have broken the pre-processor for FLOPS crew/galley processing.
### Description

The section of code below from `preprocessors.py` is meant to compute some values, mostly for flight/galley crew numbers, when the user does not submit them. In FLOPS, this was controlled by setting the values to -1.  The Aviary analog is to not set them in the csv file.  Unfortunately, recent changes have given these variables a default value, and all options that don't have a value in the csv are set to default values (which are 0 for all of them). This breaks the code below. We need some adjustment to the logic to allow these "derived defaults" to be set.

    if Aircraft.CrewPayload.NUM_FLIGHT_ATTENDANTS not in aviary_options:
        flight_attendants_count = 0  # assume no passengers

        if 0 < passenger_count:
            if passenger_count < 51:
                flight_attendants_count = 1

            else:
                flight_attendants_count = passenger_count // 40 + 1

        aviary_options.set_val(
            Aircraft.CrewPayload.NUM_FLIGHT_ATTENDANTS, flight_attendants_count
        )

    if Aircraft.CrewPayload.NUM_GALLEY_CREW not in aviary_options:
        galley_crew_count = 0  # assume no passengers

        if 150 < passenger_count:
            galley_crew_count = passenger_count // 250 + 1

        aviary_options.set_val(Aircraft.CrewPayload.NUM_GALLEY_CREW, galley_crew_count)

    if Aircraft.CrewPayload.NUM_FLIGHT_CREW not in aviary_options:
        flight_crew_count = 3

        if passenger_count < 151:
            flight_crew_count = 2

        aviary_options.set_val(Aircraft.CrewPayload.NUM_FLIGHT_CREW, flight_crew_count)

    if (
        Aircraft.CrewPayload.BAGGAGE_MASS_PER_PASSENGER not in aviary_options
        and Mission.Design.RANGE in aviary_options
    ):
        design_range = aviary_options.get_val(Mission.Design.RANGE, 'nmi')

        if design_range <= 900.0:
            baggage_mass_per_pax = 35.0
        elif design_range <= 2900.0:
            baggage_mass_per_pax = 40.0
        else:
            baggage_mass_per_pax = 44.0

        aviary_options.set_val(
            Aircraft.CrewPayload.BAGGAGE_MASS_PER_PASSENGER,
            val=baggage_mass_per_pax,
            units='lbm',
        )

### Example

See the commented section in the N3CC csv file that Nate is submitting soon.

### Aviary Version

0.9.10-dev

_No response_

Kenneth-T-Moore:
### Summary

Certain keys need to remain empty so that the preprocessor knows when to set them.
Bonus: Fix for failing IPOPT bench.

### Related Issues

- Resolves #738

None

### New Dependencies

None


aviary/utils/process_input_decks.py
```diff
@@ -71,6 +72,7 @@ def create_vehicle(vehicle_deck='', meta_data=_MetaData, verbosity=Verbosity.BRI
         verbosity = Verbosity.BRIEF
 
     aircraft_values = get_option_defaults(engine=False)
```
jkirk5, line 74: I'm not really sure why we're loading all the defaults to begin with, it doesn't seem like this is very helpful anymore. Maybe we should consider skipping this step in the future, which should remove the need for `remove_preprocessed_options`

jkirk5 (approved):

aviary/utils/process_input_decks.py
Kenneth-T-Moore: That is a good point. I am not sure why we pre-load those into the aviary_options. All of the defaults are also present in the options dictionaries, so they get set correctly even without this step.  Maybe there is something on the GASP side that isn't doing that.

aviary/models/N3CC/N3CC_FLOPS.csv
```diff
@@ -19,14 +19,12 @@ aircraft:crew_and_payload:flight_crew_mass_scaler,1,unitless
 aircraft:crew_and_payload:mass_per_passenger,165,lbm
 aircraft:crew_and_payload:misc_cargo,0,lbm
 aircraft:crew_and_payload:non_flight_crew_mass_scaler,1,unitless
+aircraft:crew_and_payload:num_business_class,20,unitless
```
ehariton, line 22: You shouldn't have to add these passenger values because they should be automatically setup based on the design values. I think you can remove these 4 lines of changes.

Kenneth-T-Moore: I added these because they weren't getting set right, which must mean that a new bug has cropped up.

PR #777, opened 2025-05-23
Phase info revamp: part 1: Height Energy

Kenneth-T-Moore:
### Summary

This is a major overhaul of the phase_info dictionaries with the following goals:
    1. Provide a set of keys that is more consistent across the different phase types.
    2. Remove redundancies, particularly cases where we can replace 2 keys with 1.
    3. Rename keys to be more understandable to the user.
    4. Add missing options such as certain missing scaling parameters.

Before Merging:
    1. Evaluate the new phase_info keys and make changes.
    2. Finish modifying the documentation, and create a transition document.
    3. Add a few tests that use some of the new keys.

Summary of Changes:
    1. Created AviaryOptionsDictionary methods to add all options associated with a state, control, or time. Most phase builders now use these, but some of the GASP extra states/controls don't use them yet . 
      - This provides a uniform interface for states or controls.
      - Missing keys were added, such as bounds and ref/ref0 for states. Formerly, these were hardcoded values.
    2. Created PhaseBuilderBase methods to add a state or control to a phase. FlightPhaseBuilder now uses these.
    3. Many options renamed to make them more uniform and consistent. This includes some options in 2DOF, because a global search and place was convenient. (e.g, "initial_mach" -> "mach_initial")
    4. Some methods were renamed to make them clearer.
    5. Some logic was simplified in the configurators and in methods for level 2.
    6. The phase_info graphical tool was modified to output the phase_info in a pretty format.
    7. Some checks for an old version of dymos were removed, since we are reliant on a more recent version.

Follow-on work:
    1. Apply changes to 2DOF
    2. Modify MBSAE to use new phase-info
    3. Get detailed takeoff working better.

### Related Issues

- Resolves #

All user phase-info files will need to be modified.

### New Dependencies

None

### DIFF
--- a/aviary/docs/user_guide/phase_info_detailed.ipynb
+++ b/aviary/docs/user_guide/phase_info_detailed.ipynb
@@ -11,7 +11,7 @@
    "outputs": [],
    "source": [
     "# Testing Cell\n",
-    "import openmdao.api as om\n"
+    "import openmdao.api as om"
    ]
   },
   {


aviary/docs/examples/additional_flight_phases.ipynb
```diff
@@ -52,168 +52,147 @@
...
+    "            'altitude_initial': (33000.0, 'ft'),\n",
+    "            'altitude_final': (34500.0, 'ft'),\n",
     "            'altitude_bounds': ((32500.0, 35000.0), 'ft'),\n",
     "            'throttle_enforcement': 'boundary_constraint',\n",
-    "            'fix_initial': False,\n",
-    "            'constrain_final': False,\n",
-    "            'fix_duration': False,\n",
-    "            'initial_bounds': ((78.0, 234.0), 'min'),\n",
-    "            'duration_bounds': ((7.0, 21.0), 'min'),\n",
+    "            'time_initial_bounds': ((78.0, 234.0), 'min'),\n",
```
ehariton, line 152: this is a really nice change so we can easily tell that initial and duration bounds refers to time.

aviary/docs/examples/reserve_missions.ipynb
```diff
@@ -44,7 +44,7 @@
     "expected_values = {\n",
     "    'reserve': True,\n",
     "    'target_distance': (200, 'km'),\n",
```
ehariton, line 46: considering we have 'distance_solve_segments' wouldn't it make sense to update 'target_distance' to 'distance_target' ?

aviary/docs/examples/reserve_missions.ipynb
```diff
@@ -98,7 +97,6 @@
     "\n",
     "from aviary.examples.reserve_missions.run_reserve_mission_fixedrange import phase_info\n",
     "\n",
-    "check_phase_info(phase_info, HEIGHT_ENERGY)\n",
```
ehariton, line 101: I guess we don't need `check_phase_info` at all anymore because if it doesn't match with the library definition it's rejected?

aviary/interface/default_phase_info/height_energy.py
```diff
@@ -113,17 +106,75 @@ def phase_info_parameterization(phase_info, post_mission_info, aviary_inputs):
...
     if alt_cruise != old_alt_cruise:
-        phase_info['climb']['user_options']['final_altitude'] = (alt_cruise, 'ft')
-        phase_info['cruise']['user_options']['initial_altitude'] = (alt_cruise, 'ft')
-        phase_info['cruise']['user_options']['final_altitude'] = (alt_cruise, 'ft')
-        phase_info['descent']['user_options']['initial_altitude'] = (alt_cruise, 'ft')
+        new_alt = (alt_cruise, 'ft')
+
+        climb = phase_info['climb']
+        user = climb['user_options']
+        if 'altitude_final' in user and user['altitude_final'][0] is not None:
```
ehariton, line 113: Is this going to work ok with the current scheme of loading defaults which might overwrite the 'None' flag.

aviary/mission/two_dof_problem_configurator.py
```diff
@@ -350,7 +352,7 @@ def set_phase_options(self, prob, phase_name, phase_idx, phase, user_options):
             )
 
         elif 'descent' in phase_name:
-            duration_ref = user_options.get_val('duration_ref', 's')
+            duration_ref = user_options['time_duration_ref'][0]
```
ehariton, line 355: are we going to be ok with unit conversions here? This is hard-coded to seconds but what if the user inputs minutes?

aviary/utils/aviary_options_dict.py
```diff
@@ -149,3 +149,295 @@ def to_phase_info(self):
...
+            types=bool,
+            desc=desc,
+        )
+
+    def add_control_options(self, ctrl_name: str, units: str = None, defaults=None):
+        """
+        Adds all options needed for a control variable.
+
+        For a control named mach, these are mach_initial, mach_final, mach_bounds, mach_ref,
+        mach_ref0, mach_defect_ref, mach_solve_segments.
```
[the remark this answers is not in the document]
ehariton: `mach_defect_ref` isn't in here??

aviary/docs/examples/reserve_missions.ipynb
Kenneth-T-Moore: This is actually an oddball one because the state is called `distance`, but target_distance is really more like target_delta_distance. If we wanted to generalize this to other states like mass, `mass_target` wouldn't imply end_mass - start_mass.  I think we could come up with better name though.

aviary/docs/examples/reserve_missions.ipynb
Kenneth-T-Moore: We don't need it to run in `check_and_preprocess_inputs` anymore, because the args are checked against the phase builders options dictionary. There may be a secondary use for it, just for making sure the doc phase_infos are correct though.

aviary/docs/user_guide/examples_of_the_same_mission_at_different_UI_levels.ipynb
```diff
@@ -344,12 +329,13 @@
     "    'test_cruise',\n",
     "    user_options=av.AviaryValues(\n",
     "        {\n",
-    "            'initial_altitude': (alt_min_cruise, 'm'),\n",
-    "            'final_altitude': (alt_max_cruise, 'm'),\n",
-    "            'initial_mach': (cruise_mach, 'unitless'),\n",
-    "            'final_mach': (cruise_mach, 'unitless'),\n",
+    "            'altitude_polynomial_order': (3, 'unitless'),\n",
```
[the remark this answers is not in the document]
Kenneth-T-Moore: Good catch. It was a cut-and-paste error. It still works with those, but they aren't required.

EDIT: Actually, this example is level3, so I need to also get rid of the step where we pack them all in an AviaryValues.

aviary/interface/default_phase_info/height_energy.py
Kenneth-T-Moore: It should, since the default actually is None.

aviary/utils/aviary_options_dict.py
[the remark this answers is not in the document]
Kenneth-T-Moore: Good catch. Fixing this.

aviary/utils/test/test_aviary_options_dict.py
```diff
@@ -67,6 +67,59 @@ def test_errors(self):
...
+
+        opts.add_control_options('zzz')
+
+        stems = [
+            'initial',
+            'final',
+            'bounds',
+            'ref',
+            'ref0',
+            'polynomial_order',
```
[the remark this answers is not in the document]
Kenneth-T-Moore: These should be correct now. The docstring was a copy-paste error.

aviary/docs/examples/reserve_missions.ipynb
ehariton: I thought distance was always relative to the start of the phase your in vs. range which is relative to the absolute start of the mission.

PR #782, opened 2025-05-30
Overhaul of propeller map code

### LINKED ISSUE #729: propeller map implementation
### Description

when definining a turboprop engine with an external propeller map provided by the user like this : 
`shaft_model=EngineDeck(name='shaft',options=options,required_variables=required_vars)`
`propeller=PropellerMap(name='Centauri_propeller',options=options)`
`turboprop=TurbopropModel('turboprop',options=options,shaft_power_model=shaft_model,propeller_model=propeller)`

`prob.load_inputs(r'aircraft_definition.csv',phase_info,engine_builders=[turboprop],verbosity=3)`

this leads to an error in the turboprop_model.py on line 191 

`if self.propeller_model is not None:`
`    params.update(self.propeller_model.get_parameters())`
`return params`

### Example

since the PropellerMap function does not currently have attributes of get_parameters,build_pre_mission,build_mission,get_mass ... all of which are variables of propeller_builder . when adding all the attributes to propellermap class function , the leads to circular imports error due to `propeller_map` importing `propeller_performance` for the attribute `build_mission`  and `propeller_performance` imports `propeller_map`

### Aviary Version

0.9.8

_No response_

jkirk5:
### Summary

Refactors the propeller map code to fix a few key issues and streamline the code

- PropellerMap group now actually gets added to the problem - before it was being used a subsystem builder-like tool
- Mach types were simplified, PropellerMap now directly looks at the data file headers to determine which mach is being used, and promotes those as inputs directly instead of using selectedMach as an aliased proxy. OutMachs component is no longer needed.
- Propeller data file conversion was updated to order the independent variables such that they are already sorted in the way the metamodel components want them to be. Re-converted the existing propeller data files to be in this new order, and with better formatting
- Updated the propeller map test to actually test the interpolation components used inside the group

### Related Issues

- Resolves #729

You still add propeller maps to the problem the same way, using `Aircraft.Engine.Propeller.DATA_FILE`, so no api-side changes here

### New Dependencies

None


aviary/docs/user_guide/hamilton_standard.ipynb
```diff
@@ -361,14 +360,10 @@
     "file_path = av.get_path(folder_path / map_file_name)\n",
     "glue_variable(file_path.name, md_code=True)\n",
     "\n",
-    "glue_variable(OutMachType.HELICAL_MACH.value, md_code=True)\n",
```
Kenneth-T-Moore, line 364: I guess I should have looked at this first. The removal of these 2 glue vars (helical mach, mach) seem to be causing some of the doc failures.

aviary/utils/test/test_propeller_map_conversion.py
```diff
@@ -4,40 +4,31 @@
...
         # Specify the legacy code and propeller map
-        args.data_format = data_format
+        data_format = data_format
 
         # Execute the conversion
-        _exec_PMC(args, None)
+        convert_propeller_map(input_file, output_file, round_data=True)
 
-    def compare_files(self, filepath, skip_list=[]):
+    def compare_files(self, filepath, skip_list=['# created']):
```
Kenneth-T-Moore: Generally not good practice to use lists or dictionaries as default arguments: https://www.pullrequest.com/blog/python-pitfalls-the-perils-of-using-lists-and-dicts-as-default-arguments/

jkirk5: Oh wow that's pretty bad. Very inconvenient, I bet we're using mutable args as defaults in a bunch of places in the actual code

Kenneth-T-Moore: Best fix is to make the default None, then put in code like:

    if skip_list is None:
        skip_list=['# created']

aviary/docs/user_guide/hamilton_standard.ipynb
jkirk5: I'm not sure exactly how, since OutMachType doesn't exist anymore and I can't find any other references to the glued `helical_mach`. Seems like a large cascade of doc failures beyond just that variable

PR #787, opened 2025-06-03
Adding unit tests for GASP based premission

xjjiang:
### Summary

This is part of ongoing BWB task and so it is optional.

We have unit test for FLOPS based unit tests but never have one for GASP based. Two unit tests are added using `models/test_aircraft/aircraft_for_bench_GwGm.csv`: one for default premission: propulsion + geometry + aerodynamics + mass and the other for geometry + mass. When propulsion and aerodynamics are not in, the following two variables must be defined:

- `Aircraft.Engine.SCALED_SLS_THRUST`
- `Mission.Landing.LIFT_COEFFICIENT_MAX`

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/utils/preprocessors.py
```diff
@@ -43,7 +43,8 @@ def preprocess_options(aviary_options: AviaryValues, meta_data=_MetaData, verbos
             aviary_options.set_val(Settings.VERBOSITY, verbosity)
 
     preprocess_crewpayload(aviary_options, meta_data, verbosity)
-    preprocess_propulsion(aviary_options, engine_models, meta_data, verbosity)
+    if not engine_models is None:
```
jkirk5:
```suggestion
    if engine_models is not None:
```

PR #790, opened 2025-06-06
Update Aviary csv format to include input/output information in header

jkirk5:
### Summary

Adds additional optional information in the Aviary csv data file headers, `input` and `output`. These new flags are added inside parentheses along with units (e.g. `Altitude (ft, input)`).

This achieves two important aims: one, it makes data tables self-explanatory without prior knowledge about them. Currently it is hardcoded which variables are inputs and outputs deep inside Aviary code, and this is never obvious to the user. Two, this will allow for fully customizable tables for many subsystems such as aero. This also allows for much easier combining of FLOPS and GASP tabular aero into a completely generic "Aviary tabular aero" component.

### Related Issues

- Resolves #

None, the old data tables w/o input/output info are still compatible, just less verbose and rely on the hardcoded input/output lists (which now only exist as backups if the table does not provide that info)

### New Dependencies

None


aviary/docs/user_guide/hamilton_standard.ipynb
```diff
@@ -361,14 +360,10 @@
     "file_path = av.get_path(folder_path / map_file_name)\n",
     "glue_variable(file_path.name, md_code=True)\n",
     "\n",
-    "glue_variable(OutMachType.HELICAL_MACH.value, md_code=True)\n",
```
Kenneth-T-Moore: Some of the doc failures come from removal of these 2 glue variables. (either they need to be added back here, or removed from the doc text below.

Kenneth-T-Moore:
Let me know if there are any difficulties in merging the phase_info changes into here.

aviary/docs/user_guide/hamilton_standard.ipynb
jkirk5: This PR is dependent on #782, so once that gets merged in hopefully that resolves this doc issue

PR #792, opened 2025-06-10
Jax support for helper functions

jkirk5:
### Summary

Adds `primal_name` creation to helper functions `add_aviary_input`/`add_aviary_output`. Colons are replaced with double underscores to produce a primal name that is a valid python variable while retaining readable hierarchy separation.

Example: `aircraft:wing:aspect_ratio` gets the primal name `aircraft__wing__aspect_ratio`

### Related Issues

- Resolves #

Minimum supported OpenMDAO version is now 3.37

### New Dependencies

None


xjjiang:
Maybe add a simple unit test?

PR #798, opened 2025-06-11
Fixes a bug where GASP horizontal and vertical tail areas couldn't be overriden.

### LINKED ISSUE #796: Overriding GASP tail geometry variables does not work.
### Description

Looks like an easy fix, just need to remove the extra aliasing.

### Example

MBSAE case from Tim.

### Aviary Version

0.10.0-dev

_No response_

Kenneth-T-Moore:
### Summary

Fixes a bug where GASP horizontal and vertical tail areas couldn't be overriden.
- Removed hard-coded "0" tail areas from the GASP csv files. These weren't caught because the override was broken.
- This also fixes where these areas are being set to zero in flops_to_aviary.
- The components in emmpenage are slightly less "general" now, but they conform to Aviary standards common to the other components.

### Related Issues

- Resolves #796

None

### New Dependencies

None


jkirk5 (approved):

aviary/models/test_aircraft/aircraft_for_bench_FwGm.csv
```diff
@@ -62,7 +62,6 @@ aircraft:fuselage:seat_pitch,29,inch
 aircraft:fuselage:seat_width,20.2,inch
 aircraft:fuselage:tail_fineness,3,unitless
 aircraft:fuselage:wetted_area,4000,ft**2
-aircraft:horizontal_tail:area,355.0,ft**2
```
xjjiang, line 65: Why is it removed even when it is not zero?

aviary/subsystems/geometry/gasp_based/empennage.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import openmdao.api as om
 
-from aviary.variable_info.functions import add_aviary_input, add_aviary_option
+from aviary.variable_info.functions import add_aviary_input, add_aviary_output, add_aviary_option
```
xjjiang, line 4: Should we list them in alphabetic order?

aviary/subsystems/geometry/gasp_based/empennage.py
```diff
@@ -37,39 +45,36 @@ def setup(self):
...
-        self.add_input(
-            'wing_ref',
-            12.615,
-            units='ft',
-            desc=(
-                'CBARW | B: Wing reference parameter. Wing chord for a '
-                'horizontal tail. Wing span for a vertical tail.'
-            ),
-        )
+        add_aviary_input(self, self.io_names['wing_ref'], val=12.612, units='ft')
```
xjjiang, line 52: I think it will be good to keep the description so that some history is not lost. Same comment applies to other variables.

aviary/models/test_aircraft/aircraft_for_bench_FwGm.csv
Kenneth-T-Moore: Hey, Xun. I removed this because the problem is using the gasp mission, and seems to work better with the gasp-computed tail value instead of the one from the FLOPS input file. Note that, due to the bug, override wasn't working for the GASP empennage, so the 355 value was being ignored.

aviary/subsystems/geometry/gasp_based/empennage.py
Kenneth-T-Moore: Yeah, i will fix that on whatever PR i submit next.

aviary/subsystems/geometry/gasp_based/empennage.py
Kenneth-T-Moore: I guess this is a good point that there is some value to keeping some of the descriptive text from the generic "sizing" component.

PR #799, opened 2025-06-13
Fixes FLOPS alternate missions resizing aircraft in off design 

### LINKED ISSUE #771: Design mass changes for off design FLOPS missions
### Description

When running Alternate or fallout missions the masses of avionics and air conditioning systems changes, which is incorrect. The design fuel mass is also incorrectly recalculated.

### Example

Off design example running an alternate mission with increased payload and reduced range shows incorrect changes in subsystem masses.
Fallout mission with reduced payload shows incorrect recalculation of design fuel mass.

### Aviary Version

0.10.0-dev

_No response_

aalemu1:
### Summary

Aviary's current "off_design_example.py" fails to converge for both sizing and off-design missions. To address this and reduce the total amount of phase information within the repository, a slight modification of the default height energy phase information was created to allow for off-design missions. 

Aviary's alternate off-design missions would attempt to set the mission_range parameter as the Mission.Design.RANGE, therefore the aircraft would have certain components that relied on Mission.Design.RANGE resized within an alternate mission. That resulted in the alternate empty mass being inconsistent with a sizing mission. The alternate mission would also attempt to run the Mission.Design.RANGE instead of the mission_range, resulting in non-convergence in alternate missions with no difference in parameter from the sizing mission. 

This fix properly distinguishes mission_range (Alternate), Mission.Design.RANGE (CSV), and "Target_Range" (Phase Info), within off-design missions to ensure reliable and consistent parameter usage across mission types. 

### Related Issues

- Resolves #771 

None

### New Dependencies

None


Kenneth-T-Moore:
I don't have a conversion guide yet, but I can help you convert your phase_info to the new format, if you would like.

PR #800, opened 2025-06-13
Cleaned up Examples

jkirk5:
### Summary
Simplified and streamlined the example problems available with Aviary. The goal is to have each example provide a complete (and working) problem that demonstrates a single feature. There should be as few examples as possible to teach those features.

This PR is mostly targeted at the actual python scripts in the repo, and docs changes will come as a follow-on

Removed Files
- Redundant examples such as separate scripts for reserve missions with different mission methods (they don't teach the user anything new about specifying a reserve mission)
- Unnecessary models that don't have a clear purpose (such as a copy of an engine deck that just changes fuel flow units)
- Duplicate files (such as FLOPS 28k engine deck, which had 2 copies in repo)

Renames
- Example files are renamed to be clearer what feature they explain, as well as contain the word "example" in them
- Tabular data files are given `.csv` file extensions instead of arbitrary extensions such as `.deck`, `.prop`, `.map`, etc. so they can be recognized and opened by other programs
- Renamed N3CC to "advanced_single_aisle"

Reorganization
- Moved some models that are only for testing purposes into models/test_aircraft along with a README to explain what they are and their purpose
- Separated out models into "aircraft", "engines", and "missions" subfolders. Moved default phase_info to "missions" instead of "interface"

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/api.py
```diff
@@ -48,15 +48,15 @@
     SpeedType,
     Verbosity,
 )
-from aviary.interface.default_phase_info.two_dof import phase_info as default_2DOF_phase_info
-from aviary.interface.default_phase_info.two_dof_fiti import (
+from aviary.models.missions.default_phase_info.two_dof import phase_info as default_2DOF_phase_info
```
Kenneth-T-Moore: the `missions` directory only contains a `phase_info` directory, so I think we could move all the files up. Maybe just make  the dir "mission_phase_info".

aviary/examples/external_subsystems/OAS_weight/simple_OAS_mission_example.py
```diff
@@ -0,0 +1,208 @@
+"""
+This is a simple test mission to demonstrate the inclusion of a pre-mission user defined external
+subsystem. The simple mission is based on input data read from the benchmark data file bench_4.csv,
```
Kenneth-T-Moore: Note, this example won't be tested anymore after the rename, because the example tester looks for files that start with `run_`.

aviary/api.py
jkirk5: I assume at some point we might have other missions besides the defaults, but I could just change the filenames to have the word "default" in them to avoid confusion without adding another directory

aviary/examples/external_subsystems/OAS_weight/simple_OAS_mission_example.py
jkirk5: Oh no that file somehow got resurrected as part of the git merge. It's an exact copy of `run_simple_OAS_mission.py` so I will remove this one

aviary/api.py
jkirk5: Made that suggested change & added "default" to file names

PR #803, opened 2025-06-26
Aviary group rework 1

ehariton:
### Summary

Forwarded a number of functions from methods_for_level2 into AviaryGroup without changing top-level user interface. Pushing these functions down one level into prob.model will enable the first step to re-modelling core aviary to be more flexible to multi-mission analysis. Some methods were moved to utilities, and some small temporary changes were added to report generation to prevent that from breaking. Objective function declaration was removed from all problem_configurator files because the ODE method does not necessarily correspond with a specific objective function. For methods that remain in methods_for_level2 (alternative / fallout missions / objectives), some of them were modified to use prob.model instead of prob.

Some secondary effects of these changes. There were 'fake' aviary problems (essentially just OpenMDAO groups) setup in onboarding_for_level2.ipynb, one of which was maintained, but the other proved too convoluted to rehab to the new standard.

### Related Issues

N/A

None

### New Dependencies

None


aviary/examples/level2_example.py
```diff
@@ -22,15 +22,18 @@
 # Preprocess inputs
 prob.check_and_preprocess_inputs()
 
-prob.add_pre_mission_systems()
+# make all 4 of these the same thing
```
Kenneth-T-Moore: Probably should remove these design notes from the example.

aviary/interface/methods_for_level2.py
```diff
@@ -135,130 +108,28 @@ def load_inputs(
...
-        # self.engine_builders, self.pre_mission_info, self_post_mission_info
-        # self.require_range_residual, self.target_range
-        # Other specific self.*** are defined in here as well that are specific to
-        # each builder
+        self.verbosity = verbosity
 
         return self.aviary_inputs
 
-    def check_and_preprocess_inputs(self, verbosity=None):
+    def check_and_preprocess_inputs(self, verbosity=None): # forward
```
Kenneth-T-Moore: remove all these comments too.

aviary/interface/methods_for_level2.py
```diff
@@ -1199,6 +328,9 @@ def add_driver(self, optimizer=None, use_coloring=None, max_iter=50, verbosity=N
         driver.options['optimizer'] = optimizer
         if use_coloring:
             # define coloring options by verbosity
+            print("verbosity", verbosity)
```
Kenneth-T-Moore: Remove these prints.

aviary/core/AviaryGroup.py
```diff
@@ -15,6 +56,16 @@ class AviaryGroup(om.Group):
     that are present in the model.
```
jkirk5: This docstring is pretty outdated at this point and should get an update

aviary/core/AviaryGroup.py
```diff
@@ -115,3 +166,1435 @@ def configure(self):
...
+        # override was not requested
+        if self.verbosity is None:
+            self.verbosity = aviary_inputs.get_val(Settings.VERBOSITY)
+            # set default warning format for the rest of the problem
+            set_warning_format(self.verbosity)
+
+        # If user did not ask for verbosity override for this method either, use the problem's
+        # default verbosity for the rest of the method
+        if verbosity is None:
+            verbosity = aviary_inputs.get_val(Settings.VERBOSITY)
```
jkirk5:
```suggestion
```
This line doesn't exist in the original load_inputs

aviary/mission/height_energy_problem_configurator.py
```diff
@@ -23,59 +23,61 @@ class HeightEnergyProblemConfigurator(ProblemConfiguratorBase):
...
         Set any initial guesses for variables in the aviary problem.
 
         This is called at the end of AivaryProblem.load_inputs.
 
         Parameters
         ----------
-        prob : AviaryProblem
-            Problem that owns this builder.
+        aviary_group : AviaryGroup
+            Aviary model that owns this builder.
```
jkirk5:
```suggestion
            Aviary model that owns this configurator.
```

aviary/mission/problem_configurator.py
```diff
@@ -131,18 +133,60 @@ def link_phases(self, prob, phases, connect_directly=True):
...
         ----------
-        prob : AviaryProblem
-            Problem that owns this builder.
+        aviary_group : AviaryGroup
+            Aviary model that owns this builder.
         """
         pass
 
-    def add_post_mission_systems(self, prob):
+    def link_phases_helper_with_options(self, aviary_group, phases, option_name, var, **kwargs):
```
jkirk5, line 147: Why is this no longer a private method? It's never used outside of the problem_configurator as far as I can tell

aviary/utils/utils.py
```diff
@@ -256,3 +256,65 @@ def cast_type(key, val, meta_data=_MetaData):
                     break
 
     return cast_val
+
+
+def get_phase_mission_bus_lengths(traj):
```
jkirk5: This function isn't used Aviary wide, so I'd suggest moving it to a mission/utils.py

aviary/utils/utils.py
```diff
@@ -256,3 +256,65 @@ def cast_type(key, val, meta_data=_MetaData):
...
+    phase_mission_bus_lengths = {}
+    for phase_name, phase in traj._phases.items():
+        phase_mission_bus_lengths[phase_name] = phase._timeseries['mission_bus_variables'][
+            'transcription'
+        ].grid_data.subset_num_nodes['all']
+    return phase_mission_bus_lengths
+
+
+def process_guess_var(val, key, phase):
```
jkirk5: Same for this function moving to mission/utils.py

aviary/utils/functions.py
```diff
@@ -303,6 +304,40 @@ def compute(self, inputs, outputs):
     return PrintComp
 
 
+def set_warning_format(verbosity):
```
jkirk5: Personally I don't think this belongs in a utils file - it is never used anywhere outside methods_for_level2.py so moving it to a separate file seems overkill.

jkirk5 (changes requested):

aviary/mission/problem_configurator.py
ehariton: link_phases_helper_with_options is used in all 3 problem configurators. These all inherit from problem_configurator. Originally this method was kept in M4L2, but it didn't make much sense to keep it there.

aviary/mission/height_energy_problem_configurator.py
ehariton: I changed all instances of this in the HE, 2dof, and solved 2dof problem config files

aviary/api.py
```diff
@@ -96,7 +96,7 @@
 ###################
 
 # Miscellaneous
-from aviary.interface.methods_for_level2 import PreMissionGroup, PostMissionGroup
+# from aviary.interface.methods_for_level2 import PreMissionGroup, PostMissionGroup
```
Kenneth-T-Moore: Since this is commented, I think you can just delete it. Anyone who needs these is doing pretty tightly integrated work, and would be safer pulling these from the source import instead of api.py anyway.

Kenneth-T-Moore (approved):

aviary/core/AviaryGroup.py
ehariton: I gave it an update. Let me know if this works for you:
_A standard OpenMDAO group where all elements of a given aviary aircraft design and mission are defined. This includes pre_mission, mission, and post_mission analysis. This group also contains methods for loading data from .csv and phase_info files, setting initial values on the group, and connecting all the phases inside the mission analysis to each other. Instantiating multiple AviaryGroups allows for analysis and optimization of multiple aircraft or one aircraft in multiple missions simultaneously._

aviary/utils/functions.py
Kenneth-T-Moore: It's technically in AviaryGroup too, which would create a circular import.

ehariton: TODO: will move to interface/utils/markdown_utils.py and rename to utils.py

aviary/utils/utils.py
ehariton: TODO: will move to mission/utils.

PR #808, opened 2025-06-30
Remove shooting method

kaushikponnapalli:
### Summary

Removed all references to shooting, simupy/SGM and analysis scheme. Now Collocation is the only option available

### Related Issues

- Resolves #

Shooting no longer works

### New Dependencies

None

### DIFF
--- a/aviary/docs/user_guide/off_design_missions.ipynb
+++ b/aviary/docs/user_guide/off_design_missions.ipynb
@@ -2,86 +2,11 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 5,
    "metadata": {
     "tags": [
      "remove-cell"
     ]
    },
-   "outputs": [
-    {
-     "data": {
-      "text/markdown": [
-       "```\n",
-       "settings:problem_type, alternate\n",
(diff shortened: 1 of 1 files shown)


aviary/mission/gasp_based/ode/flight_path_ode.py
```diff
@@ -66,34 +64,12 @@ def setup(self):
...
-            if kwargs['method'] == 'cruise':
-                SGM_required_inputs[Dynamic.Mission.FLIGHT_PATH_ANGLE] = {
-                    'val': 0,
-                    'units': 'deg',
-                }
-            add_SGM_required_inputs(self, SGM_required_inputs)
-            prop_group = om.Group()
-        else:
-            prop_group = self
+        prop_group = self
```
jkirk5: Since there isn't an alternate for prop_group anymore I think we can skip this, just use self everywhere now

xjjiang:
> Looks like there's one last mention of AnalysisScheme in `coupled_aircraft_mission_optimization.ipynb` which is causing the docs failure

There are more places in Docs.

PR #812, opened 2025-07-07
Post revamp payload range

### LINKED ISSUE #771: Design mass changes for off design FLOPS missions

aalemu1:
### Summary

Summary of PR.

Adds payload range functionality through a toggle within csv file, settings:payload_range,true. As well as rewrites the run_off_design_example.py for convergence. 

### Related Issues

- Resolves #771 

None

### New Dependencies

None


aviary/subsystems/mass/flops_based/landing_gear.py
```diff
@@ -70,21 +71,23 @@ def compute(self, inputs, outputs):
...
             0.048
             * landing_weight**0.67
             * nose_gear_length**0.43
             * nose_gear_scaler
             / GRAV_ENGLISH_LBM
         )
-
+        outputs[Aircraft.LandingGear.MAIN_GEAR_MASS] = main_gear_mass
+        outputs[Aircraft.LandingGear.NOSE_GEAR_MASS] = nose_gear_mass
+        outputs[Aircraft.LandingGear.TOTAL_MASS] = main_gear_mass + nose_gear_mass
```
Kenneth-T-Moore: There will be problems with this if the user overrides either main or nose gear mass in the csv file, but lets aviary calculate the other one, because the replacement will happen after this calculation outside of the component. The way to resolve this is to create a new component that takes nose and main gear mass as inputs and outputs the total landing gear mass.

aalemu1: I believe I have fixed this and pushed it up.

aviary/interface/methods_for_level2.py
```diff
@@ -1931,6 +1932,12 @@ def run_aviary_problem(
             self.final_setup()
             with open('input_list.txt', 'w') as outfile:
                 self.model.list_inputs(out_stream=outfile)
+                
+        #Creates a flag to determine if the user would or would not like a payload/range diagram
+        payload_range_bool = self.aviary_inputs.get_val(Settings.PAYLOAD_RANGE)
+
+        #Set it back to false, 
+        self.aviary_inputs.set_val(Settings.PAYLOAD_RANGE, False)
```
cmbenne3: I don't think the Aviary Input should get set back to 'false' - this seems like it would be confusing if someone interrogates the Aviary problem object after the run. Is there another way of implementing this?

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+                gross_mass = float(self.get_val(Mission.Summary.GROSS_MASS))
+                operating_mass=float(self.get_val(Aircraft.Design.OPERATING_MASS))
+                fuel_capacity=float(self.get_val(Aircraft.Fuel.TOTAL_CAPACITY))
+
+                if operating_mass + fuel_capacity < gross_mass:
+                    payload_range_matrix=self.run_payload_range()
+                    print(payload_range_matrix[1])
+                    print(payload_range_matrix[0])
+                else: 
+                    print("Off-design gross mass exceeds sizing limitations, payload/range cannot be generated")
```
cmbenne3: Perhaps the best behavior would be to throw a Warning to say that the fuel capacity is unnecessarily big, but to still run the Payload range points. The max fuel for Ferry Mission would need to be calculated as gross_mass - operating_mass and this could be less than the fuel_capacity?

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+        
+
+    def run_payload_range(self):
+        #Ensure proper transfer of json files. 
+        self.save_sizing_to_json(json_filename='payload_range_sizing.json')
+
+        #Automatically Adjust duration bounds of phase information within the cruise stage 
+        #to allow the optimizer to arrive at a local maxima.
+        min_duration = self.phase_info['cruise']['user_options']['time_duration_bounds'][0][0]
```
cmbenne3: This will fail if the user doesn't have a 'cruise' segment in their phase info. I don't think this is necessarily a problem, but should be noted in the documentation.

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+        max_duration = self.phase_info['cruise']['user_options']['time_duration_bounds'][0][1]
+        cruise_units = self.phase_info['cruise']['user_options']['time_duration_bounds'][1]
+
+        #Simply doubling the amount of time the optimizer is allowed to stay in the cruise phase, as well as ensure cruise is optimized
+        self.phase_info['cruise']['user_options'].update({'time_duration_bounds':((min_duration,2*max_duration), cruise_units)})
```
cmbenne3: Is it necessary to specify a max duration?
Would this work with a 10x duration or an absolute duration of e.g. 24 hrs?
Trying to make this as robust as possible to different aircraft types/models. 2x is probably fine for the majority of conventional aircraft, but there may be some where this isn't.
This should be documented as a limitation of the automatic payload/range functionality.

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+
+        #point 1 along the y axis (range=0)
+        payload_1 = float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS))
+        range_1 = 0 
+        #point 2, sizing mission payload and range
```
cmbenne3: Sizing mission is assumed to be the "Max Payload + Fuel" point

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+        operating_mass=float(self.get_val(Aircraft.Design.OPERATING_MASS))
+        fuel_capacity=float(self.get_val(Aircraft.Fuel.TOTAL_CAPACITY))
+        payload=float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS))
+
+        payload_allowed_mass=gross_mass-operating_mass-fuel_capacity
+
+        payload_frac=payload_allowed_mass/payload
+
+        #cargo does not invoke "aviary_inputs" because it is of units lbm, while payload frac is not
+        wing_cargo_allowed=int(self.get_val(Aircraft.CrewPayload.WING_CARGO))*payload_frac
```
cmbenne3: This will fail for GASP based missions.
Perhaps we should put the check for mission_method FLOPS/GASP inside the run_payload_range() method rather than in the top level run_aviary_problem()

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+        gross_mass = float(self.get_val(Mission.Summary.GROSS_MASS))
+        operating_mass=float(self.get_val(Aircraft.Design.OPERATING_MASS))
+        fuel_capacity=float(self.get_val(Aircraft.Fuel.TOTAL_CAPACITY))
+        payload=float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS))
+
+        payload_allowed_mass=gross_mass-operating_mass-fuel_capacity
+
+        payload_frac=payload_allowed_mass/payload
+
+        #cargo does not invoke "aviary_inputs" because it is of units lbm, while payload frac is not
```
cmbenne3: I don't understand this comment - I don't think the units has anything to do with the location of the cargo vs passenger variables within the aviary problem data structure?

aviary/interface/methods_for_level2.py
```diff
@@ -1987,9 +1994,108 @@ def run_aviary_problem(
...
+        prob_fallout_max_fuel = self.fallout_mission(json_filename='payload_range_sizing.json',
+                                    num_first= num_first_allowed, num_business= num_bus_allowed, num_tourist= num_tourist_allowed,
+                                    wing_cargo=wing_cargo_allowed, misc_cargo=misc_cargo_allowed)
+
+
+        payload_3=float(prob_fallout_max_fuel.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS))
+        range_3=float(prob_fallout_max_fuel.get_val(Mission.Summary.RANGE))
+
+        #point 4, ferry mission with max fuel and 0 payload,
+        allowed_mission_mass=operating_mass+fuel_capacity
```
cmbenne3: Is 'allowed'_variable the best naming convention? perhaps we should be more explicit and name these after each payload-range point?

aviary/interface/methods_for_level2.py
```diff
@@ -2444,10 +2573,13 @@ def _read_sizing_json(aviary_problem, json_filename):
             # Identify enums and manipulate the string to find the value
             tmp_var_values = var_values.split(':')[-1]
             var_values = (
-                tmp_var_values.replace('>', '').replace(']', '').replace("'", '').replace(' ', '')
+                tmp_var_values.replace('>', '').replace('<', '').replace(']', '').replace("'", '').replace(' ', '')
             )
             var_values = convert_strings_to_data([var_values])
 
+        if var_name == "aircraft:design:lift_polar":
+            print('here!')
```
cmbenne3: Is this leftover from debugging?
We probably want to remove it.

aviary/interface/methods_for_level2.py
```diff
@@ -2564,6 +2696,9 @@ def _load_off_design(
     prob.aviary_inputs.set_val(Aircraft.CrewPayload.CARGO_MASS, cargo_mass, 'lbm')
 
     if problem_type == ProblemType.ALTERNATE:
+        # Reset payload range toggle to False
+        prob.aviary_inputs.set_val(Settings.PAYLOAD_RANGE, False)
```
cmbenne3: As per comment 1 - is this the best implementation?

aviary/interface/methods_for_level2.py
aalemu1: The calculations done for the Max fuel + Payload Point find the maximum payload that will fit on the plane once the fuel capacity is full. 
```payload_allowed_mass=gross_mass-operating_mass-fuel_capacity```

if operating mass + fuel capacity > gross mass then the payload_allowed_mass will be negative, which makes that run impossible. 

The ferry range is calculated by setting the mission_mass equal to the operating mass + fuel capacity. If the mission_mass is greater than the design gross mass aviary will be running an invalid mission that exceeds the structural constraint of the sizing mission.

File: aviary/interface/methods_for_level2.py (hunk 5 above)
Comment (aalemu1) [reply aviary/interface/methods_for_level2.py]: Yes it is as multipliers greater than 2.5/3 can result in the fallout missions exceeding the max iterations.

File: aviary/api.py [hunk 12]
```diff
@@ -59,6 +59,9 @@
 from aviary.models.missions.height_energy_default import (
     phase_info as default_height_energy_phase_info,
 )
+from aviary.models.missions.off_design_height_energy import (
+    phase_info as default_off_design_hight_energy_phase_info,
```
jkirk5:
```suggestion
    phase_info as default_off_design_height_energy_phase_info,
```

aviary/variable_info/variables.py
```diff
@@ -794,11 +794,28 @@ class Taxi:
         MACH = 'mission:taxi:mach'
 
 
+class Payload_Range:
```
jkirk5: Make this a subcategory of `Mission`, and remove the underscore (`PayloadRange`).
So for example, a variable would look like `Mission.PayloadRange.XXXX`

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+        if (
+            mass_method == LegacyCode.FLOPS
+            and equations_of_motion is EquationsOfMotion.HEIGHT_ENERGY
+        ):
+            if payload_range_bool and self.model.problem_type is ProblemType.SIZING:
+                # Checks to determine if the set gross mass for off design would be greater
+                # Than the gross mass of the sizing mission.
+                payload_range_matrix = self.run_payload_range()
+                print(payload_range_matrix[1])
+                print(payload_range_matrix[0])
```
jkirk5: You can delete these prints, we already write the info and don't want to clutter the terminal

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
         self.problem_ran_successfully = not failed
 
+        mass_method = self.aviary_inputs.get_val(Settings.MASS_METHOD)
+        equations_of_motion = self.aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
+
+        # Off-design missions do not currently work for GASP masses or missions.
+        if (
+            mass_method == LegacyCode.FLOPS
+            and equations_of_motion is EquationsOfMotion.HEIGHT_ENERGY
+        ):
```
jkirk5: Maybe we should add `and if self.problem_ran_successfully` here. I can't think of a time when we want to keep running more cases for a failed run

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
 
         self.problem_ran_successfully = not failed
 
+        mass_method = self.aviary_inputs.get_val(Settings.MASS_METHOD)
+        equations_of_motion = self.aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
+
+        # Off-design missions do not currently work for GASP masses or missions.
+        if (
+            mass_method == LegacyCode.FLOPS
```
jkirk5:
```suggestion
            mass_method is LegacyCode.FLOPS
```

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            if payload_range_bool and self.model.problem_type is ProblemType.SIZING:
+                # Checks to determine if the set gross mass for off design would be greater
+                # Than the gross mass of the sizing mission.
+                payload_range_matrix = self.run_payload_range()
+                print(payload_range_matrix[1])
+                print(payload_range_matrix[0])
+
+    def run_payload_range(self):
+        # Ensure proper transfer of json files.
+        self.save_sizing_to_json(json_filename='payload_range_sizing.json')
```
jkirk5: Should this be named something like `design.json`, since it's saving the results of the sizing run and doesn't actually have anything to do w/ payload range yet? Might be confusing for a user looking at this file manually and expecting payload-range results

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+
+        #
+        # This checks if the 'cruise' phase exists, then automatically adjusts duration bounds of the cruise stage
+        # to allow the optimizer to arrive at a local maxim for the max_fuel_plus_payload and the ferry ranges.
+        if self.phase_info['cruise']:
```
jkirk5:
```suggestion
        if 'cruise' in self.phase_info:
```
You will raise a `KeyError` if you try and directly access a key that doesn't exist, so you have to check for it's existence this way

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+                # Than the gross mass of the sizing mission.
+                payload_range_matrix = self.run_payload_range()
+                print(payload_range_matrix[1])
+                print(payload_range_matrix[0])
+
+    def run_payload_range(self):
+        # Ensure proper transfer of json files.
+        self.save_sizing_to_json(json_filename='payload_range_sizing.json')
+
+        #
```
jkirk5:
```suggestion
```

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            self.phase_info['cruise']['user_options'].update(
+                {'time_duration_bounds': ((min_duration, 2 * max_duration), cruise_units)}
+            )
+
+        # point 1 along the y axis (range=0)
+        payload_1 = float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS)[0])
+        range_1 = 0
+
+        # Set variables for variable hierarchy
+        self.aviary_inputs.set_val(Payload_Range.MAX_PAYLOAD_ZERO_FUEL_PAYLOAD, payload_1, 'lbm')
```
jkirk5: I'm reconsidering if we need these variables at all - the info is already written to a file so the dashboard won't need them, and these variables aren't used by the problem after they are saved

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            max_duration = self.phase_info['cruise']['user_options']['time_duration_bounds'][0][1]
+            cruise_units = self.phase_info['cruise']['user_options']['time_duration_bounds'][1]
+
+            # Simply doubling the amount of time the optimizer is allowed to stay in the cruise phase, as well as ensure cruise is optimized
+            self.phase_info['cruise']['user_options'].update(
+                {'time_duration_bounds': ((min_duration, 2 * max_duration), cruise_units)}
+            )
+
+        # point 1 along the y axis (range=0)
+        payload_1 = float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS)[0])
```
jkirk5: Is this ever not a float? I just ran an example to check and it was a float in the array

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            )
+
+            payload_3 = float(
+                prob_fallout_max_fuel.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS)
+            )
+            range_3 = float(prob_fallout_max_fuel.get_val(Mission.Summary.RANGE))
+        else:
+            fuel_capacity = gross_mass - operating_mass
+            payload_3 = payload_2
+            range_3 = range_2
```
jkirk5: This section doesn't make sense to me - you are filling the aircraft from OEW to TOGW with fuel, then adding back on the design (max) cargo, and also assuming range is the same as design? This scenario becomes a duplicate of point 4 (just happens to stay at TOGW here)

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+                * payload_frac
+            )
+            max_fuel_plus_payload_num_bus = int(
+                (self.aviary_inputs.get_val(Aircraft.CrewPayload.Design.NUM_BUSINESS_CLASS))
+                * payload_frac
+            )
+            max_fuel_plus_payload_num_tourist = int(
+                (self.aviary_inputs.get_val(Aircraft.CrewPayload.Design.NUM_TOURIST_CLASS))
+                * payload_frac
+            )
```
jkirk5: We can avoid the rounding here - mission analysis doesn't care about how the payload is divided, it just needs total mass. The current off-design implementation has some issues that can't be fixed until it's rewritten, but you might be able to add all of the payload mass into one of the variables and set everything else to zero to avoid the rounding for now

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+        self.aviary_inputs.set_val(Payload_Range.MAX_FUEL_ZERO_PAYLOAD_RANGE, range_4, 'NM')
+
+        # writes csv file in dashboard
+        script_name = os.path.splitext(os.path.basename(sys.argv[0]))[0]
+        output_dir = f'{script_name}_out'
+        reports_dir = os.path.join(output_dir, 'reports')
+
+        os.makedirs(reports_dir, exist_ok=True)
+
+        csv_filepath = os.path.join(reports_dir, 'payload_range_data.csv')
```
jkirk5:
```suggestion
csv_filepath = Path(prob.get_reports_dir()) / 'payload_range_data.csv'
```

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            # Write the four points directly
+            writer.writerow(['Max Payload Zero Fuel', payload_1, range_1])
+            writer.writerow(['Max Payload Plus Fuel', payload_2, range_2])
+            writer.writerow(['Max Fuel Plus Payload', payload_3, range_3])
+            writer.writerow(['Ferry Mission', payload_4, range_4])
+
+        payload_points = ['Payload (lbs)', payload_1, payload_2, payload_3, payload_4]
+        range_points = ['Range (NM)', range_1, range_2, range_3, range_4]
+
+        return (payload_points, range_points)
```
jkirk5: Doesn't seem to be a need to return the points - unused

aviary/interface/methods_for_level2.py
```diff
@@ -743,6 +747,157 @@ def run_aviary_problem(
...
+            writer.writerow(['Point', 'Payload (lbs)', 'Range (NM)'])
+
+            # Write the four points directly
+            writer.writerow(['Max Payload Zero Fuel', payload_1, range_1])
+            writer.writerow(['Max Payload Plus Fuel', payload_2, range_2])
+            writer.writerow(['Max Fuel Plus Payload', payload_3, range_3])
+            writer.writerow(['Ferry Mission', payload_4, range_4])
+
+        payload_points = ['Payload (lbs)', payload_1, payload_2, payload_3, payload_4]
+        range_points = ['Range (NM)', range_1, range_2, range_3, range_4]
```
jkirk5: If you save the data in a NamedValues object, you can use the existing csv writing utilities and take advantage of the formatting, timestamps, etc. it uses. That will also make it easy to read the data back in later

`write_data_file(filename, data, include_timestamp=True)`

aviary/interface/methods_for_level2.py
```diff
@@ -822,15 +977,23 @@ def alternate_mission(
             num_first = num_business = num_tourist = wing_cargo = misc_cargo = 0
 
         if phase_info is None:
+            # Somewhere between the sizing and off-design self.pre_mission_info gets deleted
```
jkirk5:
```suggestion
            # model.phase_info only contains mission information
```
Doesn't get deleted, just moved!

aviary/interface/methods_for_level2.py
```diff
@@ -822,15 +977,23 @@ def alternate_mission(
...
         if mission_range is None:
             # mission range is sliced from a column vector numpy array, i.e. it is a len
             # 1 numpy array
-            mission_range = self.get_val(Mission.Design.RANGE)[0]
+            # fixes issue wherein an alternate mission with no inputs "alternate_mission()" attempts to run the range
+            # within the CSV file instead of in the phase_info.
+            if mass_method == LegacyCode.GASP:
+                mission_range = self.get_val(Mission.Design.RANGE)[0]
+            elif mass_method == LegacyCode.FLOPS:
+                mission_range = self.model.post_mission_info['target_range'][0]
```
jkirk5: I'm not sure this covers all edge cases, assuming that Mission.Design.RANGE or 'target_range' does or does not exist based on mass method. What is in the phase info is determined by the mission method, and your off-design example probably doesn't test any mixed FLOPS/GASP examples.

Can we check 'target_range' exists first, then fall back to Mission.Design.RANGE if it doesn't? No code origin checks required?

aviary/interface/methods_for_level2.py
```diff
@@ -1261,15 +1444,9 @@ def _load_off_design(
                     'no specified Range'
                 )
         else:
-            prob.aviary_inputs.set_val(Mission.Design.RANGE, mission_range, units='NM')
+            """prob.aviary_inputs.set_val(Mission.Design.RANGE, mission_range, units='NM')"""
```
jkirk5, line 1504:
```suggestion
```

aviary/interface/methods_for_level2.py
```diff
@@ -685,6 +686,9 @@ def run_aviary_problem(
             with open('input_list.txt', 'w') as outfile:
                 self.model.list_inputs(out_stream=outfile)
 
+        # Creates a flag to determine if the user would or would not like a payload/range diagram
+        payload_range_bool = self.aviary_inputs.get_val(Settings.PAYLOAD_RANGE)
```
jkirk5:
```suggestion
        payload_range_bool = False
        if Settings.PAYLOAD_RANGE in self.aviary_inputs:
            payload_range_bool = self.aviary_inputs.get_val(Settings.PAYLOAD_RANGE)
```
We don't know that Settings.PAYLOAD_RANGE was always added to the csv, so we need to check it exists first.

aviary/models/aircraft/test_aircraft/aircraft_for_bench_FwFm.csv
```diff
@@ -155,4 +155,5 @@ mission:takeoff:lift_coefficient_max,3.0,unitless
 mission:takeoff:lift_over_drag,17.354,unitless
 settings:equations_of_motion,height_energy
 settings:aerodynamics_method,FLOPS
-settings:mass_method,FLOPS
\ No newline at end of file
+settings:mass_method,FLOPS
+settings:payload_range,False
```
jkirk5:
```suggestion
settings:mass_method,FLOPS
```
After applying changes to methods for L2 this won't be necessary anymore

aviary/models/missions/off_design_height_energy.py
jkirk5, line 1: add 'default' to filename to match the naming pattern of the other files

aviary/variable_info/variable_meta_data.py
```diff
@@ -8011,6 +8011,101 @@
     default_value=0.0001,
 )
 
+#  .----------------.  .----------------.  .----------------.  .----------------.  .----------------.  .----------------.  .----------------.  .----------------.
```
jkirk5: Based on earlier comments in methods_for_level2 I think we can remove these variables from the metadata

aviary/variable_info/test/test_var_structure.py
jkirk5, line 1: These changes might need to be reverted if payload-range variables are removed from metadata

aviary/variable_info/variables.py
jkirk5: Ignore this comment if we're deleting the vars

aviary/visualization/dashboard.py
```diff
@@ -1368,6 +1436,16 @@ def dashboard(script_name, problem_recorder, driver_recorder, port, run_in_backg
         Path(reports_dir) / 'mission_timeseries_data.csv',
     )
 
+    # Paylaod Range Output Pane
+    create_payload_range_frame(
+        'Payload/Range Output Chart',
```
jkirk5:
```suggestion
        'Payload/Range Diagram',
```

aviary/interface/methods_for_level2.py
jkirk5: After offline discussion we will keep it this way for now and come up with alternate solutions later

aviary/interface/methods_for_level2.py
jkirk5: Could be a follow-on thing, not critical to do right now

aviary/interface/methods_for_level2.py
aalemu1: This section was due to an edge case where a user might have a mission flying a short range mission with a small payload resulting in a sizing mission not representative of "maximum payload" therefore the aircraft's gross mass would be lighter than expected. Therefore trying to fill up that aircraft to its max fuel for the max fuel + payload and the ferry range results in errors.  

This allows that type of mission to still yield a payload range diagram.

aviary/interface/methods_for_level2.py
```diff
@@ -732,6 +738,213 @@ def run_aviary_problem(
...
+                # point 1 along the y axis (range=0)
+                payload_1 = float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS)[0])
+                range_1 = 0
+
+                # point 2, sizing mission which is assumed to be the point of max payload + fuel on the payload and range diagram
+                payload_2 = payload_1
+                range_2 = float(self.get_val(Mission.Summary.RANGE)[0])
+
+                # check if fuel capacity does not exceed sizing mission design gross mass
+                gross_mass = float(self.get_val(Mission.Summary.GROSS_MASS)[0])
```
jkirk5, line 804:
```suggestion
                gross_mass = float(self.get_val(Mission.Design.GROSS_MASS)[0])
```
If you run payload-range after running a different off-design mission you might pull the wrong mass here if you don't specifically call for design? Just covering an edge case

aalemu1: The function checks that the mission the payload/range is being ran on is a sizing mission and would disallow the user from attempting it. Therefore this edge case, I don't believe is possible anymore.

PR #815, opened 2025-07-09
Correct use of detailed wing mass flag

jkirk5:
### Summary

Updated wing group in FLOPS mass to use the existing flag for toggling detailed wing mass.
Renamed `Aircraft.Wing.DETAILED_WING` to `Aircraft.Wing.USE_DETAILED_MASS`.
Updated some file and variable names in the OAS example that got overlooked in previous work
Fixed a warning for the `solve_alpha` flag & FLOPS aero being flipped (only printing when it shouldn't)


aviary/docs/examples/OAS_subsystem.ipynb
jkirk5, line 1: Replaced references to "wing weight" with "wing mass", updated filenames/class names for OAS subsystem. No change in docs content - this is true for all docs modified in this PR

PR #816, opened 2025-07-09
Phase info Revamp 4: propagate changes into the 2dof and solved2dof phases.

Kenneth-T-Moore:
### Summary

Phase info Revamp 4: propagate changes into the 2dof and solved2dof phases.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -1464,10 +1464,9 @@ def set_initial_guesses(self, parent_prob=None, parent_prefix='', verbosity=None
 
             if self.mission_method is SOLVED_2DOF:
                 self.phase_objects[idx].apply_initial_guesses(self, 'traj', phase)
-                if (
-                    self.phase_info[phase_name]['user_options']['ground_roll']
-                    and self.phase_info[phase_name]['user_options']['fix_initial']
-                ):
+                if self.phase_info[phase_name]['user_options'].get(
+                    'ground_roll'
+                ) and not self.phase_info[phase_name]['user_options'].get('rotation'):
```
jkirk5: Was this intentional? This is pretty different behavior and I'm not sure I follow the new logic

aviary/docs/user_guide/example.csv
```diff
@@ -1,7 +1,4 @@
-# This is test data generated using write_data_file
-# Aviary data tables can support multiple comments
-
```
jkirk5, line 3: Why were the comments removed? The docs read & print these out to demonstrate how comments are stored, so we need them.

jkirk5: Actually, since this file gets written by the docs themselves, it should probably be deleted entirely

aviary/core/aviary_group.py
Kenneth-T-Moore: Fix initial is gone now, and it was used here to indicate that the first phase is the one we are trying to select with this 'if' statement. There are 2 takeoff phases that have ground_roll, and the first one does not have rotation, so it matches the same phase.

I could probably do the same thing by checking `idx == 0` instead, which is a more direct way to find the first phase.

aviary/docs/user_guide/example.csv
Kenneth-T-Moore: yeah, I think an earlier PR must have inserted these comments into the auto-generated file. We only see the change because I built the docs while working this PR.

jkirk5 (approved):

PR #817, opened 2025-07-11
Address a numpy deprecation warning about assigning a vector to a single vector index.

Kenneth-T-Moore:
### Summary

Address a numpy deprecation warning about assigning a vector to a single vector index.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/mass/flops_based/wing_detailed.py
+++ b/aviary/subsystems/mass/flops_based/wing_detailed.py
@@ -242,7 +242,7 @@ class DetailedWingBendingFact(om.ExplicitComponent):
 
                 bte = 8 * np.sum((ea[:-1] + ea[1:]) * dy[:-1] * 0.5)
 
-                inertia_factor_i = 1 - bte / bt * pod_mass[i] / gross_mass
+                inertia_factor_i = 1 - bte / bt[0] * pod_mass[i] / gross_mass[0]
                 # avoid passing an array into specific index of inertia_factor
                 inertia_factor[i] = inertia_factor_i
 


jkirk5:
Do we know why those variables are vectors when they shouldn't be? I'm fine with this fix but we might run into this issue in other places if we don't address the root cause

Kenneth-T-Moore:
> Do we know why those variables are vectors when they shouldn't be? I'm fine with this fix but we might run into this issue in other places if we don't address the root cause

It seems to mainly be openmdao behavior right now that everything you get from `inputs` is a vector, no matter how you declare it.


        import numpy as np
        
        import openmdao.api as om
        
        class MyComp(om.ExplicitComponent):
        
            def setup(self):
        
                self.add_input('x1', val = 1.0)
                self.add_input('x2', val = np.array((1, )))
                self.add_input('x3', shape=1)
                self.add_input('x4', shape=(1, ))
                self.add_input('x5', shape=0)
                self.add_output('zzz')
        
            def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
                
                print('x1', inputs['x1'])
                print('x2', inputs['x2'])
                print('x3', inputs['x3'])
                print('x4', inputs['x4'])
                print('x5', inputs['x5'])
        
        prob = om.Problem()
        model = prob.model
        model.add_subsystem('comp', MyComp())
        prob.setup()
        
        prob.run_model()
        print('done')


      /Users/ktmoore1/Work/OpenMDAO/OpenMDAO/openmdao/visualization/n2_viewer/n2_viewer.py:114: OpenMDAOWarning:All-NaN slice encountered
      x1 [1.]
      x2 [1.]
      x3 [1.]
      x4 [1.]
      x5 []
      done

jkirk5 (approved):

PR #827, opened 2025-07-25
Performance subsystem

jkirk5:
### Summary

Adds the skeleton for the "Performance" subsystem, which currently only contains a pre-mission component to compute wing loading and thrust-to-weight-ratio. There is also a new example that showcases a custom optimization problem with new design variables and constraints to demonstrate using this subsystem.

TODOs include a subsystem report, documentation for the subsystem and for the example.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -1,43 +1,34 @@
 import inspect
-from pathlib import Path
```
jkirk5, line 2: ruff import sorting

PR #829, opened 2025-07-30
GASP aero table conversion

xjjiang:
### Summary

The current conversion of GASP aero table does not convert real GASP aero tables. I believe the current conversion works on GASP outputs. The real GASP aero tables come from running high fidelity CFD tools and they have a different format.

This PR allows users to convert real GASP aero tables (the CL and CD tables). The file format is flagged as `GASP`. The existing aero table format is not the actual GASP aero table format. User still can convert them but it is flagged as `GASP_ALT` now.

In this case, we assume that `CL = f(Mach, Altitude, Alpha)` and `CD = f(Mach, Altitude, CL)`. There is a possibility that `Altitude` is replaced by `Reynolds` but we have not implemented this option.

In this case, we assume that there are only two tables: CL table and CD table. There is a possibility that a third CD0 table is provided and CD table is actually CDi table. but we have not implemented this option.

Test is done for BWB cruise using table based aero table. There is no low speed aero table test because we don't have aero tables for flaps and ground effect. GASP never has those tables.

Document is updated in "Command Line Tools" page. Read the doc page to see how to use it.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/subsystems/aerodynamics/gasp_based/test/test_table_based.py
```diff
@@ -241,5 +241,66 @@ def test_case(self):
...
 
 
+class BWBCruiseAeroTest(unittest.TestCase):
+    """
+    Test of table based BWB cruise using a table generated by a CFD tool.
+    The table is modified for demonstration purposes only. It does not represent any actual result.
+    """
+
+    @unittest.skipIf(
+        version.parse(openmdao.__version__) < version.parse('3.26'),
```
Kenneth-T-Moore: We can actually take this skipIf out wherever we see it, because Aviary is pinned to openmdao>=3.37.0.

aviary/utils/engine_deck_conversion.py
```diff
@@ -193,9 +193,10 @@ def convert_engine_deck(input_file, output_file, data_format: EngineDeckType, ro
             structured_data = _make_structured_grid(
                 tables, method='lagrange3', fields=fields, throttle_step=throttle_step
             )
-            structured_data = _make_structured_grid(
-                tables, method='lagrange3', fields=fields, throttle_step=throttle_step
-            )
+            # why twice?
+            # structured_data = _make_structured_grid(
```
Kenneth-T-Moore: Yeah, i am pretty sure this is a cut and paste error, and we can remove it safely.

aviary/utils/aero_table_conversion.py
```diff
@@ -66,11 +91,72 @@ def convert_aero_table(input_file=None, output_file=None, data_format=None):
 
     stamp = f'# {data_format.value}-derived aerodynamics data converted from {data_file.name}'
 
-    if data_format is CodeOrigin.GASP:
-        data, comments = _load_gasp_aero_table(data_file)
-        comments = [stamp] + comments
-
-        write_data_file(output_file, data, outputs, comments, include_timestamp=True)
+    if data_format in (CodeOrigin.GASP, CodeOrigin.GASP_ALT):
+        is_alternative = True if data_format == CodeOrigin.GASP_ALT else False
```
Kenneth-T-Moore: If I am not mistaken, you can just simplify this whole if structure to just 

if data_format is CodeOrigin.GASP_ALT

elif data_format is CodeOrigin.GASP

elif data_format is CodeOrigin.FLOPS
...

aviary/utils/aero_table_conversion.py
```diff
@@ -257,6 +343,151 @@ def _load_gasp_aero_table(filepath: Path):
...
+
+        if tab_data is None:
+            tab_data = map_data
+        else:
+            tab_data = np.r_[tab_data, map_data]
+
+    return tab_data
+
+
+def _make_structured_grid(data, method='lagrange3', fields=['CL', 'CD']):
```
Kenneth-T-Moore, line 423: This looks like the one that is in engine_deck_conversion, but with aero names. I suppose we might want to generalize it at some point in the future.

aviary/subsystems/aerodynamics/gasp_based/test/test_table_based.py
xjjiang: Removed it from this file for now.

aviary/utils/engine_deck_conversion.py
xjjiang: Removed it.

aviary/utils/aero_table_conversion.py
xjjiang: True. I will come back to this for other scenarios (Reynolds number and cd0) in the GASP aero tables when I have time. Otherwise, this can be done when a need comes up.

aviary/utils/aero_table_conversion.py
xjjiang: Done.

jkirk5 (approved):
Since the split table is the "true" GASP format I suggest flipping `GASP` and `GASP_ALT` table formats

PR #834, opened 2025-08-02
[Draft] Update fortran_to_aviary for BWB model

xjjiang:
### Summary

Summary of PR.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/utils/fortran_to_aviary.py
```diff
@@ -136,6 +136,7 @@ def fortran_to_aviary(
     vehicle_data['input_values'].set_val(Settings.EQUATIONS_OF_MOTION, eom)
     vehicle_data['input_values'].set_val(Settings.MASS_METHOD, mass)
     vehicle_data['input_values'].set_val(Settings.AERODYNAMICS_METHOD, aero)
+    vehicle_data['input_values'].set_val(Aircraft.Engine.DATA_FILE, ['engine_data_file'])
```
jkirk5: This value is not the default nor is it an existing engine data file - if the original Fortran file is missing this input, then the Aviary csv should be missing it as well. Basically, if an incomplete Fortran file is given to the converter, then it is reasonable to expect an incomplete Aviary file comes out of it.

jkirk5: If removing this line causes some other kind of error, then I'd like to try and fix it at the source of the error rather than rely on an arbitrary/fake value being hardcoded in the converter

aviary/utils/fortran_to_aviary.py
```diff
@@ -529,6 +530,10 @@ def update_gasp_options(vehicle_data):
         num_passengers = input_values.get_val(
             Aircraft.CrewPayload.Design.NUM_PASSENGERS, 'unitless'
         )[0]
+        num_passengers = int(num_passengers)
```
jkirk5, line 532: This can probably be fixed in the future if we use an AviaryValues object instead of NamedValues, so it can do type-checking against the metadata. Right now that isn't possible because the converter is storing every single variable in a list regardless of if it is supposed to be or not

aviary/utils/fortran_to_aviary.py
```diff
@@ -633,6 +663,14 @@ def update_gasp_options(vehicle_data):
         input_values.set_val(Aircraft.Design.RESERVE_FUEL_FRACTION, [0], units='unitless')
     else:
         ValueError('"FRESF" is not valid between 0 and 10.')
+    try:
+        if input_values.get_val(Mission.Landing.MAXIMUM_FLARE_LOAD_FACTOR)[0] > 4:
+            print(
```
jkirk5: Add a verbosity check before printing - BRIEF or greater is fine

aviary/utils/fortran_to_aviary.py
```diff
@@ -537,6 +542,22 @@ def update_gasp_options(vehicle_data):
         input_values.set_val(
             Aircraft.CrewPayload.Design.NUM_FIRST_CLASS, [num_first_class], 'unitless'
         )
+        num_tourist_class = num_passengers - num_first_class
+        input_values.set_val(
+            Aircraft.CrewPayload.Design.NUM_TOURIST_CLASS, [num_tourist_class], 'unitless'
+        )
+
+    except:
```
jkirk5: We shouldn't be using bare excepts. I'm guessing a `KeyError` is what we are looking for?
Ideally we want only the line(s) that might cause the expected error to be in the `try:`, and then the rest of the code can be moved to the `else:` section.

aviary/utils/fortran_to_aviary.py
```diff
@@ -663,6 +704,56 @@ def update_gasp_options(vehicle_data):
...
+                Aircraft.Engine.NUM_FUSELAGE_ENGINES, [num_fuselage_engines], 'unitless'
+            )
+            # BWB engine sizing algorithm does not use reference diameter
+            input_values.delete(Aircraft.Engine.REFERENCE_DIAMETER)
+    except:
+        pass
+    try:
+        engine_type = input_values.get_val(Aircraft.Engine.TYPE, 'unitless')[0]
+        if not (engine_type == 6 or engine_type == 7):
+            print('Only TURBOPROP(6) and TURBOJET(7) are allowed for now')
```
jkirk5:
```suggestion
            print(
                f'Engine type {engine_type} was provided; currently only TURBOPROP(6) and '
                'TURBOJET(7) are supported by Aviary'
            )
```
The formatting might be off here (hard to do in github), double check with ruff. Also, add a verbosity check before printing

aviary/utils/fortran_to_aviary.py
xjjiang: Removed it.

aviary/utils/fortran_to_aviary.py
xjjiang: We will revisit this utility in the future.

aviary/utils/fortran_to_aviary.py
xjjiang: Done.

aviary/utils/fortran_to_aviary.py
xjjiang: Done.

aviary/utils/fortran_to_aviary.py
xjjiang: The try-except block is here just in case that GASP dataset does not have the variable. For example, if `PAX` is not in GASP input deck, then `Aircraft.CrewPayload.Design.NUM_PASSENGERS` is not in the Aviary dataset. The except block just ignore the post-process in the `try` block.

Kenneth-T-Moore:
> @jkirk5 @Kenneth-T-Moore , One thing that I am concerned but not touched is the size of `aviary/utils/legacy_code_data/gasp_defaults.py`. There are more than 150 default values. It means that Aviary adds those variables to user's model but they may not even notice them.

Does the Fortran version of Gasp also have all of these built-in defaults?

I agree that we should migrate whatever we can into the gasp aircraft input files.

aviary/utils/fortran_to_aviary.py
crecine: instead of a try-except block, you could us an 
`if Aircraft.CrewPayload.Design.NUM_PASSENGERS in input_values:`
(similar to [Line 707](https://github.com/xjjiang/om-Aviary/blob/ab324845aee055a2b20b350407d4ec0e1c76d8bf/aviary/utils/fortran_to_aviary.py#L707)

xjjiang: Good idea. Will do.

aviary/utils/fortran_to_aviary.py
```diff
@@ -663,6 +698,118 @@ def update_gasp_options(vehicle_data):
...
+    knac = unused_values.get_item('INGASP.KNAC')[0][0]
+    if knac != 2:
+        try:
+            input_values.delete(Aircraft.Nacelle.AVG_DIAMETER)
+            input_values.delete(Aircraft.Nacelle.AVG_LENGTH)
+        except:
+            pass
+
+    # Variables required by GASP, but no default values are provided in GASP
+    if not Aircraft.Wing.ZERO_LIFT_ANGLE in input_values:
```
Kenneth-T-Moore, line 779: I wonder if it might be better to loop over all of these variables, put the missing ones in the list, and then raise the error once.  That way, the user can fix them all at once before running the converter again.

xjjiang: You are right. I made the change.

jkirk5: I don't see this change yet, was it not pushed yet?

xjjiang: It is pushed. Here:

https://github.com/xjjiang/om-Aviary/blob/BWB_fortran_to_aviary/aviary/utils/fortran_to_aviary.py#L751

xjjiang:
@jkirk5 @Kenneth-T-Moore : I removed the line `aircraft:engine:scale_factor,0.6824747298710352,unitless` in `aviary/models/aircraft/blended_wing_body/generic_BWB_GASP.csv` in order that the bench test for BWB passes. It was added by `fortran_to_aviary.py`. I am not sure why it causes problem for `test_bwb_GwGm.py` to run. Do you know why?

jkirk5:
> @jkirk5 @Kenneth-T-Moore : I removed the line `aircraft:engine:scale_factor,0.6824747298710352,unitless` in `aviary/models/aircraft/blended_wing_body/generic_BWB_GASP.csv` in order that the bench test for BWB passes. It was added by `fortran_to_aviary.py`. I am not sure why it causes problem for `test_bwb_GwGm.py` to run. Do you know why?

I figured it why this is happening - in the `gasp_defaults.py` file `INGASP.FN_REF` is set to 28690 lbf (which seems to be completely made up, since a comment mentions that variable may not even exist in GASP. The converter sees that scaled thrust and reference thrust are both provided, so it calculates the scale factor needed. That scale factor is based on nothing so it just screws the model up.

Both the fake GASP default, and the entire part of `fortran_to_aviary.py` that computes the scale factor (`update_aviary_options()`) should get removed. The preprocessors already do the scale factor calculation.

PR #836, opened 2025-08-05
Combined Core & External Subsystem Handling

### LINKED ISSUE #418: Unify how core and external subsystems are handled

In methods_for_level2.py, core subsystems and external subsystems are handled separately, but as they both use the SubsystemBuilderBase interface. It should be possible to unify how both are loaded.

_No response_

### Associated Bug Report

_No response_

jkirk5:
### Summary

Combined how Aviary handles "core" and "external" subsystems - once the user has provided the external subsystems that will be used, they are combined into a single list and no longer handled separately, with a couple exceptions I haven't unraveled yet.

External subsystems are no longer loaded through phase info to avoid making many copies of the same builder. If a user wants to only load their external subsystem for certain phases, then they can add a flag using "subsystem_options" and add some custom behavior to their subsystem to do what they need. This is so much rarer of a scenario so the decision to make that functionality slightly more complicated to simplify every other external subsystem user's models.

API Changes:
- Removed "external_subsystems" key from phase info
- New optional method for L2 interface: `load_external_subsystems()`, the new place subsystem builders are passed to an `AviaryProblem()`
- Combined `add_external_subsystems()` and `add_core_subsystems()` methods in `BaseODE` into a single `add_subsystems()`

### Related Issues

- Resolves #418 

Any models using external subsystems will need to change how they are added to the problem using the L2 interface instead of through phase info

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -15,6 +16,7 @@
 from aviary.mission.solved_two_dof_problem_configurator import SolvedTwoDOFProblemConfigurator
 from aviary.mission.two_dof_problem_configurator import TwoDOFProblemConfigurator
 from aviary.mission.utils import get_phase_mission_bus_lengths, process_guess_var
+from aviary.mission.utils import get_phase_mission_bus_lengths, process_guess_var
```
Kenneth-T-Moore: Several imports are repeated here, probably due to a weird merge conflict.

aviary/core/aviary_group.py
```diff
@@ -54,12 +58,13 @@ class AviaryGroup(om.Group):
...
         super().__init__(**kwargs)
 
-        self.pre_mission = PreMissionGroup()
         self.post_mission = PostMissionGroup()
         self.verbosity = verbosity
+        self.external_subsystems = []
         self.regular_phases = []
         self.reserve_phases = []
 
+        self.meta_data = BaseMetaData
```
Kenneth-T-Moore: This also looks like a merge error. It is redefined as None 2 lines later.

aviary/core/aviary_group.py
```diff
@@ -54,12 +58,13 @@ class AviaryGroup(om.Group):
     def __init__(self, verbosity=None, **kwargs):
         super().__init__(**kwargs)
 
-        self.pre_mission = PreMissionGroup()
```
Kenneth-T-Moore, line 58: I hadn't noticed before, but yeah, self.post_mission and self.pre_mission do collide with the subsystem names, which OpenMDAO later stores as attributes. Removing the local instantiation here, like you did, is probably cleaner for both.

aviary/interface/methods_for_level2.py
```diff
@@ -86,6 +79,7 @@ def __init__(
             self.model = om.Group()
         else:
             self.model = AviaryGroup()
+            self.meta_data = BaseMetaData.copy()
```
Kenneth-T-Moore: This short-circuits the meta_data passed in as the argument to AviaryProblem.

aviary/interface/methods_for_level2.py
```diff
@@ -144,10 +138,10 @@ def load_inputs(
 
         return self.aviary_inputs
 
-    def check_and_preprocess_inputs(self, verbosity=None):
+    def load_external_subsystems(self, external_subsystems: list = [], verbosity=None):
```
Kenneth-T-Moore, line 152: I think some thought needs to be given to how to add external subsystems to multi-mission cases, since that capability probably arrived in `main` after this method of loading them was designed.  One answer might be moving this method into aviary_group instead.

aviary/interface/methods_for_level2.py
jkirk5: Both of these metadata issues come from a merge commit, so I will remove. Seems like there were some merge problems with this PR

aviary/interface/methods_for_level2.py
jkirk5: I will try that out with the idea of calling this function at the AviaryProblem level adds the subsystem to all AviaryGroups, or you could directly call it per AviaryGroup if you are doing something weird and custom with different subsystems per group

aviary/core/aviary_group.py
```diff
@@ -266,14 +265,44 @@ def load_inputs(
...
+        # This function sets all the following defaults if they were not already set:
+        # self.engine_builders, self.pre_mission_info, self_post_mission_info,
         # self.require_range_residual, self.target_range
-        # Other specific self.*** are defined in here as well that are specific to
-        # each builder
+        # Other specific self.*** are defined in here as well that are specific to each builder
 
         return self.aviary_inputs, self.verbosity
 
+    def load_external_subsystems(self, external_subsystems: list = [], verbosity=None):
```
Kenneth-T-Moore, line 275: Default value for external_subsystems should be None.

(When we make it a list, the default can change during subsequent runs.)

aviary/core/aviary_group.py
```diff
@@ -266,14 +265,44 @@ def load_inputs(
...
+        # `self.verbosity` is "true" verbosity for entire run. `verbosity` is verbosity
+        # override for just this method
+        if verbosity is not None:
+            # compatibility with being passed int for verbosity
+            verbosity = Verbosity(verbosity)
+        else:
+            verbosity = self.verbosity  # defaults to BRIEF
+
+        for subsystem in external_subsystems:
+            if not isinstance(subsystem, SubsystemBuilder) and verbosity >= verbosity.BRIEF:
```
Kenneth-T-Moore, line 296: I kind of wonder if we should just raise an error instead.

aviary/core/aviary_group.py
```diff
@@ -351,26 +380,23 @@ def check_and_preprocess_inputs(self, verbosity=None):
                     else:
                         guesses['time'] = ((None, time_duration), units)
 
-        for phase_name in self.mission_info:
-            for external_subsystem in self.mission_info[phase_name]['external_subsystems']:
-                aviary_inputs = external_subsystem.preprocess_inputs(aviary_inputs)
+        for external_subsystem in self.external_subsystems:
```
Kenneth-T-Moore, line 383: By moving external_subsystems out of phase_info, we lose the ability to add a subsystem to just a single selected phase.  One solution to this might be to add "phase_name" as an argument to SubsystemBuilder.build_mission().

aviary/subsystems/subsystem_builder.py
```diff
@@ -3,7 +3,7 @@
 from aviary.variable_info.variable_meta_data import _MetaData
```
Kenneth-T-Moore, line 5: Probably should rename file too.

aviary/subsystems/subsystem_builder.py
```diff
@@ -319,22 +318,22 @@ def preprocess_inputs(self, aviary_inputs):
         """
         return aviary_inputs
 
-    def get_outputs(self):
+    def get_timeseries(self):
```
Kenneth-T-Moore, line 328: For this rename, I recommend adding `get_outputs` back, and give it a deprecation warning (or even an error). This will help us catch them.

aviary/core/aviary_group.py
jkirk5: I'd like to discourage using phase names as an expected argument- I intend for users to use `subsystem_options` instead, and toggle their subsystems off via a flag they add, rather than hardcode the subsystems to only work for phases of a certain name, which makes them very challenging to use with any other mission. Added benefit of no additional api changes needed.

aviary/subsystems/subsystem_builder.py
jkirk5: Duh, forgot that very important step! fixed

Kenneth-T-Moore (approved):

PR #840, opened 2025-08-11
Off design overhaul

jkirk5:
### Summary

- Modified off-design mission API so that saving and loading json file is optional. Off-design missions now have 2 methods of being used (still requires L2):
    1. Directly run an off-design mission from an AviaryProblem in memory after sizing mission using method call
    2. Load a previously saved AviaryProblem from json format, then use method 1 to execute the off-design mission

- Running an off-design mission has been streamlined into a single method call, combining the separate methods for Fallout and Alternate missions.

- Saving/loading a problem to and from json format have been separated out into method/function calls completely separate from off-design mission creation. Saving the problem is an AviaryProblem method, while loading the problem is a function call that has been added to the Aviary api.

- `AviaryProblem.phase_info` has been renamed to `AviaryProblem.mission_info` to clarify what information is actually contained in that variable and match naming convention of `pre_mission_info` and `post_mission_info` attributes

### Related Issues

- Resolves 

Existing scripts using off-design will need to be updated to use the new interface

The following variable names have been changed:
`Mission.Design.FUEL_MASS` -> `Mission.Summary.FUEL_MASS`
`Mission.Design.FUEL_MASS_REQUIRED` -> `Mission.Summary.FUEL_MASS_REQUIRED`
`Aircraft.Design.OPERATING_MASS` -> `Mission.Summary.OPERATING_MASS`
`Mission.Design.ZERO_FUEL_MASS` -> `Mission.Summary.ZERO_FUEL_MASS`

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -1,19 +1,24 @@
...
 import dymos as dm
 import openmdao.api as om
 from dymos.utils.misc import _unspecified
+from dymos.utils.misc import _unspecified
 from openmdao.utils.mpi import MPI
 
 from aviary.core.post_mission_group import PostMissionGroup
 from aviary.core.pre_mission_group import PreMissionGroup
 from aviary.interface.utils import set_warning_format
+from aviary.core.pre_mission_group import PreMissionGroup
```
Kenneth-T-Moore: These imports are repeated.

EDIT: Lots of repeated imports in this file. Maybe Ruff messed something up?

aviary/core/aviary_group.py
```diff
@@ -1389,7 +1395,9 @@ def _add_subsystem_guesses(self, phase_name, phase, target_prob, parent_prefix):
             The phase object for which the subsystem guesses are being added.
         """
         # Get all subsystems associated with the phase
-        all_subsystems = self.get_all_subsystems(self.phase_info[phase_name]['external_subsystems'])
+        all_subsystems = self.get_all_subsystems(
```
Kenneth-T-Moore, line 1485: Maybe this should be a wholesale terminology change in Aviary, phase_info -> mission_info.

aviary/interface/methods_for_level2.py
```diff
@@ -83,7 +75,7 @@ def load_inputs(
         phase_info=None,
         engine_builders=None,
         problem_configurator=None,
-        meta_data=BaseMetaData,
+        meta_data=BaseMetaData.copy(),
```
Kenneth-T-Moore: See the multimission PR for how this has changed. We definitely want None for the argument here.

aviary/interface/methods_for_level2.py
```diff
@@ -115,9 +107,11 @@ def load_inputs(
             verbosity=verbosity,
         )
 
-        # When there is only 1 aircraft model/mission, preserve old behavior.
-        self.phase_info = self.model.phase_info
+        self.meta_data = meta_data
```
Kenneth-T-Moore: Yeah, this stuff all changes too, moving to the Aviary_Group.

aviary/examples/run_off_design_example.py
```diff
@@ -39,13 +41,15 @@
 prob.setup()
 prob.set_initial_guesses()
 prob.run_aviary_problem()
-prob.save_sizing_to_json()
+prob.save_sizing_results()
```
Kenneth-T-Moore: I guess we don't need to save the results anymore,

aviary/interface/methods_for_level2.py
```diff
@@ -1312,23 +1432,23 @@ def _save_to_csv_file(self, filename):
                 writer.writerow({'name': name, 'value': value, 'units': units})
 
 
-def _read_sizing_json(aviary_problem, json_filename):
+def _read_sizing_json(json_filename, meta_data=BaseMetaData, verbosity=Verbosity.BRIEF):
```
Kenneth-T-Moore: This is an internal private function, so we can just make meta_data a required arg instead of an optional kwarg.

aviary/core/aviary_group.py
cmbenne3: I quite like this idea... I think it would be better for less experienced users who have no familiarity with dymos. I do like  having a distinction between the input dictionary, and the copy of that data that is held within the aviary problem, particularly if we need to chunk it into 3 parts pre, mission and post I think it is important we have a different name.

aviary/interface/methods_for_level2.py
```diff
@@ -952,6 +944,130 @@ def run_payload_range(self, verbosity=None):
...
+            verbosity = Verbosity(verbosity)
+        else:
+            verbosity = self.verbosity  # defaults to BRIEF
+
+        # accept str for problem type
+        problem_type = ProblemType(problem_type)
+        if problem_type is ProblemType.SIZING:
+            raise UserWarning('Off-design missions cannot be SIZING missions.')
+
+        off_design_prob = AviaryProblem()
```
cmbenne3: If a user has complex external subsystems etc in their AviaryProblem would it be better to instead take a complete copy of the AviaryProblem (which would presumably contain all of the necessary subsystems and connections in it) and then change the necessary inputs for the off-design?

aviary/interface/methods_for_level2.py
```diff
@@ -952,6 +944,130 @@ def run_payload_range(self, verbosity=None):
...
+        off_design_prob.add_driver(optimizer, verbosity=verbosity)
+        off_design_prob.add_design_variables(verbosity=verbosity)
+        off_design_prob.add_objective(verbosity=verbosity)
+        off_design_prob.setup(verbosity=verbosity)
+        off_design_prob.set_initial_guesses(verbosity=verbosity)
+        off_design_prob.run_aviary_problem(verbosity=verbosity)
+
+        return off_design_prob
+
     def alternate_mission(
```
cmbenne3: And I think we can delete these functions once it's all tested?

aviary/interface/methods_for_level2.py
```diff
@@ -1223,7 +1339,13 @@ def save_sizing_to_json(self, json_filename='sizing_problem.json'):
             Assumed to contain aviary_inputs and Mission.Summary.GROSS_MASS
         json_filename : string
             User specified name and relative path of json file to save the data into.
+        save_to_reports : bool
+            Flag that sets where the results are saved - if True, the file is saved in the OpenMDAO
+            reports directory. If False, the file is saved to the current working directory.
         """
+        if save_to_reports:
+            reports_folder = Path(self.get_reports_dir(force=True))
+            json_filename = reports_folder / json_filename
```
cmbenne3: This is a much better place to put this file! Thanks!

aviary/interface/methods_for_level2.py
```diff
@@ -1312,23 +1432,23 @@ def _save_to_csv_file(self, filename):
...
 
     Parameters
     ----------
-    aviary_problem: OpenMDAO Aviary Problem
-        Aviary problem object optimized for the aircraft design/sizing mission.
-        Assumed to contain aviary_inputs and Mission.Summary.GROSS_MASS
-    json_filename:   string
-        User specified name and relative path of json file to save the data into
+    json_filename: str, Path
+        json file to save the data into
```
cmbenne3: json file to read data from

cmbenne3:
> In GAST, there are a few input variables related to off design. Examples:
> 
> ARNGE(3) = 900., ! Off Design Altitude, Short Range ARNGE(4) = 0., ! Off Design Altitude, Long Range OFALT=0., ! off design mission altitude in feet (0.0) OFEM=.78, ! off design specified mission Mach number
> 
> Should they be added to Aviary?

@xjjiang - I am not super familiar with GASP, but from this information alone I imagine you can get the equivalent behavior by sending an altered phase_info dictionary to run the off design mission in Aviary, so I don't think it's necessary to add these parameters. If you let me know more details about what those parameters do in GASP I can run some tests to see if I can replicate that behavior without adding new variables.

aviary/core/aviary_group.py
jkirk5: I think it must be git merge messing up again. I've noticed sometimes if you try and "combine" blocks of changes it just duplicates everything

aviary/interface/methods_for_level2.py
jkirk5: I tested and AviaryProblems can't be deepcopied. It's likely possible to manually copy over all of the attributes needed to recreate everything we'd need, but since external subsystems and other advanced options don't work right now in off-design anyway I'll suggest that change as out-of-scope for this PR and done as a follow-up

aviary/interface/methods_for_level2.py
```diff
@@ -738,254 +722,199 @@ def run_aviary_problem(
...
+
+        if problem_type is ProblemType.ALTERNATE:
+            # Set mission range, aviary will calculate required fuel
+            if mission_range is None:
+                if verbosity >= Verbosity.VERBOSE:
+                    warnings.warn(
+                        'Alternate problem type requested with no specified range. Using design '
+                        'mission range for the off-design mission.'
+                    )
+                design_range = self.get_val(Mission.Summary.RANGE, units='NM')[0]
```
cmbenne3: we could streamline this by calling this mission_range and then you only have to have the phase_info['post_mission']['target_range'] = (mission_range, 'nmi',) once and we can remove 3 lines of code from the repo...

aviary/interface/methods_for_level2.py
```diff
@@ -738,254 +722,199 @@ def run_aviary_problem(
...
+        if problem_type is ProblemType.ALTERNATE:
+            # Set mission range, aviary will calculate required fuel
+            if mission_range is None:
+                if verbosity >= Verbosity.VERBOSE:
                     warnings.warn(
-                        'One or both of the fallout missions did not run successfully; payload/range diagram was not generated.'
+                        'Alternate problem type requested with no specified range. Using design '
+                        'mission range for the off-design mission.'
                     )
+                design_range = self.get_val(Mission.Summary.RANGE, units='NM')[0]
```
cmbenne3: as above can streamline this by naming this mission_range not design_range, and could even use the value of mission_range from above without the extra get_val

aviary/interface/methods_for_level2.py
```diff
@@ -738,254 +722,199 @@ def run_aviary_problem(
...
-                    'The sizing problem has not run successfully; therefore, the payload/range analysis will not be run.'
+        elif problem_type is ProblemType.FALLOUT:
+            # Set mission fuel and calculate gross weight, aviary will calculate range
+            if mission_gross_mass is None:
+                if verbosity >= Verbosity.VERBOSE:
+                    warnings.warn(
+                        'Fallout problem type requested with no specified gross mass. Using design '
+                        'takeoff gross mass for the off-design mission.'
+                    )
+                off_design_prob.aviary_inputs.set_val(
```
cmbenne3: perhaps we can be consistent in code structure to the alternate case where we name a variable in memory and then set it rather than all in one line.

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
+            range_2 = float(self.get_val(Mission.Summary.RANGE)[0])
+            gross_mass = float(self.get_val(Mission.Summary.GROSS_MASS)[0])
+            operating_mass = float(self.get_val(Mission.Summary.OPERATING_MASS)[0])
+            fuel_capacity = float(self.get_val(Aircraft.Fuel.TOTAL_CAPACITY)[0])
+            max_payload = float(self.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS)[0])
+
+            fuel_2 = self.get_val(Mission.Summary.FUEL_BURNED)[0]
+
+            # An aircraft may be designed with fuel tank capacity that, if filled, would exceed
+            # MTOW. In this scenario, Economic Range and Ferry Range are the same, and the point
```
cmbenne3: I would prefer 'max fuel + payload' to describe point 3 rather than 'economic' (since usually people use economic to refer to a max payload shorter mission, or the mission that the aircraft is most often expected to fly during it's lifetime (neither of which is going to usually be the max fuel + payload point)...
Unfortunately that naming is clunky.

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
+            # only needs to be run once.
+            if operating_mass + fuel_capacity < gross_mass:
+                # Point 3 (Economic Mission): max fuel and remaining payload capacity
+
+                economic_mission_total_payload = gross_mass - operating_mass - fuel_capacity
```
cmbenne3, line 973: perhaps we just call this payload_3 to be consistent with payload_1 and payload_2 above

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
-        """
-        # `self.verbosity` is "true" verbosity for entire run. `verbosity` is verbosity
-        # override for just this method
-        if verbosity is not None:
-            # compatibility with being passed int for verbosity
-            verbosity = Verbosity(verbosity)
-        else:
-            verbosity = self.verbosity  # defaults to BRIEF
+                range_3 = float(economic_range_prob.get_val(Mission.Summary.RANGE))
+                fuel_3 = self.get_val(Mission.Summary.FUEL_BURNED)[0]
```
cmbenne3: I think this should be economic_range_prob.get_val()? Won't self.get_val retrieve the value from the sizing mission?

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
-                        'Incomplete PAX numbers for FLOPS fallout - assume same as design'
-                    )
-                num_first = self.aviary_inputs.get_val(Aircraft.CrewPayload.Design.NUM_FIRST_CLASS)
-                num_business = self.aviary_inputs.get_val(
-                    Aircraft.CrewPayload.Design.NUM_BUSINESS_CLASS
+                prob_3_skip = False
+            else:
+                # only fill fuel until TOGW reached
+                prob_3_skip = True
+                fuel_capacity = gross_mass - operating_mass
```
cmbenne3: strictly speaking this should be called something like ferry_mission_fuel_load, but I understand why you have called it capacity so that it flows nicely into the calculation below!

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
+
+            # Point 4 (Ferry Range): maximum fuel and 0 payload
+            ferry_range_payload = operating_mass + fuel_capacity
```
cmbenne3: This should be named ferry_mission_gross_mass. The ferry_mission_payload = 0 (well in our case actually it's one passenger until that bug gets fixed)

aviary/interface/methods_for_level2.py
```diff
@@ -995,224 +924,169 @@ def alternate_mission(
...
+                wing_cargo=0,
+                misc_cargo=0,
+                cargo_mass=0,
+                mission_gross_mass=ferry_range_payload,
+                verbosity=verbosity,
+            )
+
+            payload_4 = float(ferry_range_prob.get_val(Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS))
+            range_4 = float(ferry_range_prob.get_val(Mission.Summary.RANGE))
+            fuel_4 = self.get_val(Mission.Summary.FUEL_BURNED)[0]
```
cmbenne3: I think this should be ferry_mission_prob.get_val(0 not self.get_val?

aviary/interface/methods_for_level2.py
jkirk5: Good call!

aviary/validation_cases/benchmark_tests/test_bench_off_design.py
```diff
@@ -0,0 +1,289 @@
...
+        prob_var_list = [
+            Mission.Design.RANGE,
+            Mission.Summary.RANGE,
+            Mission.Summary.FUEL_MASS,
+            Mission.Summary.TOTAL_FUEL_MASS,
+            Mission.Summary.OPERATING_MASS,
+            Aircraft.CrewPayload.CARGO_MASS,
+            Aircraft.CrewPayload.TOTAL_PAYLOAD_MASS,
+            Mission.Design.GROSS_MASS,
+            Mission.Summary.GROSS_MASS,
```
cmbenne3, line 62: I think we should add aircraft:crew_and_payload:passenger_payload_mass and aircraft:design:empty_mass to this list for completeness.

aviary/interface/methods_for_level2.py
jkirk5: I used the point names that I was taught at Georgia Tech, but yeah there's no standard. In fact I picked "design range" as my own name, GT uses "harmonic range" which is terrible in my opinion... I think "ferry" and "design" are pretty explanatory, and I'm open to changing "economic" to a better option

aviary/interface/methods_for_level2.py
jkirk5: Yeah that's a copy-paste error on my part. Good catch

aviary/validation_cases/benchmark_tests/test_bench_off_design.py
```diff
@@ -0,0 +1,289 @@
...
+            mission_gross_mass=150_000,
+            num_first_class=1,
+            num_business=5,
+            num_tourist=75,
+        )
+        assert_near_equal(
+            prob_fallout.get_val(Mission.Design.RANGE),
+            self.prob.get_val(Mission.Design.RANGE),
+            tolerance=1e-12,
+        )
```
cmbenne3, line 106: I think we should check empty_mass here to make sure it hasn't changed from the sizing mission unexpectedly. Same for alternate mission below.

aviary/interface/methods_for_level2.py
cmbenne3: the only shortening I could sensibly come up with was MTOW_MFW_ (max takeoff, max fuel point). It's a lot of acronyms so maybe just defining in a comment what 'economic' means and sticking with that works! If GT uses that terminology then I'm good with it. I haven't found anything better from a search online.

jkirk5: Maybe "max economic range" is even better, referring to the maximum range the aircraft can fly while fully loaded. There are many economic missions, but only one at this maximum range

aviary/examples/run_level3_example.py
```diff
@@ -483,37 +484,39 @@ def initialize(self):
...
-        'parallel_deriv_color',
-        'cache_linear_solution',
-        'units',
-        'min',
-        'max',
-    ],
-)
+
+# Uncomment these lines to get printouts of every variable in the openmdao model
+# prob.model.list_vars(units=True, print_arrays=True)
```
Kenneth-T-Moore, line 522: In the latest release of OpenMDAO, I changed this method so that it lists all des_var and con_var options by default.

aviary/interface/methods_for_level2.py
```diff
@@ -1638,88 +1490,179 @@ def fallout_mission(
...
-                    Aircraft.CrewPayload.Design.NUM_BUSINESS_CLASS
+        # Off-design missions do not currently work for GASP masses or missions.
+        mass_method = self.model.aviary_inputs.get_val(Settings.MASS_METHOD)
+        equations_of_motion = self.model.aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
+        if (
+            mass_method == LegacyCode.FLOPS
+            and equations_of_motion is EquationsOfMotion.HEIGHT_ENERGY
+        ):
+            # make a copy of the phase_info to avoid modifying the original.
+            phase_info = self.model.mission_info
```
Kenneth-T-Moore: Probably should add a `.copy()` on each of these.

aviary/core/aviary_group.py
```diff
@@ -1023,13 +1022,15 @@ def add_post_mission_systems(self, verbosity=None):
 
         if not ignore_capacity_constraint:
             self.add_constraint(
-                Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, ref=1000, units='lbm'
+                Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, ref=100, units='lbm'
```
cmbenne3: I think @Kenneth-T-Moore has updated this line in PR #916

aviary/core/aviary_group.py
```diff
@@ -1337,7 +1338,8 @@ def add_design_variables(self, problem_type: ProblemType = None, verbosity=None)
 
             elif problem_type is ProblemType.FALLOUT:
                 # fixed vehicle gross mass aviary finds optimal trajectory and maximum range
-                print('No design variables for Fallout missions')
+                if verbosity >= Verbosity.VERBOSE:
+                    print('No design variables for Fallout missions')
```
cmbenne3: I have found this printout confusing so good to hide under a higher verbosity level. There are design variables for the optimization problem, we just don't add any that change the loading of the aircraft (fuel, gross weight etc). The Aviary problem still contains the design variables necessary for solving the trajectory?
I think we should either remove it or update this print to be 'No additional aircraft design variables for fallout mission.'
If we keep it then we should add printouts for the design, alternate and multi missions stating which variables HAVE been added as part of this method for consistency.

aviary/docs/examples/off_design_example.ipynb
```diff
@@ -25,37 +25,12 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The methods used to run the off-design missions accept the mission parameters `payload_mass`, `mission_range`, `mission_mass`, and `phase_info` as inputs depending on the type of mission being run.\n",
+    "The methods used to run the off-design missions accept a variety of mission parameters including `phase_info`, some mission configuration variables such as `equations_of_motion`, and inputs that depend on the type of mission being run such as payload, passenger count, and range.\n",
```
cmbenne3: Need to update 'phase_info' to 'mission_info' in this doc?

aviary/docs/examples/off_design_example.ipynb
```diff
@@ -25,37 +25,12 @@
...
-    "Summary Range = 3375.00699084\n",
-    "Design Gross mass = 175871.01115631\n",
-    "Summary Gross mass = 175871.01115631\n",
-    "\n",
-    "## Alternate Results\n",
-    "Summary Range = 3375.\n",
-    "Design Gross mass = 175871.01115631\n",
-    "Summary Gross mass = 175869.64047452\n",
     "\n",
+    "The expected result is that variables that are specifically given, such a new payload, passenger count, or range, is reflected in the results of the off-design mission. Variables that should not change, such as takeoff gross mass, or `design` values, are also printed to demonstrate that they remain unchanged. This demonstrates that the aircraft was not resized and instead just a new mission was run.\n",
```
cmbenne3: Suggested update to correct a -> as is -> are and small wording change for clarity:
The expected result is that specified variables, such as new payload, passenger count, or range, are reflected in the results of the off-design mission. Variables that should not change, such as takeoff gross mass, or `design` values, are also printed to demonstrate that they remain unchanged. This proves that the aircraft was not resized or redesigned and was just flown on a new mission.\n"

aviary/docs/examples/off_design_example.ipynb
```diff
@@ -25,37 +25,12 @@
...
     "\n",
-    "In these results, `design` values represent the values used to size the aircraft, while the `summary` values represent the values used in the specific mission that is run.\n",
-    "We can see that the results are almost identical as expected."
+    "When looking at the specific variables that were selected for printing in this example, `design` values represent the values used to size the aircraft, while the `summary` values represent the values used in the specific mission that is run."
```
cmbenne3: is -> was for consistency of tense.

"specific mission that was run."

aviary/docs/user_guide/images/payload-range-envelope.png
cmbenne3, line 1: @ehariton - you might want to update the image for your SciTech presentation as this PR will change it in the docs by the time you give the presentation in January!

aviary/examples/run_off_design_example.py
```diff
@@ -11,16 +11,23 @@
 
 import aviary.api as av
 
+# We will size the aircraft in this example for a longer design range than specified in the default
+# phase_info
 phase_info = av.default_height_energy_phase_info
+phase_info['post_mission']['target_range'] = (2500.0, 'nmi')
```
cmbenne3, line 17: should we update phase_info -> mission_info, or is the phase info the input dictionary containing, pre mission, mission and post mission details?

aviary/examples/run_off_design_example.py
```diff
@@ -32,47 +39,46 @@
 # Detail which variables the optimizer can control
 prob.add_objective()
 prob.setup()
+print('Running Design Mission')
 prob.run_aviary_problem()
-prob.save_sizing_to_json()
 
 # Fallout Mission
-prob_fallout = prob.fallout_mission()
+print('Running fixed-mass, varying range off-design problem')
```
cmbenne3, line 46: Perhaps we should add 'Fallout' into this print e.g.:
>print('Running Fallout: fixed-mass, varying range off design problem')

aviary/examples/run_off_design_example.py
```diff
@@ -32,47 +39,46 @@
...
+prob_fallout = prob.run_off_design_mission(problem_type='fallout', mission_gross_mass=115000)
 
 # Alternate Mission
-prob_alternate = prob.alternate_mission()
+print('Running fixed-range, varying fuel off-design problem')
```
cmbenne3, line 50: Perhaps we should add 'Alternate' into this print e.g.:

>print('Running Alternate: fixed-range, varying fuel off design problem')

aviary/interface/methods_for_level2.py
```diff
@@ -134,14 +134,18 @@ def load_inputs(
...
-        self.phase_info = self.model.phase_info
         self.aviary_inputs = aviary_inputs
         self.verbosity = verbosity
         if self.problem_type is None:
             # if there are multiple load_inputs() calls, only the problem type from the first aviary_values is used
             self.problem_type = aviary_inputs.get_val(Settings.PROBLEM_TYPE)
 
+        # TODO try and find a better solution than a new custom flag - the issue is multimission
+        #      problems don't have a consistent variable path to check the inputs later on
+        # BUG you can't specify generating payload-range diagram after load_inputs
```
cmbenne3: Why is this a 'bug'? Should we just move this code to check_and_preprocess_inuts?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+            is chosen.
+        name : str, optional
+            Name of the off-design problem. Defaults to "{original problem name}_off_design".
+        fill_cargo : bool, optional
+            Adds a design variable to vary cargo mass to exactly fill the aircraft to design
+            takeoff gross weight. Useful for cases where precise cargo mass required is not known,
+            or when operating mass can change between missions. Defaults to False. Cannot be used at
+            the same time as fill_fuel.
+        fill_fuel : bool, optional
+            Adds takeoff gross mass as a design variable. Useful for cases when operating mass can
```
cmbenne3, line 1328: it's a little confusing that the fill_fuel flag adds takeoff mass as a design variable, it would be nice if it more directly altered the fuel load. I understand that this isn't easy with our current implementation, and I think mathematically it equates to the same thing!

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
         else:
             verbosity = self.verbosity  # defaults to BRIEF
 
-        # TODO: these self.aviary_inputs methods will need to be updated
-        mass_method = self.aviary_inputs.get_val(Settings.MASS_METHOD)
-        equations_of_motion = self.aviary_inputs.get_val(Settings.EQUATIONS_OF_MOTION)
+        # accept str for problem type
+        problem_type = ProblemType(problem_type)
+        if problem_type is ProblemType.SIZING:
+            raise UserWarning('Off-design missions cannot be SIZING missions.')
```
cmbenne3, line 1350: Does a UserWarning throw an error and stop the function from executing?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+            inputs.set_val(Settings.PROBLEM_TYPE, problem_type)
+        if equations_of_motion is not None:
+            inputs.set_val(Settings.EQUATIONS_OF_MOTION, equations_of_motion)
+
+        if problem_configurator is not None:
+            off_design_prob.model.configurator = problem_configurator
+
+        if phase_info is None:
+            # model phase_info only contains mission information, recreate the whole thing here
+            phase_info = self.model.mission_info.copy()
```
cmbenne3, line 1378: are these ok with shallow copy not deep copy?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+
+        # Update inputs that are specific to problem type
+        # Some Alternate problem changes had to happen before load_inputs, all fallout problem
+        # changes must come after load_inputs
+        if problem_type is ProblemType.ALTERNATE:
+            off_design_prob.aviary_inputs.set_val(Mission.Summary.RANGE, mission_range, units='NM')
+            # set initial guess for Mission.Summary.GROSS_MASS to help optimizer with new design
+            # variable bounds.
+            if mission_gross_mass is None:
+                mission_gross_mass = off_design_prob.aviary_inputs.get_val(
```
cmbenne3, line 1438: Why can't this happen before load inputs?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+        off_design_prob.check_and_preprocess_inputs(verbosity=verbosity)
+        off_design_prob.add_pre_mission_systems(verbosity=verbosity)
+        off_design_prob.add_phases(verbosity=verbosity)
+        off_design_prob.add_post_mission_systems(verbosity=verbosity)
+        off_design_prob.link_phases(verbosity=verbosity)
+        if optimizer is None:
+            try:
+                optimizer = self.driver.options['optimizer']
+            except KeyError:
+                optimizer = None
```
cmbenne3, line 1468: Should we add a printout error if optimizer = None, or does OpenMDAO do that automatically in add_driver?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+                    else:
+                        control_var = Aircraft.CrewPayload.WING_CARGO
+                        val = wing_cargo
+                        tol = 1.1
+                else:
+                    control_var = Aircraft.CrewPayload.MISC_CARGO
+                    val = misc_cargo
+                    tol = 1.1
+            else:
+                control_var = Aircraft.CrewPayload.CARGO_MASS
```
cmbenne3: In this case I guess we will have a mismatch where cargo_mass != misc_cargo + wing_cargo. Presumably Aviary is happy, although a printout of the mass breakdown will look funky?

aviary/interface/methods_for_level2.py
```diff
@@ -1237,478 +1240,514 @@ def run_aviary_problem(
...
+            phase_info['post_mission'] = self.model.post_mission_info.copy()
+            # This checks if the 'cruise' phase exists, then automatically extends duration bounds
+            # of the cruise stage to allow for the longer economic and ferry missions.
+            if phase_info['cruise']:
+                min_duration = phase_info['cruise']['user_options']['time_duration_bounds'][0][0]
+                max_duration = phase_info['cruise']['user_options']['time_duration_bounds'][0][1]
+                cruise_units = phase_info['cruise']['user_options']['time_duration_bounds'][1]
+
+                phase_info['cruise']['user_options'].update(
+                    {'time_duration_bounds': ((min_duration, 2.5 * max_duration), cruise_units)}
```
cmbenne3: No change required - just a note that 2.5 is arbitrary, but I think we have to just make a guess. For an aircraft with a huge fuel capacity and huge cargo capacity 2.5x the 'design' cruise duration might not be enough! In this case the user should be able to use the dashboard to figure this out and provide a custom phase info to generate their own payload-range.

aviary/interface/methods_for_level2.py
```diff
@@ -1777,7 +1826,7 @@ def save_sizing_to_json(self, json_filename='sizing_problem.json'):
 
     def _add_hybrid_objective(self, phase_info):
         phases = list(phase_info.keys())
-        takeoff_mass = self.aviary_inputs.get_val(Mission.Design.GROSS_MASS, units='lbm')
+        takeoff_mass = self.model.aviary_inputs.get_val(Mission.Design.GROSS_MASS, units='lbm')
```
cmbenne3, line 1824: Is this related to off design or was it an unrelated bug that you caught?

aviary/interface/methods_for_level2.py
```diff
@@ -1872,150 +1920,88 @@ def _read_sizing_json(aviary_problem, json_filename):
...
-                # exception?
-                warnings.warn(
-                    f'FAILURE: list_num = {counter}, Input String = {inputs}, Attempted '
-                    f'to set_value({var_name}, {var_values}, {var_units})',
-                )
+                if verbosity >= Verbosity.VERBOSE:
+                    warnings.warn(
+                        f'Could not add item in json output to AviaryValues: input string = '
+                        f'{inputs}, attempted to set_value({var_name}, {var_values}, {var_units}). '
+                        'This variable was not added to the AviaryProblem.'
```
cmbenne3: 'This variable was not added to the AviaryValues' (this method no longer returns a problem object)

aviary/interface/methods_for_level2.py
```diff
@@ -1872,150 +1920,88 @@ def _read_sizing_json(aviary_problem, json_filename):
...
         else:
             # Not in the MetaData
-            warnings.warn(
-                f'Name not found in MetaData: list_num = {counter}, Input String = '
-                f'{inputs}, Attempted set_value({var_name}, {var_values}, {var_units})'
-            )
+            if verbosity >= Verbosity.VERBOSE:
+                warnings.warn(
+                    f'While reading json output, item was not found in MetaData: {inputs}. This '
+                    'variable was not added to the AviaryProblem.'
```
cmbenne3: not added to the AviaryValues

aviary/interface/reports.py
```diff
@@ -61,6 +61,15 @@ def register_custom_reports():
         pre_or_post='post',
     )
 
+    register_report(
+        name='sizing_results',
+        func=sizing_results,
+        desc='Generates an output file of sized aircraft results',
```
cmbenne3: 'Generates an output JSON file of the sized aircraft inputs + sized design gross mass result'

aviary/interface/reports.py
```diff
@@ -112,6 +121,23 @@ def run_status(prob):
         print(file=f)  # avoid 'no newline at end of file' message
 
 
+def sizing_results(prob):
+    """
+    Creates a JSON file that contains the outputs from a sizing problem. If the ProblemType run was
```
cmbenne3: Technically it's the inputs for the sizing problem + the updated design gross mass output

aviary/models/aircraft/test_aircraft/aircraft_for_bench_GwFm.csv
```diff
@@ -6,9 +6,9 @@ aircraft:controls:cockpit_control_mass_scaler,1,unitless
 aircraft:controls:control_mass_increment,0,lbm
 aircraft:controls:stability_augmentation_system_mass,0,lbm
 aircraft:controls:stability_augmentation_system_mass_scaler,1,unitless
-aircraft:crew_and_payload:cargo_mass,10040,lbm
-#aircraft:crew_and_payload:design:cargo_mass,0,lbm
-#aircraft:crew_and_payload:design:max_cargo_mass,0,lbm
+aircraft:crew_and_payload:cargo_mass,0,lbm
+aircraft:crew_and_payload:design:cargo_mass,0,lbm
+aircraft:crew_and_payload:design:max_cargo_mass,10040,lbm
```
cmbenne3, line 11: Why was this change required?

aviary/subsystems/mass/flops_based/cargo.py
```diff
@@ -10,11 +10,38 @@
 from aviary.variable_info.variables import Aircraft
 
 
+class PayloadGroup(om.Group):
```
cmbenne3, line 13: Perhaps this should be in a separate file aviary/subsystems/mass/flops_based/payload.py or similar?

aviary/subsystems/mass/flops_based/cargo.py
```diff
@@ -57,17 +77,10 @@ def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
         outputs[Aircraft.CrewPayload.BAGGAGE_MASS] = baggage_mass_per_passenger * passenger_count
 
         outputs[Aircraft.CrewPayload.PASSENGER_PAYLOAD_MASS] = (
-            outputs[Aircraft.CrewPayload.PASSENGER_MASS]
-            + outputs[Aircraft.CrewPayload.BAGGAGE_MASS]
+            mass_per_passenger * passenger_count + baggage_mass_per_passenger * passenger_count
```
cmbenne3, line 80: Interesting - I agree with this change and the old implementation looked incorrect.

aviary/subsystems/mass/flops_based/mass_premission.py
```diff
@@ -71,7 +71,7 @@ def initialize(self):
     def setup(self):
         alt_mass = self.options[Aircraft.Design.USE_ALT_MASS]
 
-        self.add_subsystem('cargo', CargoMass(), promotes_inputs=['*'], promotes_outputs=['*'])
+        self.add_subsystem('cargo', PayloadGroup(), promotes_inputs=['*'], promotes_outputs=['*'])
```
cmbenne3, line 74: rename the subsystem 'payload'?

aviary/utils/fortran_to_aviary.py
```diff
@@ -584,6 +584,17 @@ def update_gasp_options(vehicle_data, verbosity=Verbosity.BRIEF):
...
+    ## Cargo ##
+    if (
+        Aircraft.CrewPayload.CARGO_MASS in input_values
+        and Aircraft.CrewPayload.Design.MAX_CARGO_MASS not in input_values
+    ):
+        # user has set cargo only: assume intention to set max only for backwards compatibility.
+        max_cargo, units = input_values.get_item(Aircraft.CrewPayload.Design.MAX_CARGO_MASS)
+        input_values.set_val(Aircraft.CrewPayload.Design.MAX_CARGO_MASS, max_cargo, units)
+        input_values.set_val(Aircraft.CrewPayload.CARGO_MASS, 0, units)
+        input_values.set_val(Aircraft.CrewPayload.Design.MAX_CARGO_MASS, 0, units)
```
cmbenne3: I think there is an extra line here which is incorrect - either we should be setting Design.MAX_CARGO_MASS to max_cargo or 0, but not max_cargo and overwriting it?

aviary/utils/preprocessors.py
```diff
@@ -324,43 +332,30 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
...
-                cargo = des_cargo = 0
-                if verbosity >= Verbosity.BRIEF:  # BRIEF, VERBOSE, DEBUG
-                    warnings.warn(
-                        'As-flown cargo mass was specified but design cargo mass and '
-                        'max cargo mass were not. To maintain backwards-compatibility '
-                        f'with converted GASP files, setting max cargo mass to {cargo} '
-                        'and maximum and design cargo masses to zero.'
-                    )
+                # user has set cargo only: assume design and max cargo is equal to flown cargo
+                des_cargo = max_cargo = cargo
```
cmbenne3, line 358: Intentional removal of backwards compatibility message?

aviary/utils/preprocessors.py
```diff
@@ -398,12 +391,12 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
...
+                    'No CARGO variables detected, assume CARGO_MASS, Design.MAX_CARGO_MASS, and '
+                    'Design.CARGO_MASS equal to 0.'
                 )
 
         # check for potential cargo errors:
-        if cargo > des_cargo:
+        if cargo > max_cargo:
             if verbosity >= Verbosity.BRIEF:  # BRIEF, VERBOSE, DEBUG:
                 warnings.warn(
                     f'As-flown cargo ({cargo}) is greater than design cargo ({des_cargo})'
```
cmbenne3: Must update warning if changing this check? Why was it necessary to change?

aviary/validation_cases/benchmark_tests/test_bench_FwFm.py
```diff
@@ -210,7 +210,7 @@ def test_bench_FwFm_SNOPT(self):
         prob = run_aviary(
             'models/aircraft/test_aircraft/aircraft_for_bench_FwFm.csv',
             self.phase_info,
-            verbosity=0,
+            verbosity=1,
```
cmbenne3: Was this update of verbosity intentional?

Kenneth-T-Moore: I like to have verbosity=1 in all `bench_*`, but I agree it is better to keep output clean in `test_*`.  Mostly, we turn these on all the time for debugging, but forget to turn them off.

aviary/docs/examples/off_design_example.ipynb
jkirk5: You still pass the off-design methods a full phase_info dict, so it's correct as-is

aviary/subsystems/mass/flops_based/mass_premission.py
jkirk5: There is a TotalPayload component that already exists, and the PayloadGroup is a mass summation group that combines that with the cargo mass component for organization.

FLOPS mass has a lot of similarly named groups, such as `FuelCapacityGroup`, `LandingMassGroup`, etc.

aviary/utils/preprocessors.py
jkirk5: Moved the GASP specific handling to fortran_to_aviary, the behavior in this file is now different and generic to any Aviary model. So if you convert a GASP model, the (imo unintuitive) defaulting is translated into the Aviary input file, and then the preprocessor doesn't need a GASP-specific option.

Basically I implemented the TODO

aviary/utils/fortran_to_aviary.py
jkirk5: the max_cargo local variable should be as-flown cargo. Must have gotten them mixed up, good catch!

aviary/examples/run_off_design_example.py
jkirk5: This is modifying the total phase_info dictionary

aviary/interface/methods_for_level2.py
jkirk5: Yes, it raises an exception and quits the program

aviary/interface/methods_for_level2.py
jkirk5: Yeah it is a current implementation issue, fuel isn't really a useful control because gross mass is actually the input that matters. This might cause a problem with all-electric vehicles, this flag doesn't make physical sense with that kind of vehicle. We'll find out once our electric example is working

aviary/interface/methods_for_level2.py
jkirk5: Turns out total cargo mass isn't actually a valid control when using FLOPS mass since it is an output, so I modified this code so that isn't possible anymore!

aviary/interface/methods_for_level2.py
jkirk5: The "bug" is that setting a new value in aviary_inputs doesn't reflect whatever gets stored in the flag here. So wherever you put it, after that step it is impossible to change unless you know about setting this internal flag.

Basically aviary values and this flag may mismatch if the user tries to change it

aviary/interface/methods_for_level2.py
jkirk5: Ken added in that copy a while ago, I'm not sure if a deepcopy is necessary or not

aviary/utils/preprocessors.py
jkirk5: this warning is actually entirely skipable now, since we already check if cargo or design cargo is greater than max cargo. Previously, there was special handling if cargo was greater than des_cargo but in Aviary we allow that to happen

aviary/subsystems/mass/flops_based/cargo.py
jkirk5: This group is so small and simple that a separate file seemed overkill to me

aviary/models/aircraft/test_aircraft/aircraft_for_bench_GwFm.csv
jkirk5: Due to moving a GASP-specific handling of these cargo variables these ended up getting flipped in the preprocessor. To keep things consistent and how Aviary expects them to behave, that special handling was moved to fortran_to_aviary, so the input files need to reflect that change instead

aviary/interface/methods_for_level2.py
jkirk5: load_inputs() overwrites some variables if they are defined beforehand. The summary range specifically gets pulled from phase_info and set there, while we want to use the user-provided value when they call this method

aviary/interface/methods_for_level2.py
jkirk5: That's the default we've always been using in Aviary, so I assume that is the case that OM handles it?

aviary/interface/methods_for_level2.py
jkirk5: At this point I have no idea... It is part of a commit where I was debugging the new off-design setup so I assume it is a bug I ran into along the way

PR #841, opened 2025-08-11
Glue BWB variables in documentation

xjjiang:
### Summary

All the variables in two BWB documentation pages are glued in this PR.

### Related Issues

- Resolves #

None

### New Dependencies

None


jkirk5:
> @jkirk5 my feeling is that `theory_guide/gasp_based_bwb.ipynb` is more like a developer guide. So, maybe I should move it under `developer_guide`.

I think it belongs under theory. It explains how the component works and compares to the original GASP, and isn't geared towards developers over users. I see the developers guide as teaching people how to modify source code, and this is more useful for actually understanding the component to use for analysis.

PR #842, opened 2025-08-11
Multimission v2

ehariton:
### Summary

Multi-mission enables a user to input one or more aircraft definitions (aviary_inputs) and a matching number of mission definitions (phase_info) to create a composite objective function who's weighting is specified by the user. In this PR we completely revamp the multi-mission capabilities of Aviary which were previously broken by an OpenMDAO release. As part of this PR, a number of examples were cleaned up / shortened for decrease verbosity and to make use of some of the new methods created in the PR.

**General Changes that could affect all users:**
1. There is not interface change for regular users. Their models are now instantiated in `prob.AviaryGroup`. Most all of the original functions contained in methods_for_level2 have been forwarded down into AviaryGroup(). This allows us to set values, link phases, and create subsystems in individually on each group without interfering with the other groups.
2. A new `build_model()` method combines calls to `add_pre_mission_systems()`, `add_phases()`,  `add_post_mission_systems()`, `link_phases()` that are often used by the user without modification. The individual supporting methods can still be accessed at level 2 if the user needs to add more detail to those inputs.
3. `setup()` will now automatically call `set_initial_guesses()` This will allow users to remove `set_initial_guesses()` from their code making things less verbose.
4. There are some new methods for objective declaration: `add_composite_objective()` and `add_composite_objective_adv()`. These methods can be used with any mission type.
5. `Mission.Summary.FINAL_MASS` and `Mission.Summary.FINAL_TIME` were added to the variable meta data to enable users to reference the last point in the mission phase without using `loc=-1` calls. This allows users easy access to information from the last mission phase run in dymos. If you are adding custom ExecComp() at the end of your mission i.e. to do taxi-to-gate without a dymos phase, these values will not reflect that and you will need to update them for your specific implementation.
6. AviaryProblem can now accept `meta_data` as an input. This allows the user to first combine all different meta data they have using `av.merge_hierarchies` and then to input that into the AviaryProblem.

**Multi-mission specific changes**
1. When setting `ProbelmType = MultiMission` separate aviary problems will now be instantiated in `prob.model.AviaryGroup0`, `prob.model.AviaryGroup1` etc.
2. Setting problem_type = multimission supersedes and ignores problem_type information written in the .csv . We did not want these problem types conflicting with eachother.
3. It is not possible to run range-based optimizations with multi-mission at the moment (follow-on work).
4. Some new methods will help with setup multimission problems including: `add_aviary_group()`, `promote_inputs()`, `add_design_var_default()`, and `set_design_range()`. 
8. Use `add_composite_objective()` and `add_composite_objective_adv()` to quickly specify composite objectives such as fuel from mission 1 + payload capacity from mission 2.
9. The reports were updated to accommodate creating one report per aircraft + mission definition pair.
10. `add_aviary_group()` will not accept .csv files. If the user needs to start from an aircraft defined by a .csv, we recommend examining `large_single_aisle_2_FLOPS_data.py` for an example of how the .csv needs to be converted. An automated converter will be supplied in the future. You may be able to use `parse_inputs()` to achieve this before an official converter is supplied by the dev team.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -70,23 +70,15 @@ def __init__(self, verbosity=None, **kwargs):
...
-        )
-        self.options.declare(
-            'aviary_metadata', types=dict, desc='metadata dictionary of the full aviary problem.'
-        )
-        self.options.declare('phase_info', types=dict, desc='phase-specific settings.')
-        self.builder = []
+        self.aviary_inputs = None
+        self.meta_data = None
+        self.phase_info = None
+        # self.core_subsystems = {}
```
Kenneth-T-Moore: remove commented-out line

aviary/interface/methods_for_level2.py
```diff
@@ -73,18 +76,26 @@ def __init__(self, verbosity=None, **kwargs):
...
     def load_inputs(
         self,
         aircraft_data,
         phase_info=None,
         engine_builders=None,
         problem_configurator=None,
-        meta_data=BaseMetaData,
+        meta_data=None,
         verbosity=None,
+        check=False,
```
Kenneth-T-Moore: check should default to True on the user-facing load_inputs, so that the user doesn't ever have to type it. I think the case where they will ever need to call `check_and_preprocess_inputs` is pretty rare.

aviary/docs/getting_started/onboarding_level3.ipynb
```diff
@@ -67,14 +67,16 @@
     "In `aviary/validation_cases/benchmark_tests` folder, there is an N3CC aircraft full mission benchmark test `test_FLOPS_based_sizing_N3CC.py`. Now, we will show how to create an N3CC example in level3. The key is that we follow the same steps:\n",
     "- {glue:md}`__init__()`\n",
     "- {glue:md}`load_inputs()`\n",
-    "- {glue:md}`check_and_preprocess_inputs()` (optional)\n",
+    "- {glue:md}`prob.check_and_preprocess_inputs()`\n",
```
Kenneth-T-Moore: New additiona have a `prob.` in front of them, which isn't consistent with the rest.

aviary/interface/methods_for_level2.py
ehariton: I'm planning on leaving this as check=False because 1) we don't want to brick current users setups that are already working 2) the whole check system will be deprecated in the future as all users will be required to load .csv inputs with a separate function.

aviary/examples/external_subsystems/OAS_mass/run_OAS_wing_mass_example.py
```diff
@@ -210,7 +206,6 @@
 prob.set_val(OAS_sys + 'engine_location', np.array([25, -10.0, 0.0]), units='m')
 
 # finish setting up Aviary problem and run
-prob.set_initial_guesses()
```
ehariton, line 213: I'm actually worried that this may cause a test to fail because we are calling a set_val after setup_model(). I don't have OAS on my computer so can't check locally.

aviary/mission/flight_phase_builder.py
```diff
@@ -47,15 +47,13 @@ def declare_options(self):
         # TODO: These defaults aren't great, but need to keep things the same for now.
         defaults = {
             'mass_ref': 1e4,
-            'mass_defect_ref': 1e6,
```
ehariton, line 50: I don't remember why we needed to change these default defect refs

aviary/mission/height_energy_problem_configurator.py
```diff
@@ -313,13 +313,6 @@ def link_phases(self, aviary_group, phases, connect_directly=True):
                 phase.set_state_options(Dynamic.Vehicle.MASS, input_initial=False)
                 phase.set_state_options(Dynamic.Mission.DISTANCE, input_initial=False)
 
-        aviary_group.connect(
```
ehariton, line 316: This capability was moved into the state_output subsystem in AviaryGroup because it was needed in all cases.

aviary/validation_cases/benchmark_tests/test_off_design_GwGm.py
```diff
@@ -11,7 +11,7 @@
 
 class TwoDOFTestCase(unittest.TestCase):
     def setUp(self) -> None:
-        self.sized_mass = 174039.0
+        self.sized_mass = 171044.0
```
ehariton, line 14: Ken I'm guessing this was one of your changes?

aviary/mission/two_dof_problem_configurator.py
```diff
@@ -502,24 +502,12 @@ def link_phases(self, aviary_group, phases, connect_directly=True):
             src_indices=[-1],
         )
 
-        connect_map = {
```
ehariton, line 505: This capability was moved into AviaryGroup state_output subsystem because it was needed in all cases.

aviary/validation_cases/benchmark_tests/test_off_design_GwGm.py
Kenneth-T-Moore: Yes, the design gross mass of the baseline Gasp problem is lower now because Xun fixed a bug where one of the form factors for aero was using a default value instead of the one in the csv file.

aviary/examples/external_subsystems/OAS_mass/run_OAS_wing_mass_example.py
Kenneth-T-Moore: The OAS builder doesn't define any `get_initial_guesses`, so the order here shouldn't matter.

aviary/mission/flight_phase_builder.py
Kenneth-T-Moore: We had some convergence problems with that value for the defect ref. I've seen this happen in some other cases too, and we have changed this value on a few outstanding Aviary forks.

aviary/interface/methods_for_level2.py
```diff
@@ -138,25 +158,49 @@ def check_and_preprocess_inputs(self, verbosity=None):
...
 
             for subsystem in external_subsystems:
                 meta_data = subsystem.meta_data.copy()
                 self.meta_data = merge_meta_data([self.meta_data, meta_data])
 
-        self.model.meta_data = self.meta_data  # TODO: temporary fix
+        # Update the reference to the newly merged meta_data.
+        group.meta_data = self.meta_data
+
+    def add_aviary_group(self, name: str, aircraft: AviaryValues, mission: dict, verbosity=None):
```
kanekosh: Could you add `engine_builders` and `meta_data` arguments here and pass them to `load_inputs` so we can bring in custom meta data and engine builders?

kanekosh:
This will clean-up our multimission code quite a bit, thank you! I'm trying to adopt this change in our code, still getting some errors but will let you know how it works.

This is not directly relevant, but I think `mission:summary:gross_mass` and `mission:design:gross_mass` are a bit confusing and might benefit from renaming. In my understanding `mission:summary:gross_mass` is the actual takeoff weight for a mission, and `mission:design:gross_mass` is the design MTOW. In my opinion MTOW would belong to`aircraft:` variables rather than `mission:`. So I was wondering if the renamings like `mission:summary:gross_mass -> mission:takeoff_gross_mass` and `mission:design:gross_mass -> aircraft:design:max_takeoff_gross_mass` would make more sense.

aviary/interface/methods_for_level2.py
ehariton: `engine_builders` has been added. `meta_data` can be loaded with add AviaryProblem.

aviary/interface/methods_for_level2.py
```diff
@@ -7,11 +7,14 @@
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
+import enum
```
Kenneth-T-Moore: enum and warn_deprecation look like unused imports.

aviary/interface/methods_for_level2.py
```diff
@@ -563,17 +652,344 @@ def add_objective(self, objective_type=None, ref=None, verbosity=None):
...
+
+    def add_design_var_default(
+        self,
+        name: str,
+        lower: float = None,
+        upper: float = None,
+        units: str = None,
+        src_shape=None,
+        default_val: float = None,
+    ):  # TODO: Add Ref
```
Kenneth-T-Moore: Some of these new user_facing methods still need docstrings in the standard format with parameters/returns.

aviary/examples/multi_mission/run_multimission_example.py
```diff
@@ -1,319 +1,171 @@
 """
-authors: Jatin Soni, Eliot Aretskin
+Authors: Eliot Aretskin-Hariton, Kenneth Moore, Jutin Soni
```
jkirk5:
```suggestion
Authors: Eliot Aretskin-Hariton, Kenneth Moore, Jatin Soni
```

PR #851, opened 2025-08-19
Phase info corrections

### LINKED ISSUE #751: Throttle enforcement for default Height-Energy 'cruise' phase

The default Height-Energy phase info uses boundary constraints for the throttle enforcement rather than a path constraint. I think this needs to be modified to avoid invalid throttle settings during cruise.

Image

The throttle constraints are already correctly set as path constraints for climb and descent phases.

There doesn't seem to be the same throttle constraints for the default 2dof phase info.

_No response_

### Associated Bug Report

_No response_

cmbenne3:
### Summary

This PR removes the example_phase_info since it is duplicated by the default_height_energy phase info, and adds a small correction to the cruise phase throttle constraint to enforce a path constraint rather than a boundary constraint which makes no sense for the throttle parameter.

image

### Related Issues

- Resolves #751 

None

### New Dependencies

None


aviary/examples/run_level1_example.py
```diff
@@ -11,11 +11,13 @@
...
 """
 
+from aviary.models.missions.height_energy_default import phase_info
+
 import aviary.api as av
 
 prob = av.run_aviary(
     aircraft_data='models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv',
-    phase_info='examples/example_phase_info.py',
+    phase_info=phase_info,
```
Kenneth-T-Moore: I think this example was meant to show how aviary could find the phase info from a path string. I think 'models.missions.height_energy_default.py' would work here.

aviary/models/missions/height_energy_default.py
```diff
@@ -39,7 +39,7 @@
             'altitude_final': (34000.0, 'ft'),
             'altitude_bounds': ((23000.0, 38000.0), 'ft'),
             'altitude_polynomial_order': 3,
-            'throttle_enforcement': 'boundary_constraint',
+            'throttle_enforcement': 'path_constraint',
```
Kenneth-T-Moore, line 42: Just a note that 'boundary_constraint' isn't necessarily wrong. The thought process was that the most demanding point for the engine would be the top of climb, since the weight and drag should decrease over cruise, so constraining the upper bound on throttle at the first point in cruise is equivalent to constraining all points because the later points will never hit the boundary.

cmbenne3: I think you're absolutely right Ken and I'm sure this will be the case most of the time. I was thinking about the optimizer 'flying' the aircraft and didn't want it to be able to do something that's not physically possible (in this case push the throttle past 100 % or below 0 %). I don't believe the boundary constraints do this, and it would be possible (although not optimal) for the optimizer to choose an invalid throttle setting in the middle of the cruise phase.
In my opinion switching to a path constraint, which matches the climb and descent phases makes more sense to me. If we would like to keep the boundary constraints in there as an example of an option the user has, then I'm happy to do that, but perhaps we could add a comment explaining why using boundary constraints is 'safe' in this case.

aviary/examples/run_level1_example.py
cmbenne3: Thanks Ken - I've changed this to 'models/missions/height_energy_default.py' so users can see that they can specify a file path string here too. Provided there is a dictionary named phase_info in the python file then Aviary can find it.

Kenneth-T-Moore (approved):

cmbenne3:
> `default_height_energy` phase info is kept in a totally different location than the `example_phase_info`, which was kept in the examples folder. I think there is value in keeping the example phase info where it is and removing the `default_height_energy` instead.

From a repo organization point of view I would prefer that all default phase infos be kept in the same place (that makes more sense to me than having the height_energy_default in aviary/examples/ and the two_dof_default in aviary/models/missions/), however I do understand wanting to have everything we need to run examples in the examples folder.

ehariton (approved):

Issue #828, opened 2025-07-29
Add an Excess Fuel Output to Dashboard for FLOPS Missions


cmbenne3:
### Desired capability or behavior.

It is currently possible to execute and optimize an aircraft sizing mission where the Aircraft.Fuel.TOTAL_CAPACITY < MIssion.Design.FUEL_MASS. Legacy FLOPS was also able to do this, and had a variable 'EXSFUL' (Excess Fuel) that would show up in the output file -ve and with a warning message to the user if this condition was triggered.
I think we should add this to aviary, printing a message to the command line and adding a flag in the dashboard to show that although the optimization problem was solved the results may not be representative of a physical aircraft.

jkirk5:
We may want to enforce this via a constraint so the design fails/is infeasible when this happens (we can call a warning that specifies that as one of the reasons the run failed). If we only print out a message somewhere, the optimization will claim success and the dashboard will show the big green checkmark that everything worked, and many users may miss that

cmbenne3:
Rather than exactly replicating the feedback of design.Fuel_MASS into the Fuel.Total_capacity variable it was decided that we would translate this functionality into ignoring or enabling the excess_fuel constraint that is added to the Aviary Problem.
IFUFU is directly linked to this ignore_excess_fuel_constraint variable and gives equivalent behviour of allowing Aviary to converge a mission with negative excess fuel, but Aviary does not need to insert fuel into the fuselage to make this happen. The fuel sits in an imaginary tank and the designer must interpret the results at their own risk.

PR #854, opened 2025-08-20
Fuel Variables and Subsystems Revamp

### LINKED ISSUE #470: Improve Fuel Capacity/Volume/Density/Ratio variables

Fuel variables and calculations can be improved for usability.

'Capacity' naming is unclear - it should be made more explicit by being changed to volume or mass.
Aircraft.Fuel.Capacity_Factor should be replaced as it is not clear what value this should take (16 = 50%, 23 = 70% default, 33 = 100% of the available wing volume is occupied with fuel see table below).

This is the derivation of the FLOPS fuel equations:
image

These could be updated to use some parameters such as:
1. wing_fuel_volume_fraction, unitless
2. wing_volume, m^3
3. allowing calculation of wing_fuel_volume, m^3
4. fuel density, kg/m^3 specified as absolute rather than a ratio (remove Aircraft.Fuel.Density_Ratio)
5. wing_fuel_mass, kg

to calculate the fuel mass / volume in a way that doesn't require the user to use a parameter with a scale  of 0 - 33.

If these changes are made then the fortran_to_aviary conversion function needs to be updated to translate the FWMAX and FULDEN parameters correctly into an absolute fuel density, and a wing_fuel_volume_fraction.

Any changes should be updated in the documentation.

_No response_

### Associated Bug Report

_No response_

cmbenne3:
### Summary

This PR fixes 3 connected issues with the fuel subsystems.

Removes 2 variables, and updates fortran_to_aviary to correctly handle the new assignments:
1. Aircraft.Fuel.DENSITY_RATIO (instead user should just specify the absolute FUEL_DENSITY)
2. Aircraft.Fuel.CAPACITY_FACTOR (This variable is from legacy FLOPS and has no physical meaning. Switch to using Aircraft.Fuel.WING_FUEL_FRACTION)

Adds a new post-mission exec comp subsystem to calculate the excess fuel for the mission. (TOTAL_FUEL_CAPACITY - TOTAL_MISSION_FUEL).

Adds a new constraint for the optimization to force that excess fuel must be greater than zero.

### Related Issues

- Resolves #470 #828 

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -978,6 +979,47 @@ def add_post_mission_systems(self, verbosity=None):
             promotes_outputs=[('mass_resid', Mission.Constraints.MASS_RESIDUAL)],
         )
 
+        ecomp = om.ExecComp(
```
ehariton, line 989: I recommend changing `mission_fuel` to `total_fuel_mass`. Because mission_fuel is a misnomer, it sounds like it doesn't include the reserve fuel components.

aviary/subsystems/mass/flops_based/test/test_fuel_capacity.py
```diff
@@ -61,8 +61,8 @@ def configure(self):
             case_name,
             input_keys=[
                 Aircraft.Fuel.AUXILIARY_FUEL_CAPACITY,
-                Aircraft.Fuel.CAPACITY_FACTOR,
-                Aircraft.Fuel.DENSITY_RATIO,
+                Aircraft.Fuel.WING_FUEL_FRACTION,
+                Aircraft.Fuel.DENSITY,
```
ehariton, line 65: Are any of these tests running a check partials? You updated the core compute of fuel_capacity.py and that's typically an opportunity for partials mistakes to crop up.

aviary/subsystems/mass/flops_based/unusable_fuel.py
```diff
@@ -116,8 +116,11 @@ def compute_partials(self, inputs, J):
             0.448 * tank_count * total_capacity**-0.72 * density_ratio * scaler / GRAV_ENGLISH_LBM
         )
 
-        J[Aircraft.Fuel.UNUSABLE_FUEL_MASS, Aircraft.Fuel.DENSITY_RATIO] = (
-            (11.5 * num_eng_fact * term1 + 0.07 * wing_area + 1.6 * tank_count * term2)
+        J[Aircraft.Fuel.UNUSABLE_FUEL_MASS, Aircraft.Fuel.DENSITY] = (
```
ehariton, line 120: Are you accounting for `Aircraft.Fuel.UNUSABLE_FUEL_MASS` in your calculation for `Excess_fuel`??
Right now your comps for this are `Excess_fuel = TOTAL_FUEL_CAPACITY - TOTAL_FUEL_MASS` which doesn't account for any unusable fuel mass!!

aviary/subsystems/mass/flops_based/test/test_fuel_capacity.py
ehariton: do_validation_test() auto-runs check_partials!

aviary/core/aviary_group.py
cmbenne3: good suggestion - updated to overall_fuel to be consistent with unpromoted naming convention in mass_constraint and fuel_calc subsystems

aviary/subsystems/mass/flops_based/unusable_fuel.py
cmbenne3: Correct - good spot - I have pushed another commit to fix this.

aviary/variable_info/variable_meta_data.py
```diff
@@ -7032,6 +7013,14 @@
 #  \_____|  \___/  |_| |_| |___/  \__| |_|     \__,_| |_| |_| |_|  \__| |___/
 # ===========================================================================
 
+add_meta_data(
+    Mission.Constraints.EXCESS_FUEL_CAPACITY,
+    meta_data=_MetaData,
+    historical_name={'GASP': None, 'FLOPS': None, 'LEAPS1': None},
+    units='lbm',
+    desc='Difference between the usable fuel capacity on the aircraft and the total fuel (including reserve) required for the mission. Must be >= 0 to ensure that the aircraft has enough fuel to complete the mission',
```
ehariton: need a slight mod in here to also mention unusuable fuel mass

cmbenne3: We decided wording of _usable_ fuel capacity was enough. No further changes required.

ehariton (approved):

aviary/core/aviary_group.py
```diff
@@ -982,6 +983,49 @@ def add_post_mission_systems(self, verbosity=None):
...
+        )
+
+        # determine if the user wants the excess_fuel_capacity constraint active and if so add it to the problem
+        try:
+            # for backwards compatability check to see if variable exists, if not assume default value = False
+            ignore_capacity_constraint = self.aviary_inputs.get_val(
+                Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT, units='unitless'
+            )
+        except:
+            warnings.warn(
```
Kenneth-T-Moore: Probably want to move these warnings under the Verbosity level check so that they don't warn when quiet.

aviary/core/aviary_group.py
```diff
@@ -982,6 +983,49 @@ def add_post_mission_systems(self, verbosity=None):
...
+                val=ignore_capacity_constraint,
+                units='unitless',
+            )
+
+        if not ignore_capacity_constraint:
+            self.add_constraint(Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, units='lbm')
+        else:
+            warnings.warn(
+                'Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT = True therefore EXCESS_FUEL_CAPACITY constraint was not added to the Aviary problem. The aircraft may not have enough space for fuel, check value of Mission.Constraints.EXCESS_FUEL_CAPACITY for details.'
+            )
```
Kenneth-T-Moore, line 1027: "True therefore EXCESS_FUEL_CAPACITY " --> "True, therefore the EXCESS_FUEL_CAPACITY "
" space for fuel, check value" --> " space for fuel, so check the value"

aviary/subsystems/mass/flops_based/unusable_fuel.py
```diff
@@ -47,14 +47,14 @@ def setup_partials(self):
...
-                Aircraft.Fuel.DENSITY_RATIO,
+                Aircraft.Fuel.DENSITY,
             ],
         )
 
     def compute(self, inputs, outputs, discrete_inputs=None, discrete_outputs=None):
         tank_count = self.options[Aircraft.Fuel.NUM_TANKS]
         scaler = inputs[Aircraft.Fuel.UNUSABLE_FUEL_MASS_SCALER]
-        density_ratio = inputs[Aircraft.Fuel.DENSITY_RATIO]
+        density_ratio = inputs[Aircraft.Fuel.DENSITY] / 50.1194909
```
Kenneth-T-Moore: probably need a note on where this magic number comes from.

aviary/subsystems/mass/flops_based/fuel_capacity.py
```diff
@@ -148,20 +148,20 @@ def compute(self, inputs, outputs):
...
             span = inputs[Aircraft.Wing.SPAN]
             taper_ratio = inputs[Aircraft.Wing.TAPER_RATIO]
             thickness_to_chord = inputs[Aircraft.Wing.THICKNESS_TO_CHORD]
-
-            fuel_cap_wing = (
-                density_ratio
-                * cap_factor
+            # calculate volume of wing assuming truncated rectangular based pyramid:
+            volume_of_wing = (
+                (2 / 3)
```
Kenneth-T-Moore, line 159: I guess this 2/3 also comes from non-legacy flops?

cmbenne3: Yes - here is the derivation of that relationship. I couldn't come up with a way of describing this in a comment in the code other than "calculate volume of wing assuming truncated rectangular based pyramid". What do you think we should do here?

image

aviary/subsystems/mass/flops_based/unusable_fuel.py
cmbenne3: I have added a comment in the code.

aviary/core/aviary_group.py
cmbenne3: I have updated the wording and format of this warning

aviary/core/aviary_group.py
cmbenne3: Good idea! I've now done this.

aviary/subsystems/mass/flops_based/fuel_capacity.py
Kenneth-T-Moore: Maybe we need a place in the repository to keep theory/derivation related notes. It could be somewhere under the docs subdirectory I suppose. This is fine as a jpg.

cmbenne3: I have added a section to the theory guide for the flops_based fuel capacity component and added this png as an image. Let me know if this is what you had in mind.

cmbenne3:
@Kenneth-T-Moore @ehariton - I've made a few extra changes since you both last reviewed this:
1. Updates to theory guide documentation for fuel capacity
2. New preprocessors function to check for inconsistent user inputs and adding an assumption that if fuselage or auxiliary fuel variables are not set then they are assumed to be zero. Without this assumption if you didn't have either total_capacity or both auxiliary and fuselage fuel variables in the input file then Aviary wouldn't be able to converge the problem due to the structure of the Flops fuel capacity subsystem.
Please let me know if you have any questions/concerns!

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
```diff
@@ -62,15 +62,15 @@ aircraft:fins:area,0,ft**2
...
-aircraft:fuel:density_ratio,1,unitless
+#aircraft:fuel:auxiliary_fuel_capacity,0,lbm
+aircraft:fuel:wing_fuel_fraction,0.7752,unitless
+aircraft:fuel:density,6.7,lbm/galUS
 aircraft:fuel:fuel_system_mass_scaler,0.93202,unitless
-aircraft:fuel:fuselage_fuel_capacity,0,lbm
+#aircraft:fuel:fuselage_fuel_capacity,0,lbm
 aircraft:fuel:num_tanks,7,unitless
-aircraft:fuel:total_capacity,33136.4,lbm
+#aircraft:fuel:total_capacity,33136.4,lbm
```
Kenneth-T-Moore: Were these capacity values commented out just for testing, or was this intentional? I can't really remember if they came from the FLOPS input file or if they were computed offline, but if they are original, we probably want to preserve them and use a different phase_info for testing the corner cases.

aviary/utils/preprocessors.py
```diff
@@ -487,6 +488,86 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
     return aviary_options
 
 
+def preprocess_fuel_capacities(aviary_options: AviaryValues, meta_data=_MetaData, verbosity=None):
+    """Preprocesses the AviaryValues object to ensure the user has provided a consistent set of fuel capacity overrides."""
```
Kenneth-T-Moore: This preprocessor needs the parameters in the docstring. See the other preprocessors for example.

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
cmbenne3: Ah - yes sorry I commented them out for testing and obviously didn't reset everything - I'll undo these changes!

aviary/core/aviary_group.py
```diff
@@ -982,6 +983,52 @@ def add_post_mission_systems(self, verbosity=None):
...
+            # for backwards compatability check to see if variable exists, if not assume default value = False
+            ignore_capacity_constraint = self.aviary_inputs.get_val(
+                Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT, units='unitless'
+            )
+        except:
+            if verbosity >= Verbosity.BRIEF:
+                warnings.warn(
+                    'No value for Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT specified, assume False'
+                )
+            ignore_capacity_constraint = False
```
jkirk5:
```suggestion
            ignore_capacity_constraint = self.meta_data[Aircraft.Fuel.IGNORE_CAPACITY_CONSTRAINT]['default_value']
```

aviary/core/aviary_group.py
```diff
@@ -982,6 +983,52 @@ def add_post_mission_systems(self, verbosity=None):
...
+        try:
+            # for backwards compatability check to see if variable exists, if not assume default value = False
+            ignore_capacity_constraint = self.aviary_inputs.get_val(
+                Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT, units='unitless'
+            )
+        except:
+            if verbosity >= Verbosity.BRIEF:
+                warnings.warn(
+                    'No value for Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT specified, assume False'
+                )
```
jkirk5: A warning might be overkill here since we're just using a default I assume most users want anyway. I'd suggest removing it or making it a higher verbosity level

aviary/core/aviary_group.py
```diff
@@ -982,6 +983,52 @@ def add_post_mission_systems(self, verbosity=None):
...
+            )
+
+        if not ignore_capacity_constraint:
+            self.add_constraint(Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, units='lbm')
+        else:
+            if verbosity >= Verbosity.BRIEF:
+                warnings.warn(
+                    'Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT = True, therefore EXCESS_FUEL_CAPACITY constraint was not added to the Aviary problem.'
+                    'The aircraft may not have enough space for fuel, so check the value of Mission.Constraints.EXCESS_FUEL_CAPACITY for details.'
+                )
```
jkirk5, line 1033: Make this higher verbosity - user had to manually set this flag to True, so we can assume they don't need to see this warning basically every time they run a problem.

Actually, it might be best to only warn the user IF this scenario comes up at all. We could check if excess capacity is negative and warn the user in post-mission?

aviary/docs/theory_guide/images/wing_fuel_capacity_derivation.png
jkirk5, line 1: Can you further crop this image to cut out the table on the left side?

jkirk5, line 1: Where does the 50 lb/ft^3 come from? Is it directly in FLOPS?

aviary/docs/theory_guide/mass.md
```diff
@@ -6,3 +6,45 @@ This includes the empty weight, fuel weight, payload weight, and other mass prop
...
+
+3. Auxiliary Fuel capacity
+
+    Auxiliary Fuel Capacity = Total Fuel Capacity - Wing Fuel Capacity - Fuselage Fuel Capacity
+
+4. Total Fuel Capacity
+
+    Total Fuel Capacity = Wing Fuel Capacity + Fuselage Fuel Capacity + Auxiliary Fuel Capacity
+
+It is possible for Aviary to calculate -ve fuel capacities depending on user overrides and the input wing geometry. These situations are not currently flagged to the user.
```
jkirk5:
```suggestion
It is possible for Aviary to calculate negative fuel capacities depending on user overrides and the input wing geometry. These situations are not currently flagged to the user.
```
Typo? I think this is what you meant?

aviary/utils/preprocessors.py
```diff
@@ -487,6 +499,94 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
...
+        verbosity = aviary_options.get_val(Settings.VERBOSITY)
+
+    if Settings.MASS_METHOD in aviary_options:
+        mass_method = aviary_options.get_val(Settings.MASS_METHOD)
+    else:
+        raise UserWarning('MASS_METHOD not specified. Cannot preprocess fuel capacity inputs.')
+
+    if mass_method == LegacyCode.FLOPS:
+        # find which fuel capacity variables the user has set:
+        try:
```
jkirk5: I think you can remove all of these try blocks by replacing all of the `if var_name is None:` checks later with `if var_name not in aviary_options:`. Maybe makes the sums more complicated though

aviary/utils/preprocessors.py
```diff
@@ -487,6 +499,94 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
...
+        )
+
+        if total_capacity is None:
+            # Aviary will need to calculate the total capacity and can only do so if we assume any missing subsystem capacities are zero
+            if fuselage_capacity is None:
+                fuselage_capacity = 0
+                aviary_options.set_val(
+                    Aircraft.Fuel.FUSELAGE_FUEL_CAPACITY, fuselage_capacity, 'lbm'
+                )
+                if verbosity >= Verbosity.BRIEF:
```
jkirk5: Higher verbosity level here. BRIEF = every Aviary run should output this. Setting a default value is expected behavior, so overkill to warn about every single run.

aviary/subsystems/mass/flops_based/fuel_capacity.py
jkirk5, line 1: While you are changing this file, can you fix the `add_aviary_input/output` calls in setup on all of the components here? I noticed for some reason they are all placed inside tuples for some reason which is super weird.

jkirk5: I think the original lines ended in a comma (again for no known reason) and ruff found that and assumed that meant we wanted it to be a tuple, so it "corrected" the bad formatting

aviary/utils/preprocessors.py
```diff
@@ -487,6 +499,94 @@ def preprocess_crewpayload(aviary_options: AviaryValues, meta_data=_MetaData, ve
...
+            if capacity is not None
+        )
+
+        if total_capacity is None:
+            # Aviary will need to calculate the total capacity and can only do so if we assume any missing subsystem capacities are zero
+            if fuselage_capacity is None:
+                fuselage_capacity = 0
+                aviary_options.set_val(
+                    Aircraft.Fuel.FUSELAGE_FUEL_CAPACITY, fuselage_capacity, 'lbm'
+                )
```
jkirk5: Do we actually have to set it to zero if missing here? This is a component input/output (not an option), so it should get set to the default value in the metadata if it is missing when we setup the component. This looks like it might could be potentially setting an override instead of letting it naturally be zero?

aviary/variable_info/variable_meta_data.py
```diff
@@ -2801,45 +2801,16 @@
...
 add_meta_data(
     Aircraft.Fuel.DENSITY,
     meta_data=_MetaData,
-    historical_name={'GASP': 'INGASP.FUELD', 'FLOPS': None, 'LEAPS1': None},
+    historical_name={'GASP': 'INGASP.FUELD', 'FLOPS': 'WTIN.FULDEN', 'LEAPS1': None},
     units='lbm/galUS',
-    desc='fuel density',
+    desc='fuel density (jet fuel typical density of 6.7 lbm/galUS used in the calculation of wing_capacity'
+    '(if wing_capacity is not input) and in the calculation of fuel system weight.',
     default_value=0.0,
```
jkirk5: We need a default density value imo. Otherwise users have to always define it (and know what the "correct" value is)

aviary/variable_info/variable_meta_data.py
```diff
@@ -2909,6 +2880,18 @@
...
 )
 
+add_meta_data(
+    Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT,
+    meta_data=_MetaData,
+    historical_name={'GASP': None, 'FLOPS': 'WTIN.IFUFU', 'LEAPS1': None},
+    units='unitless',
+    desc='Flag to control enforcement of fuel_capacity constraint. '
+    'If False (default) Aviary will add the excess fuel constraint and only converge if there is enough fuel capacity to complete the mission.'
+    'If set True Aviary will ignore this constraint, and allow mission fuel > total_fuel_capacity. Use Carefully!',
```
jkirk5:
```suggestion
    'If set True Aviary will ignore this constraint, and allow mission fuel > total_fuel_capacity. Use carefully!',
```

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
```diff
@@ -63,8 +63,8 @@ aircraft:fins:mass_scaler,1,unitless
 aircraft:fins:num_fins,0,unitless
 aircraft:fins:taper_ratio,-10,unitless
 aircraft:fuel:auxiliary_fuel_capacity,0,lbm
-aircraft:fuel:capacity_factor,25.903,unitless
-aircraft:fuel:density_ratio,1,unitless
+aircraft:fuel:wing_fuel_fraction,0.7752,unitless
+aircraft:fuel:density,6.7,lbm/galUS
```
jkirk5: Shouldn't be necessary to define in an input file if we are using the assumed default value - see my comment in the variable metadata

aviary/core/aviary_group.py
cmbenne3: Agree - I have removed it.

aviary/docs/theory_guide/images/wing_fuel_capacity_derivation.png
cmbenne3: Done

cmbenne3: FLOPS manual states that jet fuel density is assumed to be 6.7 lbm/galUS.

If you convert to lbm/ft**3 then you get 50.11948
For clarity in my derivation I rounded this to 50.

The FLOPS default FWMAX = 23
If you're being pedantic then the default volume fraction is actually: 23/(50.11948*2/3) = 68.8%, not the rounded 70% I state in the derivation.

image

aviary/subsystems/mass/flops_based/fuel_capacity.py
cmbenne3: Done

aviary/utils/preprocessors.py
cmbenne3: I have removed one try/except block, and moved them all under the main if/else. It would be possible to remove the others, but would require replacing with if/else and wouldn't impact number of lines. Let me know if you think the new way is any cleaner?

aviary/utils/preprocessors.py
cmbenne3: If it is not overwritten then Aviary unhelpfully calculates it as Total Fuel - Wing Fuel. As Total Fuel is also not specified then it defaults to zero and so the fuselage fuel calculates as negative wing fuel capacity. This results in the total fuel remaining at zero, and the mission not converging since there is no fuel in the aircraft.

aviary/utils/preprocessors.py
cmbenne3: Done

aviary/variable_info/variable_meta_data.py
cmbenne3: Agree - I have updated this

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
cmbenne3: I have removed this value from all the aviary input .csv files.

cmbenne3: Unfortunately setting this as a default causes problems with some of the GASP tests. I don't really understand these errors - perhaps you can help!
image

aviary/core/aviary_group.py
cmbenne3: I have moved this warning to methods for level 2 at the end of the run_aviary_problem method. Do we already have a location for post-mission validation (post-processors?). Maybe we leave it here for now and as we add more checks maybe we should do this in a more formalized way...? What do you think?

cmbenne3: I moved this back as discussed. I will add an issue to figure out what we want to do with post optimization validation

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
cmbenne3: We have fixed this by making the gasp fuel density inputs consistent with lbm/galUS. Conversions required in subsystem compute and compute partial methods.

PR #857, opened 2025-08-22
Prevent Aviary running invalid off design missions

### LINKED ISSUE #797: Aviary should disallow "invalid" missions that still converge

Aviary's off design should include at least a message or error within the dashboard informing users that the mission they desire to run has exceeded the structural capabilities defined by the sizing mission and is therefore not valid. This process could be completed by comparing the alternate mission's gross mass to that of the sizing mission and returning an error if the alternate gross mass is greater. 

It would also be possible to modify the off design calculations and make gross mass a constraint and therefore cause the optimizer to fail to converge when running mission ranges beyond the aircraft's maximum payload-range capability. That would allow a more integrated solution that would prevent the user from running "invalid" solutions all together.

When working within Aviary's off-design, alternate missions ran with equal payload mass but increased range still converge despite not resizing. This means that Aviary adds in additional fuel which increases the gross mass of the alternate missions to be greater than the gross mass of the sizing mission. Which means that Aviary is including fuel past the gross mass point of structural failure.

Image

As seen, the sizing mission for the default FwFm has a gross mass of 159057 lbm, increasing the mission range from 1906 to 2000, 3000, and 4000 simply increased the amount of fuel mass carried by the aircraft and entered gross masses of over 180000 lbm.

### Associated Bug Report

_No response_

cmbenne3:
### Summary

This PR prevents Aviary from running successful off-design missions with gross mass greater than the gross mass of the sizing mission.
Alternate missions have an upper bound added to the gross mass design variable 
Fallout missions check the user input mission_mass and throw an error if mission_mass greater than design gross mass.

### Related Issues

- Resolves #797 

None

### New Dependencies

None

### DIFF
--- a/aviary/docs/user_guide/aviary_commands.ipynb
+++ b/aviary/docs/user_guide/aviary_commands.ipynb
@@ -803,6 +803,44 @@
    "source": [
     "More discussion on {glue:md}`aviary dashboard` command can be found in [Postprocessing and Visualizing Results from Aviary](postprocessing_and_visualizing_results.ipynb)."
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "(aviary-rtplot-command)=\n",
+    "### aviary rtplot\n",
+    "\n",
+    "The {glue:md}`aviary rtplot` command runs an Aviary script and also displays a real-time plot of the variables as an optimization progresses. It makes use of the data being written to a case recorder file, the script must add a recorder to the `Driver`."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "```\n",
+    "aviary rtplot -h\n",
+    "```"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "!aviary rtplot -h 2>/dev/null"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "An example of using this command would be\n",
+    "```\n",
+    "aviary rtplot aviary_script.py\n",
+    "```"
+   ]
   }
  ],
  "metadata": {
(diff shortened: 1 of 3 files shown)


cmbenne3:
@jkirk5 - I think the fallout change will conflict with your Off design overhaul #840 so maybe best to either incorporate these changes in your branch first, or we hold off on merging this until that other PR is in.

cmbenne3:
SLSQP doesn't seem to like having to solve the problem with the design variable pinned to the upper bound (run_off_design_example.py). IPOPT is happy with it. I attempted another variation by adding a separate constrain, and leaving the design variable bounds alone and it still wasn't happy with that. If you relax the limit to within 300 pounds it solves fine.

aviary/core/aviary_group.py
```diff
@@ -1245,10 +1245,12 @@ def add_design_variables(self, verbosity=None):
             # fixed vehicle (design GTOW) but variable actual GTOW for off-design
             # mission range
             elif self.problem_type is ProblemType.ALTERNATE:
+                # get the design gross mass and set as the upper bound for the gross mass design variable
+                MTOW = self.aviary_inputs.get_val(Mission.Design.GROSS_MASS, 'lbm')
                 self.add_design_var(
                     Mission.Summary.GROSS_MASS,
                     lower=10.0,
-                    upper=900e3,
+                    upper=MTOW,
```
Kenneth-T-Moore, line 1277: When you run the design mission, the `aviary_inputs` contains the initial guess for the design gross mass. Does that value get updated to the optimized design gross_mass somewhere in this process prior to this point?

cmbenne3: Thanks for this suggestion @Kenneth-T-Moore - I have updated the aviary_inputs inital_guess to be 90% of sizing mission gross mass in the _load_off_design() method for alterante missions and this seems to allow SLSQP to find the correct solution for the run_off_design test. I also updated the ref = MTOW so that it's dynamic rather than static (only for the alternate mission design variable).

Kenneth-T-Moore (approved):

PR #858, opened 2025-08-26
add variables for FLOPS based BWB

xjjiang:
### Summary

For FLOPS based BWB, a few new variables are needed.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/variable_info/variable_meta_data.py
+++ b/aviary/variable_info/variable_meta_data.py
@@ -3435,6 +3435,15 @@ add_meta_data(
 )
 
 add_meta_data(
+    Aircraft.Fuselage.LENGTH_WING_ROOT,
+    meta_data=_MetaData,
+    historical_name={'GASP': None, 'FLOPS': 'FUSEIN.XLW', 'LEAPS1': None},
+    units='ft',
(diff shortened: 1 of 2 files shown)


aviary/variable_info/variable_meta_data.py
```diff
@@ -3597,6 +3606,39 @@
     desc='seats abreast in fuselage',
     types=int,
     option=True,
+    default_value=5,
+)
+
+add_meta_data(
+    Aircraft.Fuselage.NUM_SEATS_ABREAST_BUSINESS,
```
ehariton: So right now we have terms like:
`Aircraft.CrewPayload.Design.NUM_BUSINESS_CLASS` - how many seats are in the aircraft
`Aircraft.CrewPayload.NUM_BUSINESS_CLASS,` - how many passengers are in those types of seats

you're creating `Aircraft.Fuselage.NUM_SEATS_ABREAST_BUSINESS`, to describe the seat configuration
should that go in `Aircraft.CrewPayload.Design.` ?

xjjiang: Agree. Done.

PR #866, opened 2025-09-04
Level 3 Example Shooting

kaushikponnapalli:
### Summary

Modified the level 3 example to work with other transcriptions. Currently this will not solve if using PicardShooting due a bug on the dymos side. When that bugfix is merged in, the problem should work with the new transcription.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/core/aviary_group.py
+++ b/aviary/core/aviary_group.py
@@ -167,7 +167,9 @@ class AviaryGroup(om.Group):
 
                 phase.nonlinear_solver = om.NonlinearRunOnce()
                 phase.linear_solver = om.LinearRunOnce()
-                if isinstance(phase.indep_states, om.ImplicitComponent):
+                if isinstance(phase.options['transcription'], dm.Radau) and isinstance(
+                    phase.indep_states, om.ImplicitComponent
+                ):
                     phase.indep_states.nonlinear_solver = om.NewtonSolver(solve_subsystems=True)
                     phase.indep_states.linear_solver = om.DirectSolver(rhs_checking=True)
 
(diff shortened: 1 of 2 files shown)


aviary/core/aviary_group.py
```diff
@@ -167,7 +167,9 @@ def configure(self):
 
                 phase.nonlinear_solver = om.NonlinearRunOnce()
                 phase.linear_solver = om.LinearRunOnce()
-                if isinstance(phase.indep_states, om.ImplicitComponent):
+                if isinstance(phase.options['transcription'], dm.Radau) and isinstance(
```
Kenneth-T-Moore, line 170: This seems to be leftover code from some other work, or maybe it is the start of supporting Shooting in level 2. The level3 example doesn't use aviary_group.

aviary/examples/FwFm_L3_Example.py
```diff
@@ -106,13 +106,22 @@ def initialize(self):
...
 default_mission_subsystems = [
     prob.model.core_subsystems['aerodynamics'],
     prob.model.core_subsystems['propulsion'],
 ]
-for phase_idx, phase_name in enumerate(phases):
+phases = {}
+
+seg_ends, _ = dm.utils.lgl.lgl(5 + 1)
+transcription = dm.Radau(
+    num_segments=5,
```
Kenneth-T-Moore, line 119: This transcription isn't used.

Kenneth-T-Moore (changes requested):
PR is likely missing the example.

PR #870, opened 2025-09-10
Remove ability to rename openmdao/dymos reports

### LINKED ISSUE #809: Custom report names break the dashboard
### Description

The `record_filename` argument of `run_aviary()` (level 1) and the corresponding argument in `run_aviary_problem()` (level 2) allow for the filename with the problem record (a .db file) to be renamed according to user desire - however this completely breaks the dashboard, which is looking for the default filename.

As far as I can tell, there isn't really any purpose to letting the user rename a specific report, since it now always gets placed inside the problem's reports folder which prevents it from getting confused with other problem reports. This functionality should be removed, as it has limited usefulness and is an easy way to break the dashboard.

### Example

Attempting to open the dashboard when this option has been used creates the following warning:
`/usr/share/miniconda/envs/test/lib/python3.13/site-packages/aviary/visualization/dashboard.py:1096: OpenMDAOWarning:Given Problem case recorder file <problem_name>/dymos_solution.db does not exist`

### Aviary Version

0.10.0-dev

_No response_

jkirk5:
### Summary

Currently the ability to rename some of the openmdao and dymos reports exists in Aviary. There doesn't seem to be any good reason to let users do this, and it breaks dashboard functionality as it can't find those renamed report files, so those options have been removed.

### Related Issues

- Resolves #809

Any models that use those arguments in L1 or L2 will throw errors until they are removed.

### New Dependencies

None


aviary/docs/user_guide/postprocessing_and_visualizing_results.ipynb
```diff
@@ -51,32 +51,30 @@
...
+    "| Model        | N2                                            | ./*name_of_run_script*_out/reports/n2.html                                     |\n",
+    "| Model        | Trajectory Linkage Report                     | ./*name_of_run_script*_out/reports/traj_linkage_report.html                    |\n",
+    "| Optimization | Driver Scaling Report                         | ./*name_of_run_script*_out/reports/driver_scaling_report.html                  |\n",
+    "| Optimization | Total Coloring Report                         | ./*name_of_run_script*_out/reports/total_coloring.html                         |\n",
+    "| Optimization | Optimization Report                           | ./*name_of_run_script*_out/reports/opt_report.html                             |\n",
+    "| Optimization | SNOPT Output (similarly for other optimizers) | ./*name_of_run_script*_out/reports/SNOPT_print.out                             |\n",
+    "| Results      | Subsystem Results                             | ./*name_of_run_script*_out/reports/subsystems/*name_of_subsystem.md (or .html)*|\n",
```
Kenneth-T-Moore, line 65: Results: Trajectory Results Report and Optimization: Driver recording are missing from the new list. Maybe was intentional?

aviary/examples/run_level3_example.py
```diff
@@ -492,40 +493,40 @@ def initialize(self):
...
-        'units',
-        'min',
-        'max',
-    ],
-)
+# prob.model.list_vars(units=True, print_arrays=True)
+# prob.list_driver_vars(
+#     print_arrays=True,
+#     desvar_opts=[
+#         'lower',
```
Kenneth-T-Moore: I think we can delete the desvar_opts and cons_opts lines from this example.

aviary/docs/user_guide/postprocessing_and_visualizing_results.ipynb
jkirk5: Good catch that shouldn't have gotten removed

Kenneth-T-Moore (approved):

PR #872, opened 2025-09-11
BWB components in Geometry/flops_based subsysterms

xjjiang:
### Summary

This PR implements the BWB aircraft geometry subsystem based on FLOPS design.

The FLOPS based geometry subsystems are shown in the following table. The left column is for the conventional aircraft and the right column is for BWB aircraft. In both case, user can choose a simple layout or a detailed layout.

FLOPS has multiple scenarios of fuselage layout algorithms. We choose to implement the simplest and most detailed algorithms in Aviary. Other scenarios can be added if needed.

The BWB aircraft always requires a detailed wing distribution. It is either given (`BWBUpdateDetailedWingDist`) or created (`BWBComputeDetailedWingDist`).

`Aircraft.BWB.MAX_NUM_BAYS` is changed from an option to Aviary variable because it needs to be updated during a run.

`Aircraft.Wing.INPUT_STATION_DIST` is still an option. We created a variable `BWB_INPUT_STATION_DIST` from it.

**Comparision of Conventional Aircraft and BWB Aircraft**

| Tube+Wing | BWB |
| ------------- | ----- |
| `SimpleCabinLayout` or `DetailedCabinLayout`  | `BWBSimpleCabinLayout` or `BWBDetailedCabinLayout` |
|   | `BWBUpdateDetailedWingDist` or `BWBComputeDetailedWingDist` |
| `FuselagePrelim` | `BWBFuselagePrelim` |
| `WingPrelim` | `BWBWingPrelim` |
| `_Prelim` | `_Prelim` |
| `_Wing` | `_BWBWing` |
| `_Tail` | `_Tail` |
| `_Fuselage` | `_BWBFuselage` |
| `_FuselageRatios` | `_FuselageRatios` |
| `Nacelles` | `Nacelles` |
| `Canard` | `Canard` |
| `WingCharacteristicLength` | `BWBWingCharacteristicLength` |
| `OtherCharacteristicLengths` | `OtherCharacteristicLengths` |
| `TotalWettedArea` | `TotalWettedArea` |

As we see in the above table, some geometrical components have no special treatments in the BWB case (e.g. `Nacelles` and `_Tail`). So, the new implementation of BWB geometry uses the same components in conventional aircraft.

In the unit tests, we choose two typical cases: `BWBSimpleCabinLayout` + `BWBComputeDetailedWingDist` and `BWBDetailedCabinLayout` + `BWBUpdateDetailedWingDist`. Both are checked against FLOPS runs and they both match pretty well.

None of the current Aviary model that uses FLOPS based geometry carry out fuselage cabin layout computation. If fuselage dimension variables are missing, they will be computed using GASP based geometry.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/core/aviary_group.py
+++ b/aviary/core/aviary_group.py
@@ -1115,7 +1115,7 @@ class AviaryGroup(om.Group):
         self.configurator.check_trajectory(self)
 
     def _add_bus_variables_and_connect(self):
-        all_subsystems = self.get_all_subsystems()
+        all_subsystems = self.get_all_subsystems(group='pre_mission')
 
         base_phases = list(self.phase_info.keys())
 
@@ -1124,6 +1124,7 @@ class AviaryGroup(om.Group):
             if bus_variables is not None:
                 for bus_variable, variable_data in bus_variables.items():
                     mission_variable_name = variable_data['mission_name']
+                    src_indices = variable_data.get('src_indices', None)
 
                     # check if mission_variable_name is a list
                     if not isinstance(mission_variable_name, list):
@@ -1161,6 +1162,7 @@ class AviaryGroup(om.Group):
                                     self.connect(
                                         f'pre_mission.{bus_variable}',
                                         f'traj.{phase_name}.parameters:{mission_var_name}',
+                                        src_indices=src_indices,
                                     )
 
                             else:
@@ -1176,6 +1178,7 @@ class AviaryGroup(om.Group):
                                 self.connect(
                                     f'pre_mission.{bus_variable}',
                                     'traj.parameters:' + mission_var_name,
+                                    src_indices=src_indices,
                                 )
 
                     if 'post_mission_name' in variable_data:
@@ -1188,17 +1191,20 @@ class AviaryGroup(om.Group):
                             self.connect(
                                 f'pre_mission.{bus_variable}',
                                 post_mission_var_name,
+                                src_indices=src_indices,
                             )
 
     def _connect_mission_bus_variables(self):
-        all_subsystems = self.get_all_subsystems()
+        all_subsystems = self.get_all_subsystems(group='post_mission')
 
         # Loop through all external subsystems.
         for external_subsystem in all_subsystems:
             for phase_name, var_mapping in external_subsystem.get_post_mission_bus_variables(
                 aviary_inputs=self.aviary_inputs, phase_info=self.phase_info
             ).items():
-                for mission_variable_name, post_mission_variable_names in var_mapping.items():
+                for mission_variable_name, variable_data in var_mapping.items():
+                    post_mission_variable_names = variable_data['post_mission_name']
+                    src_indices = variable_data.get('src_indices', None)
                     if not isinstance(post_mission_variable_names, list):
                         post_mission_variable_names = [post_mission_variable_names]
 
@@ -1206,7 +1212,7 @@ class AviaryGroup(om.Group):
                         # Remove possible prefix before a `.`, like <external_subsystem_name>.<var_name>"
                         mvn_basename = mission_variable_name.rpartition('.')[-1]
                         src_name = f'traj.{phase_name}.mission_bus_variables.{mvn_basename}'
-                        self.connect(src_name, post_mission_var_name)
+                        self.connect(src_name, post_mission_var_name, src_indices=src_indices)
 
     def add_design_variables(self, problem_type: ProblemType = None, verbosity=None):
         """
@@ -1442,10 +1448,13 @@ class AviaryGroup(om.Group):
                 self, phase_name, phase, guesses, target_prob, parent_prefix
             )
 
-    def get_all_subsystems(self, external_subsystems=None):
+    def get_all_subsystems(self, external_subsystems=None, group='pre_mission'):
         all_subsystems = []
         if external_subsystems is None:
-            all_subsystems.extend(self.pre_mission_info['external_subsystems'])
+            if group == 'pre_mission':
+                all_subsystems.extend(self.pre_mission_info['external_subsystems'])
+            elif group == 'post_mission':
+                all_subsystems.extend(self.post_mission_info['external_subsystems'])
         else:
             all_subsystems.extend(external_subsystems)
 
--- a/aviary/core/post_mission_group.py
+++ b/aviary/core/post_mission_group.py
@@ -1,10 +1,19 @@
+import openmdao
 import openmdao.api as om
+from packaging import version
 
 from aviary.utils.functions import promote_aircraft_and_mission_vars
+
+use_new_openmdao_syntax = version.parse(openmdao.__version__) >= version.parse('3.28')
 
 
 class PostMissionGroup(om.Group):
     """OpenMDAO group that holds all post-mission systems."""
+
+    def setup(self, **kwargs):
+        if use_new_openmdao_syntax:
+            # rely on openMDAO's auto-ordering for this group
+            self.options['auto_order'] = True
 
     def configure(self):
         """
--- a/aviary/subsystems/aerodynamics/gasp_based/test/test_common.py
+++ b/aviary/subsystems/aerodynamics/gasp_based/test/test_common.py
@@ -3,6 +3,7 @@ import unittest
 import numpy as np
 import openmdao.api as om
 from openmdao.utils.assert_utils import assert_check_partials, assert_near_equal
+from openmdao.utils.testing_utils import use_tempdirs
 
 from aviary.subsystems.aerodynamics.gasp_based.common import (
     AeroForces,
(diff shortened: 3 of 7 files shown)


aviary/docs/examples/modified_aircraft.csv
```diff
@@ -119,7 +119,7 @@ aircraft:wing:chord_per_semispan,0.31,0.23,0.084,unitless
 aircraft:wing:composite_fraction,0.2,unitless
 aircraft:wing:control_surface_area,137,ft**2
 aircraft:wing:control_surface_area_ratio,0.1,unitless
-aircraft:wing:detailed_wing,True,unitless
+aircraft:wing:detailed_wing,1,unitless
```
Kenneth-T-Moore, line 122: Seems like this should be True, even though 1 is truthy.

aviary/variable_info/enums.py
```diff
@@ -45,6 +46,17 @@ class AlphaModes(Enum):
...
+    """
+    To flag where users provide detailed wing or want to create detailed wing (if the aircraft is BWB)
+    or not to use detailed wing
+    """
+
+    TO_PROVIDE = 1
+    TO_COMPUTE = -1
+    NOT_TO_USE = 0
+
+
```
Kenneth-T-Moore: I'm not sure what to think about this 3-value enum.  Two choices are valid for transport, and two choices are valid for BWB. That makes me think it should be 2 separate options, maybe DetailedWing and BWBDetailedWing.  (And in that case, transport could still be a bool, but BWB would still be an enum with COMPUTED or PROVIDED as the enum keys.

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
```diff
@@ -0,0 +1,391 @@
+import numpy as np
+import openmdao.api as om
```
Kenneth-T-Moore, line 2: I think i would rename this file to bwb_wing_detailed.py

aviary/subsystems/geometry/flops_based/prep_geom.py
```diff
@@ -546,6 +623,62 @@ def compute_partials(self, inputs, J, discrete_inputs=None):
...
+
+        add_aviary_input(self, Aircraft.Fuselage.MAX_WIDTH, units='ft')
+        add_aviary_input(self, Aircraft.Wing.GLOVE_AND_BAT, units='ft**2')
+        add_aviary_input(self, Aircraft.Wing.SPAN, units='ft')
+        self.add_input('BWB_CHORD_PER_SEMISPAN_DIST', shape=num_stations, units='unitless')
+        self.add_input('BWB_THICKNESS_TO_CHORD_DIST', shape=num_stations, units='unitless')
+
+        add_aviary_output(self, Aircraft.Wing.WETTED_AREA, units='ft**2')
+
+        self.declare_partials('*', '*', method='fd', form='forward')
```
Kenneth-T-Moore: Looks like this is safe to use 'cs' instead of 'fd'.

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
xjjiang: Good idea. Done.

aviary/subsystems/geometry/flops_based/prep_geom.py
xjjiang: done.

aviary/variable_info/enums.py
xjjiang: I created a separate variable `Aircraft.BWB.DETAILED_WING_PROVIDED` instead which is also a boolean.

aviary/subsystems/geometry/flops_based/prep_geom.py
```diff
@@ -39,61 +48,119 @@ class PrepGeom(om.Group):
...
+                    promotes_outputs=['*'],
+                )
+            else:
+                self.add_subsystem(
+                    'fuselage_layout',
+                    DetailedCabinLayout(),
+                    promotes_inputs=['*'],
+                    promotes_outputs=['*'],
+                )
+        else:
```
Kenneth-T-Moore: I think this `else` branch isn't needed because the `add_aviary_option` sets the types to whatever is in the "types" field in the metadata, which here is `types=AircraftTypes,`.  The OpenMDAO options dictionary will type check is, so that only valid AicraftTypes are allowed to be set.

aviary/variable_info/enums.py
```diff
@@ -6,6 +6,7 @@ class AircraftTypes(Enum):
 
     TRANSPORT = 'transport'
     BLENDED_WING_BODY = 'BWB'
+    GENERAL_AVIATION = 'GA'  # incomplete in FLOPS, unavailable in GASP
```
Kenneth-T-Moore: This is a good addition, but I would comment it out for now. If someone sets this option to GA, it will break some things without giving a clear indication about why.

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
```diff
@@ -0,0 +1,391 @@
...
+        outputs['BWB_CHORD_PER_SEMISPAN_DIST'][0] = length
+        outputs['BWB_CHORD_PER_SEMISPAN_DIST'][1] = xl_out
+
+        for i in range(2, num_stations):
+            x = inputs[Aircraft.Wing.THICKNESS_TO_CHORD_DIST][i]
+            outputs['BWB_THICKNESS_TO_CHORD_DIST'][i] = x
+        outputs['BWB_THICKNESS_TO_CHORD_DIST'][0] = tc
+        outputs['BWB_THICKNESS_TO_CHORD_DIST'][1] = tc
+
+        for i in range(0, num_stations):
```
Kenneth-T-Moore: For speed and code simplicity, it is best to replace loops with various ways that numpy can handle vector broadcasts.


outputs['BWB_CHORD_PER_SEMISPAN_DIST'] = inputs[Aircraft.Wing.CHORD_PER_SEMISPAN_DIST]
idx = np.where(inputs[Aircraft.Wing.CHORD_PER_SEMISPAN_DIST] < 5.0)
outputs['BWB_CHORD_PER_SEMISPAN_DIST'][idx] *= rate_span

outputs['BWB_CHORD_PER_SEMISPAN_DIST'][0] = length
outputs['BWB_CHORD_PER_SEMISPAN_DIST'][1] = xl_out
		

outputs['BWB_THICKNESS_TO_CHORD_DIST'][0] = tc
outputs['BWB_THICKNESS_TO_CHORD_DIST'][1] = tc
outputs['BWB_THICKNESS_TO_CHORD_DIST'][2:] = inputs[Aircraft.Wing.THICKNESS_TO_CHORD_DIST][2:]
		
outputs['BWB_LOAD_PATH_SWEEP_DIST'][:] = inputs[Aircraft.Wing.LOAD_PATH_SWEEP_DIST]

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
```diff
@@ -0,0 +1,391 @@
...
+        num_stations = len(self.options[Aircraft.Wing.INPUT_STATION_DIST])
+        for i in range(2, num_stations):
+            x = self.options[Aircraft.Wing.INPUT_STATION_DIST][i]
+            if x <= 1.0:
+                y = x * rate_span + width / wingspan
+            else:
+                y = x + width / 2.0
+            self.options[Aircraft.Wing.INPUT_STATION_DIST][i] = y
+        self.options[Aircraft.Wing.INPUT_STATION_DIST][0] = 0.0
+        self.options[Aircraft.Wing.INPUT_STATION_DIST][1] = width / 2.0
```
Kenneth-T-Moore: You should never set options inside of a component. An option is meant to be something the user sets. If this is actually being propagated to downstream components, it is unintentional, and probably only works because the same distribution array object happens to be shared by all of the components that have this same option. 

I see the input station distribution depends on span. This variable can change during a design optimization, which means that we need derivatives of and wrt it.  Also, if this component were to run more than once in an optimization, it might have a cumulative multiplicative affect on the values.

Alternatively, maybe any adjustments to the station distribution should be done in a pre-processor.

aviary/variable_info/enums.py
xjjiang: Commented out.

aviary/subsystems/geometry/flops_based/prep_geom.py
xjjiang: removed it.

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
xjjiang: All the options that update are converted to variables now. I thought about per-processor but realized that it was not possible.

aviary/subsystems/geometry/flops_based/bwb_wing_detailed.py
xjjiang: Done.

Kenneth-T-Moore (approved):

PR #876, opened 2025-09-15
Add checking for benchmark test convergence

cmbenne3:
### Summary

This PR adds an additional check into most of the benchmark tests to ensure that the problem has actually converged successfully. Prior to this it was possible for the tests to pass even though the problem was failing.
Some tests are still missing this check and issue #881 lists them. These need to be fixed in future.

The test_bwb_GwGm test has been updated to use the default phase info and the generic_BWB_phase_info file has been removed.

### Related Issues

None

None

### New Dependencies

None


xjjiang:
> are you happy that this change to the BWB model is ok?

Sure. Thanks for the improvement. I used `verbosity=2` to debug. Then I forgot to set it back to 1.

aviary/validation_cases/benchmark_tests/test_bwb_GwGm.py
```diff
@@ -25,46 +24,48 @@ def setUp(self):
...
+
         rtol = 1e-3
 
         # There are no truth values for these.
         assert_near_equal(
             prob.get_val(Mission.Design.GROSS_MASS, units='lbm'),
-            151764.0407,
+            141260.50569975,
             tolerance=rtol,
         )  # WG = 150000.0
```
xjjiang, line 44: Perhaps we can remove the numbers from GASP (e.g. `WG = 150000.0` and so on) since we are not trying to compare to GASP runs.

PR #880, opened 2025-09-16
Expose primal_name to user for better support of jax aviary components.

Kenneth-T-Moore:
### Summary

The former implementation was a little clunky, in that the user had to keep long names (like "aircraft__wing__span") in their `compute_primal` code. This change exposes the "primal_name" argument to `add_aviary_input` and `add_aviary_output` so that they can choose variable names that are more readable and maintainable.

### Related Issues

- Resolves #

None

### New Dependencies

None


xjjiang:
Will there be a doc page on `compute_primal` and `primal_name`?

PR #893, opened 2025-10-01
Fix draw_mission so that it can read in the file it writes.

### LINKED ISSUE #892: Mission GUI utility cannot load phase info files
### Description

The Mission GUI has the ability to read in existing phase_info files, however with the updates to which keys are available in phase_info, this ability is now broken.

### Example

Running `aviary draw_mission`, creating any mission profile and saving it, then trying to load that same file causes an error and the mission profile cannot be loaded.

> Exception in Tkinter callback
> Traceback (most recent call last):
>   File "/home/miniforge3/envs/aviary/lib/python3.12/tkinter/__init__.py", line 1968, in __call__
>     return self.func(*args)
>            ^^^^^^^^^^^^^^^^
>   File "/Documents/python/Aviary/aviary/interface/graphical_input.py", line 1405, in open_phase_info
>     value=phase_dict['user_options']['polynomial_order']
>           ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
> KeyError: 'polynomial_order'

### Aviary Version

0.10.0-dev

_No response_

Kenneth-T-Moore:
### Summary

The draw_mission utility can now read in a mission file that it authored. There were a few phase_info keys that changed names in the revamp, but weren't updated here.

Note: the builder will have trouble reading in hand-written phase_info files because the format is more flexible than the draw_mission utility can handle.

### Related Issues

- Resolves #892 

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/aerodynamics/gasp_based/gaspaero.py
+++ b/aviary/subsystems/aerodynamics/gasp_based/gaspaero.py
@@ -282,9 +282,13 @@ class BWBBodyLiftCurveSlope(om.ExplicitComponent):
     def compute(self, inputs, outputs):
         verbosity = self.options[Settings.VERBOSITY]
         mach = inputs[Dynamic.Atmosphere.MACH]
-        if any(x < 0.0 or x >= 1.0 for x in mach.real):
+
+        # Provide a little buffer for being close to the limit.
+        eps = 1e-8
+
+        if any(x < -eps or x >= 1.0 + eps for x in mach.real):
             raise om.AnalysisError('Mach number must be within the range (0, 1).')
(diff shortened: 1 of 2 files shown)


cmbenne3 (approved):
I have tested this and approve it to fix the issue where the GUI was unable to read in Phase Infos that were created in it. 
This is outside the scope for this PR, but I noticed that we are unable to read in and display our default Height Energy and 2DOF phase infos using this utility:
Error with aviary/models/missions/height_energy_default.py:
image

Error with aviary/models/missions/two_dof_default.py:
image

naylor-b (approved):

PR #905, opened 2025-10-22
Support `src_indices` in bus variable connections

kanekosh:
### Summary
This PR proposes adding support for `src_indices` for bus variable connections from pre-mission to mission, pre-mission to post-mission, and mission to post-mission. I think this is especially handy for mission-to-post-mission connections because we often want to use the initial or final value of mission vectors in post-mission computations.

This PR also fixes a bug in `_connect_mission_bus_variables` where `get_post_mission_bus_variables` of an external builder is not called when this builder is only added to the `post_mission` dict in `phase_info`. This is because `_connect_mission_bus_variables` was looping over all external subsystems added to the `pre_mission` dict; it now uses external subsystems from the `post_mission` dict.

Finally, not directly related to the title, but I added auto_order to the post-mission group, so external groups will be added in appropriate order. (I was trying to use fuel burn in my post-mission builder, but currently, the external group is added before Aviary's default groups, like fuel burn component, so there was an unnecessary feedback loop.)

### Related Issues

- Resolves #

To allow users to define `src_indices`, I made a backward-incompatible change in the format of the dictionary returned by `get_post_mission_bus_variables`. The new dictionary format is closer to the one used in `get_pre_mission_bus_variables`. 

It looks like `get_post_mission_bus_variables` is not very frequently used (at least within the Aviary main repo), so hopefully this won't affect people's code. I can also add support to the current format to not break the backward compatibility if needed, so please let me know.

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/test/test_external_subsystem_bus.py
+++ b/aviary/subsystems/test/test_external_subsystem_bus.py
@@ -209,11 +209,17 @@ class CustomBuilder(SubsystemBuilderBase):
             phase_d = {}
             if phase_data.get('do_the_zz_thing', False):
                 if self.mangle_names:
-                    phase_d[f'{name}.{name}_zz'] = f'{name}.{phase_name}_{name}_zz'
-                    phase_d[Dynamic.Mission.VELOCITY] = f'{name}.{phase_name}_{name}_velocity'
+                    phase_d[f'{name}.{name}_zz'] = {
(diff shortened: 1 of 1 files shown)


Kenneth-T-Moore:
Thanks, Shugo. This looks really good. The only person using the post-mission bus right now is here at NASA GRC, so I don't think the UI change will be a problem.

PR #906, opened 2025-10-23
Improve robustness of Advanced Single Aisle; Add capability to choose engine deck variable to interpolate first.

Kenneth-T-Moore:
### Summary

1. Made a few tweaks on the advanced single aisle mission to get it to perform better in the sizing optimization.
2. Added a new setting "aircraft:engine:interpolation_sort" that let's you choose "altitude" as the first table dimension in the semi-structured metamodel instead of the default "mach". The first dimension defines the outer structure in this metamodel, and how you choose can affect the semi-extrapolated values when you are outside of the clustered data. The graphics below (from the OpenMDAO view_mm feature) illustrate this, where low-mach, high-altitude has some peculiarities if you choose mach as the first table var.  You wouldn't fly in this area, but you visit during optimization.

Note: I didn't move any aircraft over to use "altitude" as the first dimension yet.

mach (aviary default)
p1

altitude (can choose this now)
p2

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/.github/workflows/test_docs.yml
+++ b/.github/workflows/test_docs.yml
@@ -56,7 +56,7 @@ jobs:
         id: build_docs
         shell: bash -l {0}
         run: |
-          pip install -U jupyter-book
+          pip install -U "jupyter-book<2"
           cd aviary/docs
           echo "============================================================="
           echo "Build the docs"
--- a/aviary/docs/examples/modified_aircraft.csv
+++ b/aviary/docs/examples/modified_aircraft.csv
@@ -133,7 +133,7 @@ aircraft:wing:shear_control_mass_scaler,1.0,unitless
 aircraft:wing:span_efficiency_reduction,False,unitless
 aircraft:wing:span,117.83,ft
 aircraft:wing:strut_bracing_factor,0.0,unitless
-aircraft:wing:surface_ctrl_mass_scaler,1.0,unitless
+aircraft:wing:surface_control_mass_scaler,1.0,unitless
 aircraft:wing:sweep,25.0,deg
 aircraft:wing:taper_ratio,0.278,unitless
 aircraft:wing:thickness_to_chord_dist,0.145,0.115,0.104,unitless
--- a/aviary/interface/test/sizing_problem_for_test.json
+++ b/aviary/interface/test/sizing_problem_for_test.json
@@ -1151,7 +1151,7 @@
         "<class 'int'>"
     ],
     [
-        "aircraft:wing:surface_ctrl_mass_scaler",
+        "aircraft:wing:surface_control_mass_scaler",
         1,
         "unitless",
         "<class 'int'>"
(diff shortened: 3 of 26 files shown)


aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1460,6 +1513,7 @@ def _sort_data(self):
         Sort unpacked engine data in order of Mach number, altitude, throttle,
```
cmbenne3: This comment should be updated to say "sort unpacked engine data in order depending on value of Aircraft.Engine.INTERPOLATION_SORT" or something similar, or removed as it is no longer completely accurate

aviary/variable_info/variable_meta_data.py
```diff
@@ -2196,6 +2196,20 @@
...
+add_meta_data(
+    Aircraft.Engine.INTERPOLATION_SORT,
+    meta_data=_MetaData,
+    historical_name={'GASP': None, 'FLOPS': None, 'LEAPS1': None},
+    units='unitless',
+    option=True,
+    default_value='mach',
+    types=str,
+    desc='Specify the first interpolation variable in the semi-structured metamodel. '
+    'Choose from mach or altitude. Mach is usually the first column in the deck, but '
```
cmbenne3, line 2235: In the current implementation the choice is actually 'altitude' or anything else that is not 'altitude'. Perhaps we should change the implementation to be 'altitude' and 'mach' and if it's neither of those options throw a warning and assume the user meant 'mach'.

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -805,7 +807,11 @@ def _build_engine_interpolator(self, num_nodes, aviary_inputs):
 
         # add inputs and outputs to interpolator
         # independent variables that currently MUST be inputs
-        independent_variables = [MACH, ALTITUDE, THROTTLE, HYBRID_THROTTLE]
+        if interp_sort == 'altitude':
```
cmbenne3, line 819: Should we code this as follows:
> if == 'altitude'
> elif == 'mach'
> else raise UserWarning(f"invalid value of Aircraft.Engine.INTERPOLATION_SORT. Expected 'altitude' or 'mach' found {interp_sort}. Assuming 'mach' for backwards compatability.")

If user accidentally sets the value to "Altitude" then there is nothing to help them spot that they've made a mistake...

aviary/subsystems/propulsion/engine_deck.py
Kenneth-T-Moore: good catch fixed.

aviary/subsystems/propulsion/engine_deck.py
Kenneth-T-Moore: I've added value checking, but placed it in "_preprocess_inputs", since that occurs a little earlier. Also added a test for the error message.

cmbenne3 (approved):

PR #915, opened 2025-11-10
Update Dynamic.Mission.DISTANCE to Dynamic.Mission.GROUND_DISTANCE for clarity

### LINKED ISSUE #413: Unify mission range and distance variable names

`Dynamic.Mission.RANGE` and `Dynamic.Mission.DISTANCE` have the same meaning in Aviary. We should choose one and rename throughout Aviary.
The difference is probably a holdover from the LEAPS2 and GASPy codes.

@johnjasa: We only need one of them. I'd say distance is probably most clear?
@ehariton: My preference is to go with `Mission.Dynamic.Range` since `Distance` could be interpreted as vertical or horizontal, but `Range` has a stronger connection with horizontal movement. We should leave an alias reference so people familiar with one can see that we've changed the var name.

_No response_

### Associated Bug Report

_No response_

cmbenne3:
### Summary

This PR addresses issue 413 and changes the name of the Dynamic.Mission.DISTANCE variable to be more explicit Dynamic.Mission.GROUND_DISTANCE. So there is no confusion about which distance is actually being calculated.

### Related Issues

- Resolves #413 

Variable change - any custom scripts people have written that interrogate the value of the old variable Dynamic.Mission.DISTANCE will fail as this variable no longer exists.

### New Dependencies

None

### DIFF
--- a/aviary/core/aviary_group.py
+++ b/aviary/core/aviary_group.py
@@ -916,12 +916,12 @@ class AviaryGroup(om.Group):
                     ),
                 )
                 self.connect(
-                    f'traj.{phase_name}.timeseries.distance',
+                    f'traj.{phase_name}.timeseries.ground_distance',
                     f'{phase_name}_distance_constraint.final_distance',
                     src_indices=[-1],
                 )
                 self.connect(
-                    f'traj.{phase_name}.timeseries.distance',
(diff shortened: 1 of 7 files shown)


cmbenne3:
Closing this PR as stale - issue still exists.

PR #916, opened 2025-11-14
Some updates for benchmark performance.

Kenneth-T-Moore:
### Summary

1 Reduced the SNOPT feasibility tolerance from 1e-7 to 1e-6 to match the settings for IPOPT and take the extra burden off the optimizer.
2 Changed the values of some fixed refs.
3 In the Alternate problem, relaxed the upper bound of the gross mass constraint because it was causing the optimizer some problems.
4 Cleaned up the add_control call in the phase builder to remove possible subtle differences between None and _undefined in dymos.
5 Add missing "continuity_ref" to the add_state call, and set it to the same value as the ref.
6 Fixed broken target_range constraint in the 2dof configurator.
7 Restored a separate phase-info for the BWB because it flies a different altitude, and needed some specific tuning of scaling parameters.
8 Fixed a few issues in the default 2dof phase_info:
- Angles in the Rotation phase mistakenly were declared as radians, but were actually in degrees.
- Added missing initial altitude for climb 1. (This must have broken at some point. The accel phase is strange, with no altitude to connect to.)
- Changed test values to correspond to correct phase info.
9 Added unique distance ref for later phases in the height_energy phase_info since those phases have a value for distance.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore:
> The file name `generic_BWB_phase_info.py` should be changed to include `2dof` because we might need another `phase_info` file for FLOPS based models.

thanks, I've added 2dof into the title of that mission info.

aviary/core/aviary_group.py
```diff
@@ -1023,7 +1023,7 @@ def add_post_mission_systems(self, verbosity=None):
 
         if not ignore_capacity_constraint:
             self.add_constraint(
-                Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, ref=1000, units='lbm'
+                Mission.Constraints.EXCESS_FUEL_CAPACITY, lower=0, ref=1.0e6, units='lbm'
```
cmbenne3, line 1026: No change required - Combined with feasibility tolerance of 1e-6, optimizer will satisfy this condition to within 1lbm.

aviary/core/aviary_group.py
```diff
@@ -1318,7 +1318,7 @@ def add_design_variables(self, problem_type: ProblemType = None, verbosity=None)
                 )
 
                 if self.require_range_residual:
-                    self.add_constraint(Mission.Constraints.RANGE_RESIDUAL, equals=0, ref=10)
+                    self.add_constraint(Mission.Constraints.RANGE_RESIDUAL, equals=0, ref=1000)
```
cmbenne3, line 1321: Range residual satisfied to within 1e-3 Nautical Miles = 6ft.

Should we add units to this constraint for completeness?

aviary/mission/phase_builder_base.py
```diff
@@ -420,6 +420,12 @@ def add_state(self, name, target, rate_source):
         defect_ref, _ = options[f'{name}_defect_ref']
         solve_segments = options[f'{name}_solve_segments']
 
+        extra_options = {}
```
cmbenne3, line 423: Does this add functionality or is it just a different way of achieving the same thing? I found the old code easier to follow.

Kenneth-T-Moore: This change is a little harder to follow. In dymos, the default value for all of these arguments is `_undefined`, which is an object that was created to represent no input from the user.  Our default in Aviary for unspecified phase_info arguments is None, which is more pythonic. I was a little worried that there might be some subtle behavior differences if we passed in None as a value instead of just not specifying a value, so I changed this just to be safe.

Kenneth-T-Moore:
> In the PR description you talk about updating the add_control call (4), but I think the changes are to the add_state call?

Oops, you are right. I will update the description.

aviary/core/aviary_group.py
Kenneth-T-Moore: Most of the time, "units" isn't needed on an add_constraint call because it uses the units defined on the output being constrained. It is available for the corner case where you might want to give the optimizer a different (potentially better-scaled) unit.

xjjiang:
Are there user docs to be updated?

aviary/mission/phase_builder_base.py
jkirk5: Can you add a comment to explain that in the code? `options default to "undefined" in dymos if not specified in extra_options?` or something similar

aviary/core/aviary_group.py
```diff
@@ -1328,12 +1328,12 @@ def add_design_variables(self, problem_type: ProblemType = None, verbosity=None)
                 self.add_design_var(
```
jkirk5, line 1328: What stops the problem from converging to a gross mass above design TOGW? I think this could happen if you have an aircraft loaded up with a lot of cargo and a long mission - maybe the optimizer might pick a design where cargo + fuel takes it above TOGW, but fuel is within capacity?

jkirk5 (commented):
Quick question on one of the changes to constraints

aviary/core/aviary_group.py
Kenneth-T-Moore: I think this could happen, perhaps in a case where the fuel capacity is higher than needed for the design mission, so it could fly much heavier or further. I guess the question is whether it is something that should be prevented in the optimization problem.

cmbenne3: The user might prefer a more stable converged solution with an 'error' message added to post mission stating that the requested alternate mission cannot be flown within the design gross mass. It's very frustrating if the optimizer can't find the answer and just 'fails'. We know that for a lot of 'valid' missions the design variable will be hard up on the upper bound.

jkirk5: Right now that error message doesn't exist, so the problem would converge, say "success" on the dashboard, but is actually physically invalid. IMO that is worse, because it hides that from the user. At least an unconverged problem is clearly a failure. And a message doesn't help once you close the terminal... then you just have to know to always check that design variable before using any results

Kenneth-T-Moore: let me try rerunning without this bounds-change on Jason's off-design branch, since he reworked some of the tests.

Yeah, I don't think this is needed anymore. The test that fails without this has been replaced in Jason's PR.

PR #926, opened 2025-12-01
WIP fixes for turboprop aircraft

jkirk5:
### Summary

A variety of bugfixes and changes to improve the turboprop freighter model. NOTE this is a partial fix and does not mean the turboprop freighter is finished at this point.

Changes include cleaning up and standardizing the electric breguet range ODEs and phases, getting parts of 2DOF working with external subsystems (a new feature technically!), and lots of fixes geared towards the freighter model itself

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -1108,7 +1108,21 @@ def link_phases(self, verbosity=None, comm=None):
                     phases_to_link.append(phase_name)
 
             if len(phases_to_link) > 1:  # TODO: hack
```
jkirk5, line 1121: Experimental fix for external subsystem linked variables across analytic phases

aviary/mission/height_energy/ode/energy_ODE.py
```diff
@@ -186,7 +186,7 @@ def setup(self):
 
             sub1.nonlinear_solver.linesearch = om.BoundsEnforceLS()
             sub1.linear_solver = om.DirectSolver(assemble_jac=True)
-            sub1.nonlinear_solver.options['err_on_non_converge'] = True
+            sub1.nonlinear_solver.options['err_on_non_converge'] = False
```
jkirk5: Was really hard to diagnose/debug problems with error on non-converge prematurely halting the problem and not letting optimizer spin out & report results

aviary/subsystems/propulsion/propeller/propeller_builder.py
```diff
@@ -52,24 +52,24 @@ def get_design_vars(self):
...
-                'upper': None,
-                # 'val': 8,  # initial value
-            },
-            Aircraft.Engine.Propeller.INTEGRATED_LIFT_COEFFICIENT: {
-                'units': 'unitless',
-                'lower': 0.0,
-                'upper': 0.5,
-                # 'val': 0.5,
-            },
+            # Aircraft.Engine.Propeller.ACTIVITY_FACTOR: {
```
jkirk5, line 55: This is actually a new problem - these are the design variables that _could_ get added, but you don't always want them added to the problem, only when you want to resize the propeller (and even then, maybe you don't want diameter to change, for example). We don't have a good way to easily let the user set this right now

Kenneth-T-Moore: In MBSAE, the way to handle this has been to manually comment out design vars, but you could add a flag to the builder initialization arguments, and use this flag to determine the var configuration.

aviary/mission/height_energy/ode/energy_ODE.py
Kenneth-T-Moore: "err_on_non_converge" should raise an AnalysisError that tells SNOPT or IPOPT to try a smaller stepsize. I wonder if this is being re-raised as a different error at a higher level.

jkirk5: Might be worth re-running the model with that flag False now that this branch is on latest main and see if I still have issues

PR #931, opened 2025-12-05
Fixes an exception when the user adds a hierarchy variable without metadata.

Kenneth-T-Moore:
### Summary

Fixed an exception when the user adds a hierarchy variable without metadata. While we don't want this to happen with internal Aviary variables (and we have a test to prevent it), the user can always create a variable such as "aicraft:something" without adding it to their custom metadata. We should still be able to generate the input report in that case, and this PR fixes a bug that prevented it.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/reports.py
+++ b/aviary/interface/reports.py
@@ -414,12 +414,12 @@ def input_check_report(prob: AviaryProblem, **kwargs):
 
                 try:
                     units = metadata['units']
-                except:
+                except (TypeError, KeyError):
                     metadata = aviary_metadata.get(var.split('.')[-1])
 
                     try:
                         units = metadata['units']
-                    except:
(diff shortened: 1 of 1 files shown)


aviary/interface/reports.py
```diff
@@ -410,15 +410,22 @@ def abs2prom(abs_name):
...
+
                 try:
                     units = metadata['units']
                 except:
                     metadata = aviary_metadata.get(var.split('.')[-1])
-                    units = metadata['units']
+
+                    try:
+                        units = metadata['units']
+                    except:
```
jkirk5:
```suggestion
                    except KeyError:
```
Avoid bare except so we can catch unrelated issues in the future

PR #935, opened 2025-12-10
L3 shooting

cmbenne3:
### Summary

This PR makes the necessary changes to enable the Dymos PicardShooting transcription to be used in addition to the current collocation (Radau) transcription.

A new setting 'transcription' has been added into the phase_info declaration. This has two options:

1. 'Collocation'
2. 'PicardShooting' 

The make_default_transcription() method of the flight_phase_builder class has been updated to use this setting to set the dymos transcription of each phase. An error message is thrown if the user sets this to something that is not supported (anything other than the above options). If the user omits does not specify a transcription then 'Collocation' is used by default to maintain backwards compatibility.

Adding this setting to the phase_info gives this functionality to level 1 and level 2 users.

To support this change the way the control, state and time initialization values are set has been updates to use the correct Dymos methods, and Dymos handles the differences for shooting under the hood.

The L3 example has been updated to unpack the trajectory setup and expose the interactions with Dymos and OpenMDAO. The L3 example still runs the same aircraft and mission as the L2 and L1 examples.

2 new tests are added to the test_height_energy_mission suite to test that the shooting method works:
1. For basic case of FwFm with the nominal phase info
2. For a more advanced case where mach, and altitude are optimized and a throttle path constraint is added.

Since the phase info polynomial control orders are still set to 1 I think both of these missions will still consist of straight line segments for each phase.

The docs have been sparsely updated in the places where collocation is mentioned. I wasn't sure what to do for this page:
image
Does this still hold true for shooting?

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/test/test_height_energy_mission.py
+++ b/aviary/interface/test/test_height_energy_mission.py
@@ -15,6 +15,7 @@ from aviary.interface.methods_for_level2 import AviaryProblem
 from aviary.mission.height_energy.phases.energy_phase import EnergyPhase
 from aviary.subsystems.test.test_dummy_subsystem import ArrayGuessSubsystemBuilder
 from aviary.variable_info.variables import Dynamic
+from aviary.variable_info.enums import Transcription
 
 
 @use_tempdirs
@@ -159,7 +160,9 @@ class AircraftMissionTestSuite(unittest.TestCase):
     def test_mission_basic_shooting_pyopt(self):
         modified_phase_info = self.phase_info.copy()
         for phase in ['climb', 'cruise', 'descent']:
-            modified_phase_info[phase]['user_options']['transcription'] = 'PicardShooting'
+            modified_phase_info[phase]['user_options']['transcription'] = (
+                Transcription.PICARDSHOOTING
+            )
         prob = self.run_mission(modified_phase_info, 'IPOPT')
         self.assertIsNotNone(prob)
         self.assertTrue(prob.result.success)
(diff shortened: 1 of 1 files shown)


aviary/mission/flight_phase_builder.py
```diff
@@ -411,14 +417,29 @@ def make_default_transcription(self):
...
+            )
+
+        elif transcription_type == 'PicardShooting':
+            nodes_per_seg = order * num_segments  # get approximately same number of nodes as radau
+            transcription = dm.PicardShooting(
+                num_segments=1, nodes_per_seg=nodes_per_seg, solve_segments='forward'
+            )
+
+        else:
+            raise UserWarning(
```
Kenneth-T-Moore, line 439: I wonder if we should use an Enum for the transcription instead of a String.  We already use an Enum for SpeedTypes in the 2dof phase_infos, and I am adding one for phase_builders, so we have a precedent for it being an acceptable api for the user when writing a phase_info file.

aviary/mission/flight_phase_builder.py
```diff
@@ -411,14 +417,29 @@ def make_default_transcription(self):
...
+        transcription_type = user_options['transcription']
 
-        transcription = dm.Radau(
-            num_segments=num_segments,
-            order=order,
-            compressed=True,
-            segment_ends=seg_ends,
-        )
+        if transcription_type == 'Collocation':
+            seg_ends, _ = dm.utils.lgl.lgl(num_segments + 1)
```
Kenneth-T-Moore, line 423: Probably a question for Rob, but I wonder if this LGL (clustering points near the ends) is also supported in Picard shooting.

aviary/mission/flight_phase_builder.py
cmbenne3: I also wondered about this, but wasn't sure if we were allowing enums in the phase_info. If we already have that elsewhere in Aviary then I'll update this to an enum!

cmbenne3: Transcription is now converted to an Enum.

PR #936, opened 2025-12-10
Add a new var Aircraft.HorizontalTail.NUM_TAILS

xjjiang:
### Summary

Traditional aircraft have a horizontal tail but BWB aircraft may not have one and that may cause the system to fail. This PR adds a new option `Aircraft.HorizontalTail.NUM_TAILS` like the existing option `Aircraft.VerticalTail.NUM_TAILS`. 

Each characteristic lengths of wing, nacelle, canard, fuselage, horizontal tail, and vertical tail has its own component. A group component `CharacteristicLengths` is added to group them together.

All the current unit tests pass. There is a need to add unit tests for zero horizontal tail case. We will add them in when we work on BWB implementation.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/geometry/flops_based/characteristic_lengths.py
+++ b/aviary/subsystems/geometry/flops_based/characteristic_lengths.py
@@ -11,13 +11,9 @@ from aviary.variable_info.variables import Aircraft
 class CharacteristicLengths(om.Group):
     def initialize(self):
         add_aviary_option(self, Aircraft.Design.TYPE)
-        add_aviary_option(self, Aircraft.HorizontalTail.NUM_TAILS)
-        add_aviary_option(self, Aircraft.VerticalTail.NUM_TAILS)
 
     def setup(self):
         design_type = self.options[Aircraft.Design.TYPE]
-        num_horizontal_tails = self.options[Aircraft.HorizontalTail.NUM_TAILS]
-        num_vertical_tails = self.options[Aircraft.VerticalTail.NUM_TAILS]
 
         if design_type is AircraftTypes.BLENDED_WING_BODY:
             self.add_subsystem(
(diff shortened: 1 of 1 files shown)


aviary/subsystems/mass/flops_based/horizontal_tail.py
```diff
@@ -23,16 +26,20 @@ def setup_partials(self):
...
+
         area = inputs[Aircraft.HorizontalTail.AREA]
         gross_weight = inputs[Mission.Design.GROSS_MASS] * GRAV_ENGLISH_LBM
         scaler = inputs[Aircraft.HorizontalTail.MASS_SCALER]
         taper_ratio = inputs[Aircraft.HorizontalTail.TAPER_RATIO]
 
         outputs[Aircraft.HorizontalTail.MASS] = (
-            scaler * 0.530 * area * gross_weight**0.20 * (taper_ratio + 0.50) / GRAV_ENGLISH_LBM
-        )
+            scaler * 0.530 * area * gross_weight**0.20 * (taper_ratio + 0.50) * num_tails
```
jkirk5, line 37: The FLOPS method uses an exponent of 0.7 on number of tails, which means the original regression was built with this variable in mind. We don't have enough information to adjust the scalars used in this equation for number of horizontal tails correctly, so any mass provided for multiple tails will potentially be very off.

I don't want to modify the FLOPS method in this way now knowing it will be incorrect

jkirk5 (changes requested):
Having reviewed I don't think this is a good path forward for the FLOPS method - I really like all cleanup you did, such as for the characteristic lengths file, but adding additional H-tails is going outside the regression's bounds.

My suggestion is to remove `Horizontal_Tail.NUM_TAILS` and associated components from this PR. We should use Ken's suggestion to skip adding H-tail components when aircraft type is BWB. The FLOPS method doesn't support H-tail mass on the BWB so simply adding it on will also result in incorrect values. If users want an H-tail on their BWB they will need to provide better mass estimates as an external subsystem.

True horizontal tails on a BWB make it a hybrid-wing-body which these mass regressions don't model. Any significantly angled tails on the BWB are probably better modeled as vertical tails since at least the method attempts to account for structural integration when multiple are present.

xjjiang:
I will redo this task based on Jason's comment.

PR #940, opened 2025-12-16
Renamed the FLOPS fuselage AVG_DIAMETER to REF_DIAMETER to differentiate it from the GASP variable that is calculated differently.

### LINKED ISSUE #420: Misleading Fuselage Average Diameter

The fuselage average diameter variable is used by both FLOPS, GASP, and external codes (such as VSP). However, FLOPS uniquely does not define average diameter in a literal sense, and instead uses a non-physical diameter calculated using maximum fuselage height and width. This causes confusion and has lead to incorrect results for the TTBW model.

FLOPS-based analysis should probably use a new variable, Aircraft.Fuselage.REFERENCE_DIAMETER, with appropriate variable description, to avoid future confusion.

Also, Aircraft.Fuselage.AVG_DIAMETER should be expanded to Aircraft.Fuselage.AVERAGE_DIAMETER

_No response_

### Associated Bug Report

_No response_

Kenneth-T-Moore:
### Summary

Renamed the FLOPS fuselage AVG_DIAMETER to REF_DIAMETER to differentiate it from the GASP variable that is calculated differently. This is also one of the two variables that are calculated by both codes, so the CODE_ORIGIN_OVERRIDE now only contains the fuselage wetted area.

### Related Issues

- Resolves #420 

None

### New Dependencies

None


aviary/variable_info/variable_meta_data.py
```diff
@@ -3692,6 +3690,19 @@
...
+add_meta_data(
+    Aircraft.Fuselage.REF_DIAMETER,
+    meta_data=_MetaData,
+    historical_name={
+        'GASP': None,
+        'FLOPS': ['EDETIN.XD'],
+        'LEAPS1': 'aircraft.outputs.L0_fuselage.avg_diam',
+    },
+    units='ft',
+    desc='A course average diameter that is the mean of the width and depth.',
```
jkirk5:
```suggestion
    desc='A coarse average diameter calculated using the mean of max width and depth.',
```

Kenneth-T-Moore: Oops, gotta fix that.

PR #944, opened 2025-12-19
Mission directory renames.

Kenneth-T-Moore:
### Summary

1. Rename `mission/flops_based` to `mission/height_energy`.
2. Rename `mission/gasp_based` to `mission/two_dof`.
3. Gathered up the mission and ode files for solved 2dof and moved them into the new location `mission/solved_two_dof`.

This is the start of a refactor of the 2dof mission to make it work like the height energy mission. This PR is just moving things around and renaming them, so submitting it as its own PR  makes it easier to review.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/mission/solved_two_dof/phases/solved_twodof_phase.py
```diff
@@ -288,14 +288,16 @@ def _extra_ode_init_kwargs(self):
         }
 
 
-TwoDOFPhase._add_initial_guess_meta_data(
```
ehariton, line 291: this is a bit of a subtle change
"TwoDOFPhase -> SolvedTwoDOFPhase"

aviary/mission/two_dof/phases/test/__init__.py
ehariton, line 1: why did you add this as an empty file?

Kenneth-T-Moore: A file named __init__.py is required at any level if you want to be able to import modules from that directory. I usually add them in the test directories too for consistency.

aviary/mission/solved_two_dof/phases/solved_twodof_phase.py
Kenneth-T-Moore: Yeah, this was to alleviate our confusion, for every time we looked at this file and thought it was used in the regular 2dof mission.

jkirk5: I've hit that dead end multiple times trying to track where all the 2DOF stuff gets added (controls, states, constraints, etc.) and then realizing my breakpoints never get hit because this is for Solved only

aviary/mission/two_dof/phases/test/__init__.py
ehariton: ahh.  it's been so long since I added python folders :)

ehariton (approved):

PR #946, opened 2025-12-23
Remove requirement that engines compute max thrust and other values

### LINKED ISSUE #493: Automate handling of max thrust/SHP condition in propulsion

Currently engines are responsible for providing analysis to determine what their maximum possible thrust output for a given flight condition is, for the purpose of satisfying max excess power constraints. This results in a lot of parallel code that must be set up for every new `EngineModel`, and is tedious and difficult to set up.

Instead, new code at the `PropulsionMission` level is needed to set up a second copy of the engine models, passing them a set of max throttle parameters instead. This may potentially replace hundreds of lines of code with a few dozen, and will be much cleaner to use going forward.

_No response_

### Associated Bug Report

_No response_

jkirk5:
### Summary

Maximum thrust at a given flight condition is required to compute excess specific power for mission. Previously, EngineModels were required to output max thrust, which is a large overhead for users and difficult to clearly communicate through documentation and the API.

The propulsion subsystem was upgraded to check if an EngineModel provides max values (via a new flag, `self.compute_max_values`, defaulted to False), and if not creates a duplicate copy of the mission components that are fed a throttle & hybrid throttle of 1 instead. No outputs are promoted from this duplicate engine to avoid promotion/connection issues. Thrust outputted by the duplicate engine is directly connected to the engine mux component's max thrust input.

Cleaned up some pieces of propulsion code including formatting, bugfixes, TODO's, etc. as I found them and was able to resolve on the spot.

Disclosure: Generative AI was used to draft coding approaches to handle the parallel engine groups, I ended up not using its suggestions but I did keep a few minor fixes it tried. Code in `propulsion_mission.py`: lines 6 and 49 (plus indenting the lines I already wrote immediately following that line) were added by AI. The Cursor IDE was used, which does not disclose which LLM models it uses.

### Related Issues

- Resolves #493

EngineModels that already compute max thrust can continue to do so as long as they add the `compute_max_values` flag as `True` in their models. If they don't, their component will get duplicated, which theoretically shouldn't break anything but will be very inefficient.

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/propulsion/engine_deck.py
+++ b/aviary/subsystems/propulsion/engine_deck.py
@@ -1102,14 +1102,13 @@ class EngineDeck(EngineModel):
             if self.global_throttle or (self.global_hybrid_throttle and self.use_hybrid_throttle):
                 engine_group.add_subsystem(
                     'fixed_max_throttles',
-                    fixed_throttles,  # , promotes_outputs=['*']
+                    fixed_throttles,
                 )
             else:
                 engine_group.add_subsystem(
                     'interp_max_throttles',
                     interp_throttles,
                     promotes_inputs=['*'],
-                    # promotes_outputs=['*'],
                 )
 
             engine_group.add_subsystem(
@@ -1118,7 +1117,7 @@ class EngineDeck(EngineModel):
                 promotes_inputs=[
                     Dynamic.Atmosphere.MACH,
                     Dynamic.Mission.ALTITUDE,
-                ],  # , promotes_inputs=['*']
+                ],
             )
 
             # manually connect max throttles - do not promote them out of group
--- a/aviary/subsystems/propulsion/turboprop_model.py
+++ b/aviary/subsystems/propulsion/turboprop_model.py
@@ -330,7 +330,6 @@ class TurbopropMission(om.Group):
         except (AttributeError, KeyError):
             propeller_kwargs = {}
 
-        # propeller_group = om.Group()
         propeller_model_mission = propeller_model.build_mission(
             num_nodes, aviary_inputs, **propeller_kwargs
         )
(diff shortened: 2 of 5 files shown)


jkirk5:
I am currently testing out removing max value computation from the turboprop model. This is pretty tricky and may have exposed some broken edge cases in the current implementation. Please do not merge yet!!

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1085,23 +1091,38 @@ def build_mission(self, num_nodes, aviary_inputs, **kwargs) -> om.Group:
         if self.use_thrust or self.use_shaft_power:
             if self.global_throttle or (self.global_hybrid_throttle and self.use_hybrid_throttle):
                 engine_group.add_subsystem(
-                    'fixed_max_throttles', fixed_throttles, promotes_outputs=['*']
+                    'fixed_max_throttles',
+                    fixed_throttles,  # , promotes_outputs=['*']
```
Kenneth-T-Moore: Can remove all of the commented out promotes statements.

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -72,27 +71,43 @@ def __init__(
...
+        else:
+            gearbox_solver = False
+        if self.propeller_model is not None:
+            prop_solver = self.propeller_model.needs_mission_solver()
+        else:
+            prop_solver = False
+        mission_solver = np.any([shp_solver, gearbox_solver, prop_solver])
+        return mission_solver
+
+    # BUG if using multiple custom subsystems that happen to share a kwarg but need different values,
```
Kenneth-T-Moore, line 101: Would like to hear more about this bug to understand what is going on.

jkirk5: In `phase_info` you currently pass all model options for the turboprop in a single dict, which gets offered to each subcomponent of the turboprop:
`{'turboprop_name': {'option_1': 'val', 'option_2': 'val'}}`

Let's say both the motor and the propeller both need `option_1`, but want different values - the current code doesn't support that. A potential solution is to add support for nested kwargs for turboprop, so:
`{'turboprop_name': {'motor_name': {'option_1': 'val_1'}, 'propeller_name': {'option_1': 'val_2'}, 'option_2': 'val'}}`

PR #959, opened 2026-01-12
Address problems found during OpenMDAO connection graph refactor.

Kenneth-T-Moore:
### Summary

The OpenMDAO team has refactored how connections are managed and stored internally, moving over to a full graph-based approach. This has revealed many corner cases that the old system was letting slide. These fall under two categories:
1. Multiple inputs promoted together with different units or values: OpenMDAO caught many of these, but sometimes it would ignore the ambiguity and just take the first value/unit that it found.
2. There were some weird cases where an input connected to an upstream output could be promoted and connected to another output, or multiple input-output connections could be promoted together.

This PR fixes any of those found in Aviary.

1. In two_dof, the initial time for ascent was connected to the final time of groundroll, but was also promoted as a design variable.
  - Fixed this by creating a slack design variable and adding a constraint, which was the intent.

2. In two_dof pre_mission, flaps ratios were computed with FlapsGroup for 3 different deflection cases. These groups also computed some basic flaps geometry parameters, which would be the same for all three of the cases, but they got promoted together, which broke the connection to one of them with a downstream component that computed mass.
  - Fixed this by moving the static geo calculations out of the FlapsGroup and just doing them once.
  - Some test truth values have changed now that mass is being calculated with the correct value (instead of a default value of 0).

3. In the two_dof landing group, there were several places that the altitude was not being correctly promoted to the airport altitude. We never noticed because we usually set it to 0 ft.

4. In a few places, we needed to add more calls to set_input_defaults to intermediate groups.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/subsystems/aerodynamics/gasp_based/premission_aero.py
```diff
@@ -44,7 +47,11 @@ def setup(self):
         # to avoid multiple sources for computed Mach (gets calculated somewhere upstream)
```
jkirk5, line 47: Something went wrong here... if input_speed_type is MACH then mach should never get computed again - it's supposed to become an input to the flight conditions component.

I don't want flight conditions to be skippable, that is very dangerous. We need to make sure that all speed types are available at all times, removing it means that downstream components can suddenly break, for example because they want TAS but only Mach was provided in the problem. This is a weird case because it is premission for a single subsystem so we are in full control of what is downstream, but I don't want this option to be available. And it shouldn't be needed at all, the fact that this broke means something is wrong with the flight conditions component!

aviary/utils/preprocessors.py
```diff
@@ -595,9 +595,12 @@ def preprocess_propulsion(
 
             # If dtype has multiple options, prefer type of default value
             # Otherwise, use the first type in the tuple
-            if isinstance(dtype, tuple):
+            if isinstance(dtype, (tuple, list)):
```
jkirk5: I doubled checked and we don't actually enforce any type for how we want metadata to communicate multiple type options. We should probably pick one and actually enforce it in `add_meta_data()`. Because this opens up a can of worms - what if it is a numpy array? Or a set? Or any other iterable? We aren't covering those here either. OM options `types` requests a tuple, which I am fine with enforcing in Aviary too.

As a side note I couldn't find any variables using a list for multiple acceptable types, only a tuple. Was this a preemptive change or was there some specific place this was happening? I couldn't find it manually searching

Kenneth-T-Moore: My bad. We only need to check lists on the default_value, not the types. I've removed it from there.

aviary/subsystems/aerodynamics/gasp_based/premission_aero.py
Kenneth-T-Moore: I took a deeper look at this, and it might be another case of outputs being promoted together. Mach number shouldn't be promoted out of the flaps groups because it needs different values for each flaps case, so it ultimately shouldn't interfere with the promoted mach input from atmosphere. I will see if i can fix this by removing the '*' from promotes_inputs for those groups, and making them manual.

Kenneth-T-Moore: This worked, though it changed some values in gasp tests. The new numbers should be more correct because the 3 flaps cases are allowed to have different mach number now, like they should have all along.

Kenneth-T-Moore: Actually, no numeric changes at all.

PR #965, opened 2026-01-15
Updated some unittests to use subtests

jkirk5:
### Summary

A subset of unittests were updated to use unittest's subTest feature, which allows for multiple failures to occur within a test, which prevents a developer from needing to update a test one assert at a time while debugging.

This work was done with AI (Claude 4.5 opus), and was a test of automating simple but tedious updates across multiple files. Only some files were updated due to limits/cost on repeatedly running an expensive LLM model and limitations on how many files in a row an agentic query could get through before stalling.

Updates to the FLOPS validation test helper functions was done manually.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/validation_cases/validation_tests.py
```diff
@@ -140,6 +130,7 @@ def do_validation_test(
 
 
 def flops_validation_test(
+    test: unittest.TestCase,
```
Kenneth-T-Moore, line 133: `case_name` should probably be removed from here too, since it is not longer used due to the change in arguments in do_validation_test.

jkirk5: This function actually does use `case_name`, it is only `do_validation_test` that doesn't ever use that argument

Kenneth-T-Moore (approved):

PR #968, opened 2026-01-18
Fix #953: Initialize self.subsystems for AviaryGroup

### LINKED ISSUE #953: run_level3_example.py fails in subsystem_report
### Description

Running aviary/examples/run_level3_example.py results in an error in the subsystem_report function:

```
  File "aviary/interface/reports.py", line 178, in subsystem_report
    subsystems = model.subsystems  # TODO: redo for multimissions
                 ^^^^^^^^^^^^^^^^
AttributeError: 'AviaryGroup' object has no attribute 'subsystems'
```

### Example

python ./aviary/examples/run_level3_example.py

### Aviary Version

 0.10.0.dev0

_No response_

OwenShepherd:
### Summary

Initialize self.subsystems to an empty list in the AviaryGroup __init__. Otherwise, self.subsystems isn't initialized until a successful run of check_and_preprocess_inputs() is made. This also allows full reports to be made for the level3 example without crashing. Prior, the level3 example would crash before writing the full report, and fail to write these files:
- status.json
- mission_summary.md
- mission_timeseries_data.csv
- traj_results_report.html

### Related Issues

- Resolves #953 

None

### New Dependencies

None

### DIFF
--- a/aviary/core/aviary_group.py
+++ b/aviary/core/aviary_group.py
@@ -69,6 +69,7 @@ class AviaryGroup(om.Group):
         self.engine_models = []
         self.regular_phases = []
         self.reserve_phases = []
+        self.subsystems = []
 
         self.aviary_inputs = None
         self.meta_data = None


jkirk5 (approved):
Good catch, we should definitely have been initializing that! Thanks for the contribution!!

PR #975, opened 2026-01-23
Fix needed for the NPSS example

Kenneth-T-Moore:
### Summary

NPSS example wasn't running because of a missing variable in the aircraft csv file.

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_FLOPS.csv
```diff
@@ -46,6 +46,7 @@ aircraft:engine:fuel_flow_scaler_linear_term,1,unitless
 aircraft:engine:generate_flight_idle,True,unitless
 aircraft:engine:geopotential_alt,False,unitless
 aircraft:engine:ignore_negative_thrust,False,unitless
+aircraft:engine:interpolation_method,slinear,unitless
```
cmbenne3, line 49: @Kenneth-T-Moore - Perhaps we could update the variable_meta_data.py with a link to the openMDAO page documentation so that someone can easily work out what methods are available.

Image
I think this is the doc link: https://openmdao.org/newdocs/versions/latest/features/building_blocks/components/metamodelstructured_comp.html

Does OpenMDAO have enums that we can import? Might get a bit messy if we make this an Aviary enum (in case OpenMDAO makes updates) so I think best to keep as a string for now - what do you think?

Kenneth-T-Moore: I've kind of been wanting to modify OpenMDAO to use enums for all of the interpolation methods.

Kenneth-T-Moore:
@cmbenne3
I've extended the description for the interp method in the metadata to include a url and a bit more information.

PR #983, opened 2026-02-05
Name change for output directories of off design problems.

### LINKED ISSUE #957: Multiple mission runs and reports directory
### Description

When running `aviary/examples/run_off_design_example.py` there are some issues with the reports folders. 

1. Since there are 3 missions runs in the script, there should be 3 reports folders but there are only 2. The third mission run overwrites the directory of the second mission run
2. After the second mission run, the directory has a full set of reports. But after the third run, it has a greatly reduced set of reports. 

The normal set of reports is 

```driver_scaling_report.html
input_checks.md
inputs.html
mission_summary.md
mission_timeseries_data.csv
n2.html
opt_report.html
overridden_variables.md
status.json
subsystems
total_coloring.html
traj_linkage_report.html
traj_results_report.html
```

but after the third run, only get

```
driver_scaling_report.html
n2.html
total_coloring.html
traj_results_report.html
```

### Example

Run

`python aviary/examples/run_off_design_example.py`

### Aviary Version

0.10.0.dev0

_No response_

OwenShepherd:
### Summary

This adds logic and a test to essentially do the following to resolve issue #957 and follow up @jkirk5's comment on PR #981:

If your problem is named 'test', the baseline output directory will be named 'test_out', and if you run off design problem types (say, fallout, alternate, etc) without a name the problem type and the time the problem was run will be added to its name to produce the following directory outputs:
- test_fallout_020526182200_out
- test_alternate_020526182223_out
- and so on.

### Related Issues

- Resolves #957 
- Related to #981 

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -1367,7 +1367,7 @@ class AviaryProblem(om.Problem):
 
         if name is None:
             name = self._name + '_' + str(problem_type.value)
-            curr_time = datetime.now().strftime("%m%d%y%H%M%S")
+            curr_time = datetime.now().strftime('%m%d%y%H%M%S')
             name += '_' + curr_time
         off_design_prob = AviaryProblem(name=name)
 
--- a/aviary/interface/methods_for_level2.py
+++ b/aviary/interface/methods_for_level2.py
@@ -1366,12 +1366,12 @@ class AviaryProblem(om.Problem):
             )
 
         if name is None:
-            suggestion = self._name + "_" + str(problem_type.value)
-            # For while loop / numbering names
+            # Numbering problem names.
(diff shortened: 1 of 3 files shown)


jkirk5 (commented):
Thanks Owen! This is basically what I was planning on (but never got around to). I am concerned that the folders become this hard-to-parse string of numbers that will seem like random/cryptic code to new/basic users. The only reason we care about exact seconds it in case someone runs a lot of quick cases back-to-back, but other than that I don't think it matters much.

This is probably be good to merge, since folder w/ complex name is better than overwriting the previous folder. I am thinking about ways to streamline, I'll post those ideas in the issue so they are easier to find.

jkirk5:
This issue should be resolved via another PR #981 so I will close this PR, and if there are still issues with the filenames a new issue can be created

PR #986, opened 2026-02-11
Two-dof refactor part 3: Added a new ODE/Phase for a simple cruise that is more consistent with the other phases.

Kenneth-T-Moore:
### Summary

Motivation: The original Cruise phase in the two-dof equations of motion is based on the Breguet Range equations, and uses an AnalyticPhase from Dymos, and additionally integrates over mass instead of time. This causes problems when trying to add an external subsystem with a state that needs to be integrated over time. This PR adds an alternative that eliminates these problems.

Added a new SimpleCruise eom, ode, and phase to the two-dof formulation.
- Has 1 dynamic state: mass
- Integrates in time, so initial time and duration are also controlled by the optimizer.
- Has no controls: throttle is solved, and altitude and velocity are fixed. Distance is a simple calculation from those.

Added a new option "phase_builder" to the phase_options to allow the user to specify what phase_builder to use for this phase. The following choices are valid:
```
    DEFAULT: Use the default phase builder for this EquationsOfMotion.

    BREGUET_RANGE: Use a phase builder that implements the Breguet Range equations.

    SIMPLE_CRUISE: Use a phase builder that implements a single DOF (mass) cruise.
```
Presently, this option only works for the two-dof equations of motion, but there is a potential to expand the capability to use these phases in other equations.

- The old Cruise ode/eom/phase have all been renamed to include Breguet in their names.
- All tests and benchmarks that used the old Cruise phase have been switched over to the new SimpleCruisePhase.
- One new benchmark has been added for the BreguetCruisePhase

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/docs/user_guide/reserve_missions.ipynb
+++ b/aviary/docs/user_guide/reserve_missions.ipynb
@@ -93,7 +93,7 @@
     "        phases.append(getattr(module, phase_name + 'Phase'))\n",
     "\n",
     "for phase in phases:\n",
-    "    phase_name = phase().__class__.__name__\n",
+    "    phase_name = phase(subsystems=[]).__class__.__name__\n",
     "    if 'Groundroll' in phase_name:\n",
     "        continue  # no reserve groundroll\n",
     "    check_contains(\n",


aviary/mission/two_dof/ode/simple_cruise_ode.py
```diff
@@ -0,0 +1,148 @@
...
+        kwargs = {
+            'num_nodes': nn,
+            'aviary_inputs': aviary_options,
+            'method': 'cruise',
+            'output_alpha': True,
+        }
+        for subsystem in subsystems:
+            # check if subsystem_options has entry for a subsystem of this name
+            if subsystem.name in subsystem_options:
+                kwargs.update(subsystem_options[subsystem.name])
```
jkirk5: I think we need to update how we handle the kwargs. Some things like num_nodes, aviary_options could be provided directly as arguments to `build_mission()`, and we can check for `AerodynamicsBuilder`s before getting the subsystem-specific kwargs to make sure 'method' and 'output_alpha' are defaulted to our desired values.

This prevents kwargs from getting more cluttered with irrelevant options as we loop through subsystems and makes sure the desired aero defaults don't get overwritten if an earlier subsystem happens to use an identically named option

jkirk5: This applies to all the 2DOF ODEs but should be able to be quickly copy-pasted to them all

aviary/mission/two_dof/ode/simple_cruise_ode.py
```diff
@@ -0,0 +1,148 @@
...
+                        promotes_inputs=subsystem.mission_inputs(**kwargs),
+                        promotes_outputs=subsystem.mission_outputs(**kwargs),
+                    )
+                else:
+                    self.add_subsystem(
+                        subsystem.name,
+                        system,
+                        promotes_inputs=subsystem.mission_inputs(**kwargs),
+                        promotes_outputs=subsystem.mission_outputs(**kwargs),
+                    )
```
jkirk5: I took a quick crack at what a revised kwargs handling could look like
```suggestion
        for subsystem in subsystems:
            kwargs = {}
            # check if subsystem_options has entry for a subsystem of this name
            if subsystem.name in subsystem_options:
                kwargs = subsystem_options[subsystem.name]
                if isinstance(subsystem, AerodynamicsBuilder):
                    # set default options for Aero if not specified by user
                    base_kwargs = {
                        'method': 'cruise',
                        'output_alpha': True
                    }
                    kwargs = base_kwargs.update(kwargs)
            system = subsystem.build_mission(num_nodes=nn, aviary_inputs=aviary_options, **kwargs)
            if system is not None:
                if isinstance(subsystem, PropulsionBuilder):
                    prop_group.add_subsystem(
                        subsystem.name,
                        system,
                        promotes_inputs=subsystem.mission_inputs(**kwargs),
                        promotes_outputs=subsystem.mission_outputs(**kwargs),
                    )
                else:
                    self.add_subsystem(
                        subsystem.name,
                        system,
                        promotes_inputs=subsystem.mission_inputs(**kwargs),
                        promotes_outputs=subsystem.mission_outputs(**kwargs),
                    )
```

aviary/mission/two_dof/phases/breguet_cruise_phase.py
```diff
@@ -100,7 +93,12 @@ def __init__(
         transcription=None,
         subsystems=None,
         meta_data=None,
+        is_analytic_phase=True,
     ):
+        if is_analytic_phase is not True:
+            msg = 'The Breguet Cruise phase does not support dynamic variables in its subsystems.'
+            raise AttributeError(msg)
+
```
jkirk5: Why even provide the option to try making Breguet cruise not analytic?
Maybe something I should have noticed earlier... does this ever need to be defined when initializing any `PhaseBuilder`? The user will typically never be able to toggle this without breaking everything

Kenneth-T-Moore: The `is_analytic_phase` gets toggled to False if one of the subsystems has a dynamic phase. I don't think the user can ever set this flag.  I could probably rename it to something like `_subsystems_have states`.

EDIT - I am going to rework this, and check the subsystems directly in Breguet Range instead of in the base class.

aviary/mission/two_dof/ode/simple_cruise_ode.py
Kenneth-T-Moore: Good point. This should be easy to clean up.

EDIT - Updated Simple Cruise, Breguet Cruise, and Flight. The other ones got deleted and replaced in part 4, so I will continue this there.

Kenneth-T-Moore:
Ok, rework done. Much cleaner now. Also added a test for that error message.

aviary/mission/two_dof/phases/breguet_cruise_phase.py
jkirk5: Ok! I don't fully know the path for how this gets set up in an actual AviaryProblem off the top of my head so I wanted to make sure it still made sense.

It seems like this could be an attribute (default value set per specific PhaseBuilder) rather than an argument.

Kenneth-T-Moore: > Ok! I don't fully know the path for how this gets set up in an actual AviaryProblem off the top of my head so I wanted to make sure it still made sense.
> 
> It seems like this could be an attribute (default value set per specific PhaseBuilder) rather than an argument.

I ended up removing the changes I made that changed the `is_analytic_phase` based on whether the subsystems had states, so the original behavior is restored.  Instead, I added a check in the breguet phase to raise an error if there are states in the external subsystems.

PR #999, opened 2026-02-18
Rounding of numbers to 3 significant digits in reports and dashboard

### LINKED ISSUE #237: Round outputs based on number of sig figs?

Tom O., an Aviary user, suggested rounding (or having an option to round) the numbers outputted from Aviary. He doesn't care about the 15th digit and cares only maybe about the third digit, particularly because the input info might only go to a few sig figs.

@crecine suggested that we could use the `round_it` function that's already in Aviary in more places accordingly.

This is not a high priority task and might not even need to be accomplished. If we're able to smoothly allow users to choose to round results or not, that's probably a good idea. There are also a number of places where we output numbers from Aviary; should we offer options to run in all places? Thinking Aviary dashboard, mission reports, timeseries data, etc.

_No response_

### Associated Bug Report

_No response_

hschilling:
### Summary

Several reports and dashboard windows showed values with an excessive number of significant digits. This PR reduces those to 3 significant digits. 

### Related Issues

- Resolves #237 

None

### New Dependencies

None

### DIFF
--- a/aviary/interface/reports.py
+++ b/aviary/interface/reports.py
@@ -679,7 +679,7 @@ def _overridden_variables_group_report(prob, group, mission_name, f):
             if isinstance(val, np.ndarray):
                 valstring = np.array2string(val, formatter={'float_kind': lambda x: f'{x:.3g}'})
             else:
-                valstring = f"{val:.3g}"
+                valstring = f'{val:.3g}'
             f.write(f'| {aircraft_variable_name} | {valstring} | {units} |\n')
         f.write('\n')
     else:


aviary/interface/reports.py
```diff
@@ -676,7 +676,8 @@ def _overridden_variables_group_report(prob, group, mission_name, f):
         f.write('| ------ |-------| ----- |\n')
         non_external_overridden_variables.sort(key=lambda x: x[0])
         for aircraft_variable_name, val, units in non_external_overridden_variables:
-            f.write(f'| {aircraft_variable_name} | {val} | {units} |\n')
+            valstring = np.array2string(val, formatter={'float_kind': lambda x: f'{x:.3g}'})
```
Kenneth-T-Moore: Some tests are failing because this val (which comes from prob.get_val) can sometimes be a float, so array2string doesn't work on it. Might need to turn it into an array first.

PR #1005, opened 2026-02-24
Fix for variable bus from pre- to post-mission subsystem

irianordaz:
### Summary

Fixes the subsystem builder bus to allow passing variables from pre-mission to post-mission. Previously, the connections were only made when the parameter contained a target in "mission". If the parameter contained a target only in post_mission, an error would occur.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/subsystems/test/test_external_subsystem_bus.py
+++ b/aviary/subsystems/test/test_external_subsystem_bus.py
@@ -47,12 +47,14 @@ class PreMissionComp(om.ExplicitComponent):
         self.add_output('for_climb', np.ones(shape), units='ft')
         self.add_output('for_cruise', np.ones(shape), units='ft')
         self.add_output('for_descent', np.ones(shape), units='ft')
+        self.add_output('for_post', np.ones((2, 3)), units='rad')
 
     def compute(self, inputs, outputs):
         shape = self.options['shape']
         outputs['for_climb'] = np.random.random(shape)
         outputs['for_cruise'] = np.random.random(shape)
         outputs['for_descent'] = np.random.random(shape)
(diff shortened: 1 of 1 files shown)


aviary/core/aviary_group.py
```diff
@@ -1152,74 +1152,83 @@ def _add_bus_variables_and_connect(self):
...
+
+                            if '.' in post_mission_var_name:
+                                # Support for non-hierarchy variables as parameters.
+                                post_mission_var_name = post_mission_var_name.split('.')[-1]
 
-                        for post_mission_var_name in post_mission_variable_name:
                             self.connect(
-                                f'pre_mission.{bus_variable}',
-                                post_mission_var_name,
+                                f'pre_mission.{subsystem.name}.{bus_variable}',
```
Kenneth-T-Moore: The change in these two lines needs to be reverted.

aviary/core/aviary_group.py
```diff
@@ -1152,74 +1152,83 @@ def _add_bus_variables_and_connect(self):
...
-                            post_mission_variable_name = [post_mission_variable_name]
+                        post_mission_var_names = variable_data['post_mission_name']
+                        src_indices = variable_data.get('src_indices', None)
+
+                        if not isinstance(post_mission_var_names, list):
+                            post_mission_var_names = [post_mission_var_names]
+
+                        for post_mission_var_name in post_mission_var_names:
+
+                            if '.' in post_mission_var_name:
```
Kenneth-T-Moore: I think this split here will break if the user wants to connect to an unpromoted input like `post_mission.sub1.sub2.comp.x`. We should probably remove it, and change the target in the connect back to post_mission_var_name.

PR #1009, opened 2026-02-25
Unify the API for subsystem builder methods.

### LINKED ISSUE #347: Refactor subsystem builder methods that can be properties
Multiple different methods in `SubsystemBuilderBase` could be properties to simplify their use. This is backwards incompatible though generally worthwhile; could deprecate the existing methods.

```suggestion
    @property
    def get_mass_names(self):
```
so it can be accessed as `BatteryBuilder.get_mass_names` instead of `BatteryBuilder.get_mass_names()`, since no arguments are used

_Originally posted by @crecine in https://github.com/OpenMDAO/Aviary/pull/317#discussion_r1651562417_

Kenneth-T-Moore:
### Summary

This PR expands the arguments to many of the subsystem builder methods to include aviary_inputs, phase_info, and the phase_name where they were applicable.

- Removed the `define_order` method from the phase_builder API. It never did anything, and that capability is a little tricky to do this way.
- Removed mission/phase_utils.py and moved the functionality into the phasebuilder base class.

### Related Issues

- Resolves #347

Users will have to edit their subsystems to add the missing arguments.
In addition, an error will be raised if a subsystem builder contains the `get_bus_variables` method, which was renamed in an earlier PR to `get_pre_mission_bus_variables`.

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -805,7 +799,7 @@ def add_post_mission_systems(self, verbosity=None):
         for subsystem in self.subsystems:
             subsystem_postmission = subsystem.build_post_mission(
                 aviary_inputs=self.aviary_inputs,
-                phase_info=self.mission_info,
+                mission_info=self.mission_info,
```
jkirk5, line 816: I though this would get the entire phase_info dict?
`self.pre_mission_info`: subdict with key 'pre_mission'
`self.mission_info`: subdict containing all keys not 'pre_mission' or 'post_mission'
`self.post_mission_info`: subdict with key 'post_mission'

jkirk5: It looks like we aren't actually saving the phase_info dict in its entirety as an attribute of the `AviaryGroup` anywhere

aviary/core/aviary_group.py
```diff
@@ -1228,7 +1224,7 @@ def _connect_mission_bus_variables(self):
         # Loop through all external subsystems.
         for subsystem in all_subsystems:
             for phase_name, var_mapping in subsystem.get_post_mission_bus_variables(
-                aviary_inputs=self.aviary_inputs, phase_info=self.mission_info
+                aviary_inputs=self.aviary_inputs, mission_info=self.mission_info
```
jkirk5, line 1232: Another place where we are not passing the whole phase_info

aviary/subsystems/propulsion/turboprop_model.py
```diff
@@ -133,19 +142,19 @@ def build_pre_mission(self, aviary_inputs, **kwargs) -> om.Group:
 
         return turboprop_group
 
-    def build_mission(self, num_nodes, aviary_inputs, **kwargs):
+    def build_mission(self, num_nodes, aviary_inputs, subsystem_options):
```
jkirk5: I think now `subsystem_options` is only giving the subdict whose key matches the name of the `TurbopropModel`? As opposed to before, where it got the whole subsystems_options subdict?

If so, then this is a big problem that breaks turboprop and I need to make a new issue to address.

aviary/subsystems/subsystem_builder.py
```diff
@@ -325,14 +384,28 @@ def get_outputs(self):
...
+
+        Returns
+        -------
+        list
+            List of outputs to add to the timeseries.
         """
         return []
 
-    def get_post_mission_bus_variables(self, aviary_inputs=None, phase_info=None):
+    def get_post_mission_bus_variables(self, aviary_inputs=None, mission_info=None):
```
jkirk5, line 420: Should this get the entire phase_info? I thought it might need `post_mission` too

aviary/subsystems/subsystem_builder.py
```diff
@@ -383,7 +456,7 @@ def get_post_mission_bus_variables(self, aviary_inputs=None, phase_info=None):
         return {}
 
     def build_post_mission(
-        self, aviary_inputs, phase_info=None, phase_mission_bus_lengths=None, **kwargs
+        self, aviary_inputs=None, mission_info=None, phase_mission_bus_lengths=None
```
jkirk5: I'm pretty confident we need to give this one the entire `phase_info` dict so it can see inside `post_mission`

aviary/subsystems/subsystem_builder.py
jkirk5, line 1: Just double checking that `get_pre_mission_bus_variables()` won't ever have a use for `phase_info`? I don't know the nuances of it vs. post-mission bus which does need it

jkirk5 (changes requested):

aviary/subsystems/subsystem_builder.py
```diff
@@ -222,47 +261,55 @@ def build_mission(self, num_nodes, aviary_inputs, **kwargs):
...
+        return ['*']
 
-    def get_design_vars(self):
+    def get_design_vars(self, aviary_inputs=None):
         """
         Return a dictionary of design variables for the subsystem.
 
-        Optional
-
         Not currently used.
```
jkirk5: This definitely does get used now, so we should remove this too!

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: Yeah, I think mission_info should be passed into here as well. I must have missed this one.

aviary/subsystems/propulsion/turboprop_model.py
Kenneth-T-Moore: Discussed this with Jason and we have a workaround on the user-specification side.

aviary/core/aviary_group.py
Kenneth-T-Moore: After discussion, I've added a subsystem_options to the build pre/post mission methods.

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: Removed this.

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: Added mission_info to the args.

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: added mission_info to the args.

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: added mission_info to the args.

aviary/core/aviary_group.py
Kenneth-T-Moore: mission_info is now passed.

aviary/subsystems/subsystem_builder.py
```diff
@@ -222,46 +270,52 @@ def build_mission(self, num_nodes, aviary_inputs, **kwargs):
...
+
+        Returns
+        -------
+        list
+            List of all mission inputs.
         """
         return ['*']
 
-    def mission_outputs(self, **kwargs):
+    def mission_outputs(self, aviary_inputs=None, subsystem_options=None):
```
jkirk5: Missing user_options?

aviary/subsystems/subsystem_builder.py
```diff
@@ -222,46 +270,52 @@ def build_mission(self, num_nodes, aviary_inputs, **kwargs):
         """
         return None
 
-    def mission_inputs(self, **kwargs):
+    def mission_inputs(self, aviary_inputs=None, subsystem_options=None):
```
jkirk5: missing user_options?

aviary/subsystems/subsystem_builder.py
```diff
@@ -205,14 +244,23 @@ def get_pre_mission_bus_variables(self, aviary_inputs=None):
         """
         return {}
 
-    def build_mission(self, num_nodes, aviary_inputs, **kwargs):
+    def build_mission(self, num_nodes, aviary_inputs, subsystem_options):
```
jkirk5: missing user_options?

aviary/subsystems/subsystem_builder.py
jkirk5: I see mission_info is provided, but I was asking about if post_mission was needed too!

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: Added the missing arg to all 3 of these methods.

PR #1013, opened 2026-03-02
Fix units for dynamic viscosity

irianordaz:
### Summary

Fix units for dynamic viscosity

### Related Issues

- Resolves #

None

### New Dependencies

None


jkirk5 (commented):
Should this be lbm*s/ft so gravitational force is not intrinsically part of the value?

PR #1027, opened 2026-03-16
Fuel Variable Accounting - Issue 1012

### LINKED ISSUE #1012: Too many identical (?) fuel variables

There are many different "total" fuel variables that sound like they should be identical (based on name and metadata description), but I am pretty sure they are not all identical but may be getting used interchangeably in various places in the code.

1) We need to get a list of which specific fuel variables we actually need, and decide on good names and useful descriptions for them
2) We need to go through the codebase and update components to use the actually correct fuel mass variables

A list of fuel mass variables I found that drove this initial confusion:
- `Mission.Summary.FUEL_BURNED`
- `Mission.Summary.FUEL_MASS`
- `Mission.Summary.FUEL_MASS_REQUIRED`
- `Mission.Summary.TOTAL_FUEL_MASS`
- `Aircraft.Fuel.TOTAL_CAPACITY`

Which ones contain unusable fuel and which do not is a key point of confusion. We probably only need 3? Tank total capacity (`Aircraft` value, fixed for sized aircraft), mass of fuel currently on the aircraft during the mission (decreases over time), and the total mass of fuel on the aircraft at mission start (does not change with time, fixed value) - because an aircraft can carry more fuel than it actually needs for the mission

_No response_

### Associated Bug Report

_No response_

ehariton:
## Summary

This PR includes a revamp and simplification of fuel mass variables to prevent users from incorrectly selecting variables that don't actually represent mission fuel burn. This included fixes to GASP to Aviary file conversion. As well as an additional doc page on how fuel is summed up. Also fixed some reports that were showing incorrect fuel burn values when takeoff was enabled.

### Name and Reference Changes
Renamed for clarity
`Mission.fuel.FUEL_MARGIN` -> `Mission.Fuel.VOLUME_MARGIN` this is a fuel margin and has little to do with fuel mass
`Aircraft.Design.RESERVE_FUEL_FRACTION` -> `Aircraft.Design.RESERVE_FUEL_MARGIN` we are tracking this as a % now

Removed from aviary variable hierarchy because they were internal variables in GASP and did not need to be set by the user:
`Mission.Summary.FUEL_MASS`
`Mission.Summary.FUEL_MASS_REQUIRED`

Updated OAS builder tests to refer to `Aircraft.Fuel.WING_FUEL_CAPACITY` rather than `Mission.Summary.FUEL_MASS` which was the mass of ALL fuel for the mission including in the fuselage and aux tanks!!

The initial mass used to calculate `Mission.Summary.FUEL_BURNED` is now always `Mission.Summary.GROSS_MASS` whereas before it was sometimes this and sometimes the first point in the first dymos phase.

### Calculations Updates
Changed the way `Mission.Summary.TOTAL_FUEL_MASS` is calculated to remove incorrect addition of  `Mission.fuel.FUEL_MARGIN`.

### Updates to GASP to Aviary conversion factors
INGASP.FVOL_MRG -> `Mission.Fuel.VOLUME_MARGIN`
Depending on the value for `FRESF`:
INGASP.FRESF -> `Aircraft.Design.RESERVE_FUEL_ADDITIONAL`
INGASP.FRESF -> `Aircraft.Design.RESERVE_FUEL_MARGIN`

### Test updates
A number of `FRESF` values were listed as small numbers (0.15 to 0.22) and were being converted to `RESERVE_FUEL_ADDITIONAL` in lbm. However, with the updates to `VOLUME_MARGIN`, this caused some BWB missions to stop converging so we modified the source .dat and .csv files for those examples to `FRESF = -.1` which kept the gross mass of the concept the same as the test values.

`reserve_fuel_margin_mass` calculation was updated to not use `Mission.Landing.TOUCHDOWN_MASS` since that is only calculated if Landing=True. This was swapped with the typical way of calculating fuel-burn by taking final mass - initial miss. 

A number of tests had ridiculously small tolerances (1e-9) that were being violated. These were relaxed to 1e-8.

The multi-mission test needed a few more iterations to converge with the change in `Mission.Summary.FUEL_BURNED` calculations so I bumped the iterations from 75 to 150.

### Report updates
The reports for mass summation from pre-mission located in mass_builder.py were changed to display `Mission.Summary.GROSS_MASS` rather than design gross mass. This was important because all the values in that section were mission specific aircraft details.

Fuel reports in `reports.py` were updated to properly include fuel burned in takeoff, which was not correctly being added when `include_takeoff=True`. This was done by referencing `mission:summary:fuel_burned` directly instead of trying to calculate initial mass - final mass again.

### Related Issues

- Resolves #1012

None

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -824,7 +824,9 @@ def add_post_mission_systems(self, verbosity=None):
 
         # Fuel burn in regular phases
         ecomp = om.ExecComp(
-            'fuel_burned = initial_mass - mass_final',  # TODO: Fix to be difference in cumulative fuel burn
+            'fuel_burned = initial_mass - mass_final',
+            # TODO: Fix to include any payloads dropped off during the mission
```
cmbenne3, line 828:
```suggestion
```

Remove TODO -It is already mentioned in issue #1032 so we track it in Git not within the codebase.

aviary/core/aviary_group.py
cmbenne3, line 1506: When do we not have a post mission?

aviary/docs/examples/OAS_subsystem.ipynb
```diff
@@ -105,7 +105,7 @@
     "                num_box_cp=51,\n",
     "            ),\n",
     "            promotes_inputs=[\n",
-    "                ('fuel', av.Mission.Summary.FUEL_MASS),\n",
+    "                ('fuel', av.Aircraft.Fuel.WING_FUEL_CAPACITY),\n",
```
cmbenne3, line 108: Strictly speaking this doesn't do the same thing - it will now calculate the maximum possible mass of the wing, not the mass of the wing when flown on a specific mission (assuming all the fuel for that mission was in the wing). I don't think this matters for the example.

aviary/docs/user_guide/images/Fuel_Breakdown_v2.png
cmbenne3, line 1: Did you want to split out Mission.Summary.Fuel_Burned into constitutent parts?
currently just:

- Takeoff
- Regular Phases

With Taxi out & Taxi In fuel to be modelled/included at a later date.

Do we have a Mission.Summary.RESERVE_FUEL? I guess in theory you could model a different reserve for your design mission than your current off-design mission...? This does seem like an unecessary complication at this stage!

aviary/docs/user_guide/fuel.ipynb
```diff
@@ -0,0 +1,101 @@
...
+    "# Fuel Burn Calculations\n",
+    "\n",
+    "The mass of the fuel consumed throughout the flight, and possibly consumed during the reserve mission as well, is calculated in aviary_group.py.\n",
+    "The fuel mass can be estimated in premission by assuming that the difference between the gross mass and the zero fuel mass is given to fuel.\n",
+    "However, for off-design missions, this estiamte will be inaccurate. For many design missions it can also be inaccurate.\n",
+    "These issues can be compounded by the fact that the aircraft could have some payload it was dropping during the mission. \n",
+    "Payloads currently do not have any place in the mass buildup.\n",
+    "\n",
+    "The most accurate fuel mass buildup is achieved only after flying the primary mission and the reserve missions. \n",
+    "\n",
```
cmbenne3:
```suggestion
```

I don't know what you mean by this? Payloads are included in the mass of the aircraft and appear in the mass buildup in the Aviary Dashboard?
I don't think the comment about buildup is necessary? Surely it's obvious you have to fly the mission before you can calculate how much fuel it required?
We should be consistent about whether we have a fuel buildup or breakdown?

aviary/docs/user_guide/fuel.ipynb
```diff
@@ -0,0 +1,101 @@
...
+    "## Fuel Mass\n",
+    "The Fuel Mass breakdown is as follows:\n",
+    "\n",
+    "- TOTAL_FUEL_MASS (mass)\n",
+    "    - FUEL_BURNED (mass)\n",
+    "    - RESERVE_FUEL (mass)\n",
+    "        - RESERVE_FUEL_BURNED (mass)\n",
+    "        - RESERVE_FUEL_ADDITIONAL (mass)\n",
+    "        - RESERVE_FUEL_MARGIN (%)\n",
+    "\n",
```
cmbenne3, line 27: Is it worth noting that unusable fuel is not counted as 'fuel' in our Aviary definition other than it takes up some of total_capacity. This is made clear by the diagram, but it is maybe unintuitive when we look at a variable called "Total Fuel Mass" in isolation.

aviary/docs/user_guide/fuel.ipynb
```diff
@@ -0,0 +1,101 @@
...
+    "## Maximum Fuel Mass \n",
+    "The total mass fuel that can fit into the vehcile, regardless of gross_mass or fuel volume constraints is as follows:\n",
```
cmbenne3:
```suggestion
    "The total mass fuel that is assumed to fit into the vehicle, regardless of gross_mass constraints is as follows:\n",
```
I think we factor in fuel volume constraints? Unless your point is that all our 'capacity' variables are actually masses not volumes?

aviary/docs/user_guide/fuel.ipynb
```diff
@@ -0,0 +1,101 @@
...
+    "    - UNUSABLE_FUEL_MASS (mass)\n",
+    "\n",
+    "## Constraints\n",
+    "The first constraint is used to ensure the fuel mass does not exceed the total fuel allowed in the fuel tanks.\n",
+    "This is enforced by evaluating {glue:md}`Mission.Constraints.EXCESS_FUEL_CAPACITY`. \n",
+    "If this value is below zero, then there is more fuel than capacity in the tanks. \n",
+    "This constraint can be disabled by setting {glue:md}`Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT` to True.\n",
+    "\n",
+    "A second constraint is used to ensure fuel mass never pushes the aircraft over the gross mass of the vehicle.\n",
+    "This constraint is alaways enforced by {glue:md}`Mission.Constraints.MASS_RESIDUAL`.\n",
```
cmbenne3, line 44: I guess it doesn't matter but I always think of this the other way -> "have I put enough fuel on the aircraft to fly the mission" -> if not then the optimizer will increase gross mass to put more fuel on! (for the design mission at least).

I guess for 'alternate' off design missions then this mass_residual will kick in and prevent you from overloading the aircraft in the way you describe.

aviary/examples/external_subsystems/OAS_mass/OAS_wing_mass_builder.py
```diff
@@ -49,7 +49,7 @@ def build_pre_mission(self, aviary_inputs):
                 num_box_cp=51,
             ),
             promotes_inputs=[
-                ('fuel', av.Mission.Summary.FUEL_MASS),
+                ('fuel', av.Aircraft.Fuel.WING_FUEL_CAPACITY),
```
cmbenne3, line 52: I think this does something subtly different, but as an example that's fine!

aviary/variable_info/variable_meta_data.py
```diff
@@ -1577,16 +1577,13 @@
 )
 
 add_meta_data(
-    Aircraft.Design.RESERVE_FUEL_FRACTION,
+    Aircraft.Design.RESERVE_FUEL_MARGIN,
```
cmbenne3, line 1602: In most places in aviary our variables are described as fractions values between 0-1, this sounds like this would need to be a value between 0-100? My preference would be to go for a fraction between 0-1 as I think this is more consistent with the rest of Aviary?

The input variable is given as a fraction or percentage, but we also call the output value RESEVE_FUEL_MARGIN, which is in units of fuel mass...? Not sure what we should do about that? Perhaps change the name of either the input or the output to make it clear one is a fraction and the other is the computed fuel total.

aviary/variable_info/variable_meta_data.py
```diff
@@ -1577,16 +1577,13 @@
...
     meta_data=_MetaData,
     historical_name={'GASP': None, 'FLOPS': None, 'LEAPS1': None},
     option=True,
     units='unitless',
-    desc='required fuel reserves: given as a proportion of mission fuel. This value must be nonnegative. '
-    'Mission fuel only includes normal phases and excludes reserve phases. '
-    'If it is 0.5, the reserve fuel is half of the mission fuel (one third of the total fuel). Note '
-    'it can be greater than 1. If it is 2, there would be twice as much reserve fuel as mission fuel '
-    '(the total fuel carried would be 1/3 for the mission and 2/3 for the reserve)',
+    desc='required fuel reserves: given as a precentage of mission fuel.'
```
cmbenne3, line 1607:
```suggestion
    desc='required fuel reserves: given as a fraction (0-1) of mission fuel.'
```

aviary/variable_info/variable_meta_data.py
```diff
@@ -3030,6 +3017,16 @@
     default_value=1.0,
 )
 
+add_meta_data(
+    Aircraft.Fuel.VOLUME_MARGIN,
+    meta_data=_MetaData,
+    historical_name={'GASP': 'INGASP.FVOL_MRG', 'FLOPS': None, 'LEAPS1': None},
+    units='unitless',
+    desc='Extra volume required in the wing fuel tank as a persantage of design mission fuel mass.'
```
cmbenne3:
```suggestion
    desc='Extra volume required in the wing fuel tank as a fraction of design mission fuel mass.'
```

aviary/interface/test/test_reserve_support.py
```diff
@@ -66,12 +66,12 @@ def test_gasp_relative_reserve(self):
...
 
-        res_frac = prob.aviary_inputs.get_val(
-            Aircraft.Design.RESERVE_FUEL_FRACTION, units='unitless'
+        reserve_percentage = prob.aviary_inputs.get_val(
+            Aircraft.Design.RESERVE_FUEL_MARGIN, units='unitless'
         )
         td_mass = prob.model.get_val(Mission.Landing.TOUCHDOWN_MASS, units='lbm')
         reserve = prob.model.get_val(Mission.Design.RESERVE_FUEL, units='lbm')
-        assert_near_equal(reserve, res_frac * (140000.0 - td_mass), 1e-3)
+        assert_near_equal(reserve, reserve_percentage / 100 * (140000.0 - td_mass), 1e-3)
```
cmbenne3, line 74: should we have the input variable as a fraction 0-1 or a percentage 0-100?
My preference would be as a fraction as I think it is generally more consistent with the rest of Aviary?

aviary/models/aircraft/blended_wing_body/generic_BWB_GASP.csv
```diff
@@ -31,7 +31,8 @@ aircraft:design:drag_coefficient_increment,0.00025,unitless
 aircraft:design:emergency_equipment_mass,100,lbm
 aircraft:design:max_structural_speed,402.5,mi/h
 aircraft:design:part25_structural_category,3,unitless
-aircraft:design:reserve_fuel_additional,0.2222,lbm
+aircraft:design:reserve_fuel_additional,0,lbm
+aircraft:design:reserve_fuel_margin,10.0,unitless
```
cmbenne3, line 35: Why do you need both volume and reserve margin?

aviary/core/aviary_group.py
ehariton: it looks like that "if post_mission" just checks if the user set it to true or false and depending on that, the reserve components get instantiated in a component with different names.

aviary/models/aircraft/blended_wing_body/generic_BWB_GASP.csv
ehariton: additional is a flat addition of a given number of LBM of fuel.

margin is + mission fuel burn %

aviary/interface/test/test_reserve_support.py
ehariton: it really hurts my brain to think of reserve fuel as a fraction which is why I moved to referring to Margin = percentage instead. Mind you I don't design aircraft....

aviary/variable_info/variable_meta_data.py
ehariton: `Aircraft.Fuel.VOLUME_MARGIN` is being used as a percentage not a fraction. See fuel.py

aviary/variable_info/variable_meta_data.py
ehariton: basically anywhere where the variable name is MARGIN that means percentage. I think we got the MARGIN wording from GASP? That's how they have been using it.

aviary/docs/user_guide/fuel.ipynb
ehariton: This is a good point, I meant deployable payloads. Payloads that do not arrive at the final destination. I'll update the wording.

ehariton: changed this to:



These issues can be compounded by the fact that the aircraft could have some payload it was dropping during the mission. 
Deployable payloads (payloads that are not delivered to the final destination) currently do not have any place in the overall aircraft mass buildup.



And removed comments on most accurate mass buildup.

aviary/docs/user_guide/images/Fuel_Breakdown_v2.png
ehariton: I will be working on a follow-on PR to address this. See #1032

aviary/docs/examples/OAS_subsystem.ipynb
ehariton: FUEL_MASS has been completely removed as a variable so we need some sort of stand-in for the example builder.

aviary/core/aviary_group.py
```diff
@@ -1519,39 +1507,45 @@ def add_fuel_reserve_component(
         else:
             reserve_calc_location = self.model
 
-        RESERVE_FUEL_FRACTION = self.aviary_inputs.get_val(
-            Aircraft.Design.RESERVE_FUEL_FRACTION, units='unitless'
+        RESERVE_FUEL_MARGIN = self.aviary_inputs.get_val(
```
Kenneth-T-Moore: Python PEP 8 reserves capitalized variables for constants, so the local variable RESERVE_FUEL_MARGIN should be lower case.

aviary/docs/user_guide/fuel.ipynb
ehariton: This is part of the reason I put together the graphic on fuel inf Fuel_Breakdown_v2.png because it helps the user understand what each of the elements contains and how it's all summed up.

I added a note in this section that unusable fuel is book kept under `USEFUL_LOAD`.

aviary/subsystems/test/test_premission.py
```diff
@@ -155,9 +155,7 @@ def test_GASP_mass_FLOPS_everything_else(self):
             Aircraft.Fuel.FUEL_SYSTEM_MASS: 1721.08,  # modified from GASP value to account for updated crew mass. GASP value is 1757
             Aircraft.Design.STRUCTURE_MASS: 50931.4,
             Aircraft.Fuselage.MASS: 18833.76678366,  # modified from GASP value to account for updated crew mass. GASP value is 18814
```
Kenneth-T-Moore, line 157: I think I would still test these removed values, but since the variables were removed from the dictionary, you will need to test the new local variable names instead.

aviary/docs/user_guide/fuel.ipynb
ehariton: we don't track fuel volume because fuel capacity for all the tanks is always measured in lbm. I did add a note regarding this is essentially a volume constraint.

aviary/docs/user_guide/fuel.ipynb
ehariton: yeah it really matters in alternate mission where the design gross mass can't change. I can add a note of that.

aviary/variable_info/variable_meta_data.py
ehariton: I'm going to leave this as a precentage

aviary/core/aviary_group.py
ehariton: RESERVE_FUEL_ADDITIONAL and RESERVE_FUEL_MARGIN updated to lowercase in local assignments

aviary/subsystems/mass/flops_based/mass_summation.py
```diff
@@ -430,23 +428,3 @@ def compute(self, inputs, outputs):
...
-
-    def setup_partials(self):
-        self.declare_partials(Mission.Summary.FUEL_MASS, Mission.Design.GROSS_MASS, val=1)
-        self.declare_partials(Mission.Summary.FUEL_MASS, Mission.Summary.ZERO_FUEL_MASS, val=-1)
-
-    def compute(self, inputs, outputs):
-        zero_fuel_mass = inputs[Mission.Summary.ZERO_FUEL_MASS]
-        gross_mass = inputs[Mission.Design.GROSS_MASS]
-
-        outputs[Mission.Summary.FUEL_MASS] = gross_mass - zero_fuel_mass
```
Kenneth-T-Moore, line 452: There is still some value in having the actual fuel in the aircraft as a variable that could be passed into the mission so that we can compute the CG of the aircraft over the mission. I think in that case, we would want:

gross_mass - zero_fuel_mass + unusable_fuel_mass

ehariton: I'll do this as a follow-up PR since this PR is already large / complicated enough.

aviary/core/aviary_group.py
```diff
@@ -1519,48 +1507,54 @@ def add_fuel_reserve_component(
...
             reserve_calc_location = self.model
 
-        RESERVE_FUEL_FRACTION = self.aviary_inputs.get_val(
-            Aircraft.Design.RESERVE_FUEL_FRACTION, units='unitless'
+        reserve_fuel_margin = self.aviary_inputs.get_val(
+            Aircraft.Design.RESERVE_FUEL_MARGIN, units='unitless'
         )
-        if RESERVE_FUEL_FRACTION != 0:
+        if reserve_fuel_margin != 0:
+            # Originally tried to reference Mission.Summary.FUEL_BURNED for fuel burn but in some tests this led to errors
```
jkirk5, line 1514: Which tests, and what error? I'm guessing that happens because 95% of tests don't run a mission, so FUEL_BURNED simply never gets calculated

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_data.py
```diff
@@ -342,7 +342,6 @@
 outputs.set_val(Mission.Summary.USEFUL_LOAD, 5100, 'lbm')
 outputs.set_val(Aircraft.Design.TOTAL_WETTED_AREA, 7956.515738, 'ft**2')
 outputs.set_val(Mission.Summary.ZERO_FUEL_MASS, 103442.0, 'lbm')
-outputs.set_val(Mission.Summary.FUEL_MASS, 26292.0, 'lbm')
```
jkirk5, line 345: Should this be changed to FUEL_BURNED? This prevents us from checking the mission fuel in tests

Kenneth-T-Moore (approved):

aviary/core/aviary_group.py
ehariton: I think this was more of a note to myself that when I tried to use `Mission.Summary.FUEL_BURNED` instead of (initial_mass - final_mass), some tests broke. I'm hoping I can revise how the fuel-burn is calculated in my follow-on PR.

aviary/models/aircraft/advanced_single_aisle/advanced_single_aisle_data.py
ehariton: `Mission.Summary.FUEL_MASS` is no longer calculated by aviary. It's one of those masses that we removed because it was just internal to a GASP calculation. Now that's it's an internal variable we could potentially check it by reaching into that subsystem and examining it's local assignment. But again, this is just meant to size the fuel tanks and fuel systems and never represented the actual fuel going onto the system.

PR #1028, opened 2026-03-17
Remove code for GA and remove BWB if alt_mass

### LINKED ISSUE #439: FLOPS aircraft type for mass estimation is not being properly set
### Description

FLOPS-based mass has equations for 3 types of aircraft - transports, blended-wing-bodies, and general aviation. However, the flag to set this is not defined in the variable hierarchy and therefore isn't set during a normal Aviary run - it always defaults to transport inside the computation components themselves.

Add `Aircraft.Design.AIRCRAFT_TYPE`, with a clear description that this used for FLOPS mass estimation. Equivalent FLOPS variable `IFITE`. FLOPS manual states GA equations are "incomplete", and we need to verify if that is still true.

Add an Enum for aircraft type and use that as accepted values for `AIRCRAFT_TYPE`

### Example

`wing_common.py` has some examples of component options being defined ("`aircraft_type`") but they are never set by the wing mass or mass summation groups. They default to "transport", which is also the only set of equations that have proper testing.

### Aviary Version

0.9.4-dev

_No response_

xjjiang:
### Summary

1. In FLOPS, if alternative mass method is used (`IALTWT = 1`), ignore BWB. 
2. Aviary does not implement general aviation mass equations. Remove relevant code.

### Related Issues

- Resolves #439

None

### New Dependencies

None


aviary/subsystems/mass/flops_based/mass_premission.py
jkirk5, line 1: I think this was actually correct as-is? We can test if this is correct by running a BWB case in FLOPS and toggle the `IALTWT` flag on - if any of the masses changed, then we know that alternate weight equations are still used in the BWB case.

xjjiang: In FLOPS, if `IALTWT` is on, it ignores aircraft type. Are you saying that we don't need to follow the FLOPS logic?

PR #1034, opened 2026-03-20
Issue 1033 - Fix Infinite Payload Range Looping 

### LINKED ISSUE #1033: csv input of settings:payload_range results in infinite looping of payload range code
### Description

In load_inputs when settings:payload_range comes from aviary_inputs, Aviary sets self.generate_payload_range = aviary_inputs.get_val(Settings.PAYLOAD_RANGE).

In run_aviary_problem the final line is to check if generate_payload_range = True then run run_payload_range.

run_payload_range involves loading the inputs from the original problem into a new off design problem and calling run_aviary_problem, which then triggers the above run_payload_range again in an infinite loop.

We need to break this cycle by resetting Settings.Payload_RANGE = False at somepoint before run_aviary_problem is called in the off design mission.

### Example

add settings:payload_range, True to any aircraft .csv file and run through level 1 or level 2 will recreate this issue

### Aviary Version

0.10.0-dev

_No response_

cmbenne3:
### Summary

This PR fixes a bug where setting payload_range in aviary inputs triggered an infinite looping of trying to generate a payload_range diagram.

The quick fix is to reset the aviary inputs value immediately after generate_payload_range has been set, ahead of running the first off design problem. Unfortunately this leaves mismatched aviary inputs.settings.payload_range = False and prob.generate_payload_range = True for the original parent problem. It is unclear how to fix this without introducing another variable, which seems unecessary!

There is also a small change to grab max_iter from the design/sizing problem and parse that to the optimizer for the off_design problems. This solved an issue where off_design problems failed due to hitting the default max iteration limit (50).

### Related Issues

- Resolves #1033 

None

### New Dependencies

None

### DIFF
--- a/aviary/variable_info/variable_meta_data.py
+++ b/aviary/variable_info/variable_meta_data.py
@@ -8199,7 +8199,7 @@ add_meta_data(
     Settings.PROBLEM_TYPE,
     meta_data=_MetaData,
     historical_name={'GASP': None, 'FLOPS': None, 'LEAPS1': None},
-    desc="Select from Aviary's built in problem types: Sizing, Alternate, and Fallout",
+    desc="Select from Aviary's built in problem types: SIZING, ALTERNATE, FALLOUT and MULTI_MISSION",
     option=True,
     types=ProblemType,
     default_value=None,


aviary/core/aviary_problem.py
```diff
@@ -1493,9 +1498,16 @@ def run_off_design_mission(
         if optimizer is None:
             try:
                 optimizer = self.driver.options['optimizer']
+                if optimizer == 'SNOPT':
+                    max_iter = self.driver.opt_settings['Major iterations limit']
+                elif optimizer == 'IPOPT':
+                    max_iter = self.driver.opt_settings['max_iter']
+                elif optimizer == 'SLSQP':
+                    max_iter = self.driver.opt_settings['maxiter']
             except KeyError:
```
Kenneth-T-Moore, line 1489: This keyerror is meant to catch options['optimizer'], but with the addition of the lines above to the block, it will also be triggered on opt_settings['max_iter']. The opt_settings is meant as a way to pass in optimizer-specific options to an optimizer, and it stars as an nearly empty dict with a couple of basic IFILE type keys. This means that the dict doesn't contain "max_iter", etc, until Aviary adds them in `add_driver`.

So, the corner case here is if the user adds a driver directly instead of through add_driver. To get around this in the best way, I think you should move the whole max_iter block into another try block, and just pass (i.e., keep current default max-Iter on the optimizer) on KeyError. Otherwise, optimizer will be None.

aviary/core/aviary_problem.py
```diff
@@ -454,6 +454,8 @@ def add_driver(self, optimizer=None, use_coloring=None, max_iter=50, verbosity=N
             optimizer = 'IPOPT'
         if use_coloring is None:
             use_coloring = True
+        if max_iter is None:
```
Kenneth-T-Moore, line 448: I don't think we actually have to move the default into the code like this.

aviary/core/aviary_problem.py
```diff
@@ -1493,9 +1498,16 @@ def run_off_design_mission(
...
+                    max_iter = self.driver.opt_settings['Major iterations limit']
+                elif optimizer == 'IPOPT':
+                    max_iter = self.driver.opt_settings['max_iter']
+                elif optimizer == 'SLSQP':
+                    max_iter = self.driver.opt_settings['maxiter']
             except KeyError:
                 optimizer = None
-        off_design_prob.add_driver(optimizer, verbosity=verbosity)
+                max_iter = None
+        off_design_prob.add_driver(optimizer=optimizer, max_iter=max_iter, verbosity=verbosity)
```
Kenneth-T-Moore, line 1502: Another option to consider is to allow the user to specify max-iter when they call run_offdesign_mission.

aviary/core/aviary_problem.py
```diff
@@ -1261,6 +1263,9 @@ def run_aviary_problem(
                 self.model.list_outputs(out_stream=outfile)
 
         if self.generate_payload_range:
+            self.aviary_inputs.set_val(Settings.PAYLOAD_RANGE, False)
+            # aviary inputs unforunately altered for the problem that triggered generation of payload range
+            # self.generate_payload_range remains True for the parent problem.
             self.run_payload_range()
```
Kenneth-T-Moore, line 1257: Maybe save and reset the original value for the PAYLOAD_RANGE switch after a successful run.

aviary/validation_cases/benchmark_tests/test_bench_off_design.py
```diff
@@ -507,7 +507,10 @@ def test_payload_range(self):
             'min',
         )
         prob.load_inputs('models/aircraft/test_aircraft/aircraft_for_bench_FwFm.csv', phase_info)
-        # prob.aviary_inputs.set_val(Aircraft.Fuel.IGNORE_FUEL_CAPACITY_CONSTRAINT, True)
+        prob.aviary_inputs.set_val(Settings.PAYLOAD_RANGE, True)
```
Kenneth-T-Moore: This is pretty hacky, and it gives more weight to the idea of changing the API a bit, replacing this step with 

  aviary_inputs = av.load_aircraft('models/aircraft/test_aircraft/aircraft_for_bench_FwFm.csv')
  aviary_inputs.set_val(whatever we want)
  prob.load_inputs(aviary_inputs, phase_info)

aviary/core/aviary_problem.py
cmbenne3: The idea was to match the structure of how the other arguments (optimizer, use_coloring and verbosity) were handled (and allow my code in run_off_design_mission() to work if the user has not specified any driver details for the sizing mission, so 'none' arguments would be parsed onto the add_driver() function call).

aviary/core/aviary_problem.py
cmbenne3: good idea - added another set_val to correct it!

aviary/core/aviary_problem.py
cmbenne3: I want to be able to run payload range from level 1 interface, and I think it's helpful if the user can change the iterations just for the design problem and have it flow through into the off design. To cope with this and your edge case I think the extra try: except: block is required anyway?

aviary/core/aviary_problem.py
cmbenne3: Added an additional try: except: block to handle the max_iter separately.

aviary/validation_cases/benchmark_tests/test_bench_off_design.py
cmbenne3: Agree this was very hacky! I have now updated this using create_vehicle as load_aircraft doesn't exist.

aviary/core/aviary_problem.py
```diff
@@ -1495,7 +1501,18 @@ def run_off_design_mission(
                 optimizer = self.driver.options['optimizer']
             except KeyError:
                 optimizer = None
-        off_design_prob.add_driver(optimizer, verbosity=verbosity)
+            try:
```
Kenneth-T-Moore: I think this try block needs to be dedented 1 level so that it triggers if optimizer is set in the args.

aviary/core/aviary_problem.py
cmbenne3: You're correct that we don't actually have to move the default into the code - we decided we would still do it to match the existing code structure with the other arguments optimizer and use_coloring

aviary/core/aviary_problem.py
cmbenne3: done

jkirk5:
> We could do this - is there ever a time when it makes sense to run a payload_range diagram from a non-design mission?

I can't think of any, and the payload-range method is always available to directly call if you did for some reason.

Issue #1049, opened 2026-03-26
Update Calculation and name for Mission.USEFUL_LOAD


cmbenne3:
### Desired capability or behavior.

Mission.USEFUL_LOAD variable is misnamed.
If we want to calculate and display USEFUL_LOAD for the user it should be a property of the sized aircraft and calculated as  Aircraft.Design.USEFUL_LOAD = Aircraft.Design.GROSS_MASS - Aircraft.Design.EMPTY_MASS
This quantity seems most useful for aircraft manufacturers trying to sell their aircraft to customers!

We currently have Mission.USEFUL_LOAD summing an intermediate bunch of weights equal to operating_mass - empty mass.
This group of weights may not need to be summed or in the hierarchy, but for consistency with rest of components could be named Mission.OPERATING_LOAD (and Mission.Design.OPERATING_LOAD if required to be saved for the sizing mission).

The current list of things that go into this quantity are:
+ cabin_crew_mass
+ flight_crew_mass
+ unusable_fuel_mass
+ oil_mass
+ pass_service_mass
+ cargo_container_mass

It's basically 'stuff you need to add to the dry empty weight of the aircraft after it leaves the factory to enable operation of the aircraft on the planned mission'.

Issue #1048, opened 2026-03-26
Investiate per mission vs per aircraft variables


cmbenne3:
### Desired capability or behavior.

Aviary has inconsistently translated the FLOPS $MISSIN variables as follows:
FACT -> Mission.FUEL_FLOW_SCALER (this is modifiable separately for each mission a sized aircraft flies)
FCDO - > Aircraft.Design.ZERO_LIFT_DRAG_COEFF_FACTOR
FCDI -> Aircraft.Design.LIFT_DEPENDENT_DRAG_COEFF_FACTOR
FCDSUB -> Aircraft.Design.SUBSONIC_DRAG_COEFF_FACTOR
FCDSUP -> Aircraft.Design.SUPERSONIC_DRAG_COEFF_FACTOR

In FLOPS all of these values are variable per mission.
In Aviary the fuel flow is, but the aerodynamic variables are not.

We should be consistent:

- If we like the ability to change all of these per mission then we should move the location of the Aero parameters within the data hierarchy under Mission.* from Aircraft.Design.*.
- If we prefer that these variables 'define' and aircraft then Mission.FUEL_FLOW_SCALAR should be removed since that functionality is available on a per-engine basis by setting: Aircraft.Engine.SUBSONIC_FUEL_FLOW_SCALER and Aircraft.Engine.SUPERSONIC_FUEL_FLOW_SCALER.

This Issue should include a full search of the repository for any other variables that need changing in a similar way.
This Issue also requires investigation of how any of these variables may impact GASP based analysis.

Issue #1046, opened 2026-03-25
Rename Mission.Constraints.MAX_MACH to Aircraft.design.MAX_MACH


cmbenne3:
### Desired capability or behavior.

This variable is actually used in aviary as an aircraft design variable and so should be renamed to Aircraft.Design.MAX_MACH.

<img width="617" height="432" alt="Image" src="https://github.com/user-attachments/assets/b8e1bdc0-f4e1-4f41-9916-3e83bd245398" />

xjjiang:
> Description should be updated to reflect what it is used to actually calculate (which subsystems use it, if it is used to compute only one or two specific variables, list them)

It is used in several FLOPS based mass subsystems: air-conditioning mass, fuel system mass, hydraulics mass, instrument mass, passenger service mass, starter mass, surface control mass. It is also used to compute lift coefficient in FLOPS based aerodynamics subsystem.

Issue #1045, opened 2026-03-25
Overhaul Aviary Time calculations


cmbenne3:
### Desired capability or behavior.

Aviary should calculate some mission time variables that correspond to useful durations within the mission.

Currently there is a variable Mission.Final_Time which is the duration of the dymos modelled regular phases within the mission. Depending on whether detailed takeoff and landing are used, or 2dof equations of motion are used then this might or might not include takeoff and landing.

Suggested change will ensure Aviary calculates the following:
Mission.BLOCK_TIME = Mission.Taxi.OUT_TIME + Mission.Takeoff.TIME + Mission.FLIGHT_TIME + Mission.APPROACH_TIME + Mission.Landing.TIME + Mission.Taxi.IN_TIME

Not all of these are modeled at the moment.

Logic will need to be added to sum the correct times with/without taxi, takeoff, approach and landing, depending what phases are modelled in dymos, and whether simple or detailed takeoff/landing are modeled.

Aviary should also calculate the Mission.RESERVE_TIME (which should be the duration of the reserve phases + holds etc)

PR #1050, opened 2026-03-27
Rename Mission.Design.XXX and Mission.Summary.XXX variables for clarity

### LINKED ISSUE #238: Consolidate the idea of `'target_range'` in `phase_info` and Mission.Design.RANGE

Currently, if you set the `'target_range'` in `post_mission` within your `phase_info` object then a constraint will be added to make the mission distance match that value. This is easily confusing with `Mission.Design.RANGE` also existing. Users might set one or the other thinking it would impact the mission they're evaluating. We should consolidate or clarify this in some way to prevent confusion.

Another consideration is that we're trying to keep the aircraft definition and mission definition as two separate files. If the user sets `Mission.Design.RANGE` in their .csv, should that directly affect the performed mission length in Aviary? This might need to wait for the notion of off-design to come into Aviary; it's definitely affected.

Please chat with me about more details on this issue before tackling it.

_No response_

### Associated Bug Report

_No response_

### LINKED ISSUE #772: Renaming of design vs summary mission variables

~~The off design mission analysis currently instantiates a whole new aviary problem object and re-runs the entirety of the pre-mission block. This is not really necessary (particularly for subsystem masses) which are not expected to change from the initial design mission. Since all these parameters are recalculated the aircraft is basically being re-designed (although if everything is correct it re-designs to be identical to the sizing aircraft). We might want to refactor the code to avoid these extra calculations.
We should also consider the multi-mission architecture and look for ways to align off design with it since there is some similarity in output.~~ (edited by @jkirk5 to split this off into its own issue)

Regardless of the above we currently have 'design' and 'summary' mission variables, which are not always consistently named and perhaps we should change 'summary' to 'as-flown' or 'mission1,2,3 etc' (open to other suggestions) to make it much clearer to the user what the summary variables are actually doing.
Not all variables have a 'summary' version (e.g. operating weight, zero fuel weight). Some variables are named differently between design and summary (e.g. mission.summary.TOTAL_FUEL_MASS vs mission.design.fuel mass vs mission.design.fuel_mass_required which I think are actually all identical for a sizing mission...?). We should standardize all of these so that a user can easily use the same code to interrogate both a sizing and off design problem object.

_No response_

### Associated Bug Report

_No response_

### LINKED ISSUE #991: Variable name change request: `Aircraft.Design.TOUCHDOWN_MASS`

Change to `Aircraft.Design.LANDING_MASS`? It is even described as that in the variable description, and is inherited from FLOPS' `WLDG`, or landing weight. I don't believe touchdown mass is a more common verbiage (personally I've never used it over landing weight).

_No response_

### Associated Bug Report

_No response_

cmbenne3:
### Summary

- Changes the names of Mission.Design.XXX to Aircraft.Design.XXX, and Mission.Summary.XXX to Mission.XXX.
- Removes some unnecessary variables.
- Removes the duplicate variable Mission.Landing.TOUCHDOWN_MASS and replaces references to it with Mission.FINAL_MASS

Mission.Design.XXX
|class Mission.Design:| New Variable Name | Notes |
|---------|---------|---------|
|Mission.Design.CRUISE_ALTITUDE = 'mission:design:cruise_altitude'| Aircraft.Design.CRUISE_ALTITUDE = 'aircraft:design:cruise_altitude' | This variable is only used to calculate max mach for GASP-based Mass in design_load.py It has hardcoded atmospheric constants and should be updated to use Aviary's atmospheric subsystem Issue #1039 .
|Mission.Design.CRUISE_RANGE = 'mission:design:cruise_range'| N/A | Removed as not used anywhere in codebase|
|Mission.Design.GROSS_MASS = 'mission:design:gross_mass'| Aircraft.Design.GROSS_MASS = 'aircraft:design:gross_mass' | We update this to Aircraft since it makes no sense to track the design mission gross mass under mission.|
|Mission.Design.LIFT_COEFFICIENT = 'mission:design:lift_coefficient'| Aircraft.Design.LIFT_COEFFICIENT = 'aircraft:design:lift_coefficient' | This is not actively set by any of our models, but can be used to overwrite the value clculated by EDET if desired |
|Mission.Design.LIFT_COEFFICIENT_MAX_FLAPS_UP = 'mission:design:lift_coefficient_max_flaps_up'| Aircraft.Design.LIFT_COEFFICIENT_MAX_FLAPS_UP = 'aircraft:design:lift_coefficient_max_flaps_up' | Currently only specified in GwFm... but seems to be an input to gasp aero...? |
|Mission.Design.MACH = 'mission:design:mach'| Aircraft.Design.MACH = 'aircraft:design:mach' | This is an aircraft Design Variable, mission mach can be updated in phase info|
|Mission.Design.RANGE = 'mission:design:range'| Aircraft.Design.RANGE = 'aircraft.design.range' | This is really an aircraft design variable used to size flops subsystems. It is used as an initial guess for target range when not set in phase info |
|RATE_OF_CLIMB_AT_TOP_OF_CLIMB = 'mission:design:rate_of_climb_at_top_of_climb' | N/A | Variable removed as not used anywhere in Aviary. We should add an example or document how a user would apply a minimum top of climb climb rate constraint as this is a common constraint to have in aircraft design. Issue #1037 tracks this |
|Mission.Design.RESERVE_FUEL = 'mission:design:reserve_fuel'| Mission.TOTAL_RESERVE_FUEL = 'mission:total_reserve_fuel' | This is moved under Mission directly as it is not a design variable. As the sum of Mission.RESERVE_FUEL + Mission.RESERVE_FUEL_ADDITIONAL + Mission.RESERVE_FUEL_MARGIN it must be allowed to vary depending on the reserve mission being modelled. Fuel additional and fuel margin are also renamed to be set at the Mission.* level |
|Mission.Design.THRUST_TAKEOFF_PER_ENG = 'mission:design:thrust_takeoff_per_eng'| Aircraft.Design.THRUST_TAKEOFF_PER_ENG = 'aircraft:design:thrust_takeoff_per_eng' | This is only used for height energy simple takeoff. By default it should probably be calculated by the propulsion subsystem (e.g. take SLS thrust condition from engine deck), and only overridden intentionally by experienced users. Issue #1038 Not moved to Mission.Takeoff. group so as not to confuse with detailed takeoff parameters. This parameter is also independent of the mission to be flown |

Mission.Summary.XXX
| class Summary: | New Variable Name | Notes |
|---------|---------|---------|
|Mission.Summary.CRUISE_MACH = 'mission:summary:cruise_mach' | Aircraft.Design.CRUISE_MACH = 'aircraft:design:cruise_mach | This variable is actually only used to calculate the landing_to_takeoff_mass_ratio, as a result moved to aircraft.design as it is not mission specific. The actual mach in cruise is specified in phase info dictionary, and might be optimized so having a single value under mission is confusing. Note Aircraft.Design.CRUISE_MACH is distinct from Aircraft.Design.MACH which is an aerodynamic design override or computed internally by EDET. There is also Mission.Constraints.MAX_MACH which is used in Aviary premission for sizing systems (equivalent to FLOPS VMMO). This is incredibly misleading and changing it is tracked in issue #1046 | 
|Mission.Summary.CRUISE_MASS_FINAL = 'mission:summary:cruise_mass_final'| N/A | This is removed as it is only used to set initial guesses for two_dof_problem_configurator. If the user wants this functionality they should set it within the .csv file using 'initialization_guesses:cruise_mass_final' or set ['cruise'][initial_guesses]['mass'] in phase info?
|Mission.Summary.FINAL_MASS = 'mission:summary:final_mass'| Mission.FINAL_MASS = 'mission:final_mass' | This variable is the mass of the aircraft at the end of the final regular phase 'traj.{final_phase}.states:mass' where final_phase = self.regular_phases[-1] It is not really used anywhere in Aviary, but is potentially useful to the user. It is clearly documented in variable MetaData. |  
|Mission.Summary.FINAL_TIME = 'mission:summary:final_time'| Mission.FINAL_TIME = 'mission:final_time' |  This variable is the duration of the dymos phases of the mission. We should instead calculate more useful time variables such as bock_time, and flight_time. These will depend on what phases are modelled and included. Issue #1045 tracks this. |
|Mission.Summary.FUEL_BURNED = 'mission:summary:fuel_burned'| Mission.FUEL | Summary is unnecessary |
|Mission.Summary.FUEL_FLOW_SCALER = 'mission:summary:fuel_flow_scaler'| Mission.FUEL_FLOW_SCALER = 'mission.fuel_flow_scaler | This PR just renames this variable, but Aviary is inconsistent with translation of FLOPS MISSIN variables since FCDO, FCDI, FCDSUB and FCDSUP are technically all mission variables in FLOPS, but are Aircraft.Design variables in Aviary. If we like this change then this variable could just be removed completely since the Aviary equivalents of Aircraft.Engine.SUBSONIC_FUEL_FLOW_SCALER and Aircraft.Engine.SUPERSONIC_FUEL_FLOW_SCALER can be used to replace the per-mission additional scaler here. Issue #1048 tracks this. |
|Mission.Summary.GROSS_MASS = 'mission:summary:gross_mass'| Mission.GROSS_MASS = 'mission:gross_mass' | Summary is unnecessary. |
|Mission.Summary.OPERATING_MASS = 'mission:summary:operating_mass'| Mission.OPERATING_MASS = 'mission:operating_mass' | Summary is unnecessary. |
|Mission.Summary.RANGE = 'mission:summary:range' | Mission.RANGE = 'mission:range' | Summary is unnecessary. |
|Mission.Summary.RESERVE_FUEL_BURNED = 'mission:summary:reserve_fuel_burned'| Mission.RESERVE_FUEL = 'mission:reserve_fuel' | Summary is unnecessary. |
|Mission.Summary.TOTAL_FUEL_MASS = 'mission:summary:total_fuel_mass'| Mission.TOTAL_FUEL = 'mission:total_fuel_mass' | Summary is unnecessary. |
|Mission.Summary.USEFUL_LOAD = 'mission:summary:useful_load'| Mission.USEFUL_LOAD = 'mission:useful_load' | This PR only modifies the name of this variable. This variable is poorly named. In aviation the useful load is Gross mass - Empty mass. Issue #1049 tracks changing this. |
|Mission.Summary.ZERO_FUEL_MASS = 'mission:summary:zero_fuel_mass'| Mission.ZERO_FUEL_MASS = 'mission:zero_fuel_mass' | Summary is unnecessary |

### Related Issues

- Resolves #772 
- Resolves #238 
- Resolves #991

None

### New Dependencies

None

### DIFF
--- a/aviary/validation_cases/benchmark_tests/test_bwb_GwGm.py
+++ b/aviary/validation_cases/benchmark_tests/test_bwb_GwGm.py
@@ -8,7 +8,7 @@ from openmdao.utils.testing_utils import require_pyoptsparse, use_tempdirs
 from aviary.models.aircraft.blended_wing_body.generic_BWB_2dof_phase_info import phase_info
 
 from aviary.interface.run_aviary import run_aviary
-from aviary.variable_info.variables import Mission
+from aviary.variable_info.variables import Aircraft, Mission
 
 
 @use_tempdirs


aviary/mission/solved_two_dof_problem_configurator.py
```diff
@@ -28,9 +28,18 @@ def initial_guesses(self, aviary_group):
         """
         # This doesn't really have much value, but is needed for initializing
         # an objective-related component that still lives in level 2.
-        aviary_group.target_range = aviary_group.aviary_inputs.get_val(
-            Mission.Design.RANGE, units='NM'
-        )
+        # aviary_group.target_range = aviary_group.aviary_inputs.get_val(
```
Kenneth-T-Moore: Can remove the commented lines.

aviary/mission/two_dof_problem_configurator.py
```diff
@@ -545,12 +538,6 @@ def link_phases(self, aviary_group, phases, connect_directly=True):
         aviary_group.connect('traj.ascent.timeseries.time', 'h_fit.time_cp')
         aviary_group.connect('traj.ascent.timeseries.altitude', 'h_fit.h_cp')
 
-        aviary_group.connect(
```
Kenneth-T-Moore, line 554: Removing this connection breaks mass continuity between the end of the last flight phase, and the start of the landing. I think you want to connect it to `Mission.FINAL_MASS` instead, based on the change in line 643.

aviary/models/aircraft/large_single_aisle_1/large_single_aisle_1_GASP.csv
```diff
@@ -137,19 +137,18 @@ aircraft:wing:taper_ratio,0.33,unitless
 aircraft:wing:thickness_to_chord_root,0.15,unitless
 aircraft:wing:thickness_to_chord_tip,0.12,unitless
 aircraft:wing:zero_lift_angle,-1.2,deg
-mission:design:cruise_altitude,37500,ft
-mission:design:gross_mass,175400,lbm
-mission:design:mach,0.8,unitless
-mission:design:range,3675,NM
-mission:design:rate_of_climb_at_top_of_climb,300,ft/min
```
Kenneth-T-Moore, line 144: Yeah, we specify this in the phase_info now with `required_available_climb_rate` (at least in energy method).

aviary/mission/two_dof_problem_configurator.py
cmbenne3: What about this in aviary_group: does it not do the same thing?

Image

Image

I was confused as to why we have extraction of mass from the final regular phase in multiple places in the code?

aviary/models/aircraft/large_single_aisle_1/large_single_aisle_1_GASP.csv
cmbenne3: Does this mean that issue #1037 is not really necessary since we can do that by setting a phase_info parameter?

aviary/mission/two_dof_problem_configurator.py
Kenneth-T-Moore: Oh, I see what is happening. There is already an output that produces FINAL_MASS, so when Touchdown mass goes away, the landing component just hooks up there.

aviary/models/aircraft/large_single_aisle_1/large_single_aisle_1_GASP.csv
Kenneth-T-Moore: Well, we may still need the capability in 2dof, so the issue could still serve that purpose.

Kenneth-T-Moore (approved):

ehariton (changes requested):
Lets update the following variables for further clarity:
Mission.Summary.TOTAL_FUEL_BURNED -> Mission.TOTAL_FUEL
Mission.Summary.FUEL_BURNED -> Mission.FUEL
Mission.Design.RESERVE_FUEL -> Mission.TOTAL_RESERVE_FUEL
Mission.Summary.RESERVE_FUEL_BURNED -> Mission.RESERVE_FUEL
Aircraft.Design.RESERVE_FUEL_ADDITIONAL -> Mission.RESERVE_FUEL_ADDITIONAL
Aircraft.Design.RESERVE_FUEL_MARGIN -> Mission.RESERVE_FUEL_MARGIN

This reflects the fact that everything about fuel is mission dependent. The only real thing the design missions effect is fuel-capacity.

aviary/examples/multi_mission/run_multimission_example.py
ehariton, line 1: I don't think these files in aviary/examples should have been moved?

aviary/mission/solved_two_dof_problem_configurator.py
```diff
@@ -28,9 +28,15 @@ def initial_guesses(self, aviary_group):
         """
         # This doesn't really have much value, but is needed for initializing
         # an objective-related component that still lives in level 2.
-        aviary_group.target_range = aviary_group.aviary_inputs.get_val(
-            Mission.Design.RANGE, units='NM'
-        )
+
```
ehariton, line 31: I hope you told users that if they specify both of these, (one in mission info, one in the .csv) only one will be used.

aviary/subsystems/aerodynamics/flops_based/test/test_tabular_aero_group.py
```diff
@@ -672,7 +672,7 @@ def setup(self):
             'pre_mission',
             CorePreMission(aviary_options=aviary_options, subsystems=default_premission_subsystems),
             promotes_inputs=['aircraft:*'],
-            promotes_outputs=['aircraft:*', 'mission:*'],
```
ehariton, line 675: Mission is no longer needed to be promoted in here because all input vars have been upd to Aircraft*

aviary/docs/examples_unreviewed/multi_mission.ipynb
```diff
@@ -473,11 +473,11 @@
...
     "mission2.aircraft:crew_and_payload:passenger_service_mass (lbm), [2524.47559296]\n",
     " \n",
-    "mission1.mission:summary:gross_mass (lbm), [157440.88610181]\n",
-    "mission2.mission:summary:gross_mass (lbm), [120056.71107322]\n",
+    "mission1.mission:gross_mass (lbm), [157440.88610181]\n",
+    "mission2.mission:gross_mass (lbm), [120056.71107322]\n",
     " \n",
-    "mission1.mission:summary:fuel_burned (lbm), [27050.67106054]\n",
-    "mission2.mission:summary:fuel_burned (lbm), [22916.49603196]\n",
+    "mission1.mission:fuel_burned (lbm), [27050.67106054]\n",
```
ehariton: I think this should be "mission1.mission:fuel"
and below should be "mission2.mission:fuel"

cmbenne3: good catch fixed now - at somepoint maybe we should glue these variable prints so that the docs fail if the variable names change.

PR #1054, opened 2026-03-28
Fix np.where on scalar boolean in EngineMass

bdphilli:
np.where(scale_mass) fails when scale_mass is a 0-d array or scalar boolean (NumPy deprecation). Wrap in np.atleast_1d() in both compute() and compute_partials() so indexing works regardless of input dimensionality.


ehariton (approved):

jkirk5 (changes requested):
I'd like to test an example case that replicates the issue and try and debug from the source (preprocessor failure?) first

PR #1055, opened 2026-03-28
Set aircraft-derived upper bounds on mass and distance states

### LINKED ISSUE #607: Current mass scaling in flight_phase_builder is not ideal.
### Description

Currently, the mass state is added in flight_phase_builder.py with a ref of 1e4 and a defect _ref of 1e6. 

1. I think these are actually switched from what they should be. The 1e6 on the defect ref means the optimizer doesn't see the mass defects as important as other defects. We have had some success in Jes's detailed take-off case with using a smaller value (1e3), so that when the optimizer couldn't satisfy all the constraints, it didn't leave you with large mass discontinuities.
2. These values are hardcoded, and it would be beneficial for them to be added to the phase_info parameters so that a user can control them directly.

### Example

Contact Ken and Jes for specific cases, though item 2 above can be done without one.

### Aviary Version

0.9.4-dev

_No response_

bdphilli:
When mass_bounds or distance_bounds have None as the upper bound (the default), Dymos sets it to ~1e21. With mass_ref=1e4 and distance_ref=1e6, this creates scaled upper bounds of ~1e17 and ~1e15 respectively, while other design variables have scaled ranges of O(1).

In build_phase(), replace None upper bounds with values derived from aviary_inputs: 2x GROSS_MASS for mass, 2.5x DESIGN_RANGE for distance. This reduces scaled upper bounds to O(10), matching the scale of other design variables.

User-specified bounds are preserved — the override only applies when the upper bound is None.

Resolves #607


ehariton:
@robfalck , I didn't realize dymos set any constraints if the user specifies there are no bounds to a state. Do you think we'd get easier convergence if we set the bound guesses as Ben suggests? If so, maybe we should we reevaluate all instances where we set bounds to None?

robfalck:
The way some optmizers work is that having a bound value beyond +/- 1E21 or 1E30 (depends on the optimizer), they treat it as "unbounded" and handle it a bit differently internally. I think the default behavior in OpenMDAO should be that bounds of None.

One of the real issues with this is that the scaling forces the upper and lower bounds to cross the "treat this as unbounded" threshold.  This is fixed in the autoscaler pull request on OpenMDAO.

ehariton:
@bdphilli, @robfalck said he wants to take another look at this.

jkirk5:
@robfalck are the None upper bounds still an issue that requires manual fixes like in this PR?

robfalck:
For the time being, we had some outside contributions come in that conflicted at the time. Let me see if I can get it PR'd today.

PR #1073, opened 2026-04-02
Engine scaling

### LINKED ISSUE #718: Unintuitive behavior of Aircraft.Engine.SCALE_PERFORMANCE

The flag Aircraft.Engine.SCALE_PERFORMANCE is intended to be a switch for users to prevent their engine from being scaled. However its behavior is very unintuitive. By bypassing scaling, it essentially mimics Aircraft.Engine.SCALE_FACTOR being 1, without actually changing it (or SLS thrust). This means that those variables are ignored for performance calcs, but not other calculations like mass calculations. It also means the optimizer can still try and adjust this variable, but it will have no effect.

To fix this, the flag's behavior should be changed so that it instead changes whether SCALE_FACTOR is added as a design variable. The name of the variable may need to change to better reflect this. Intended behavior of scaling parameters is this:
User sets SCALE_FACTOR or SCALED_SLS_THRUST - this is either the fixed state of the engine if not letting Aviary size/optimize the engine, or the initial condition if optimizing.
Engine mass calculations now always reflect SCALE_FACTOR and updated thrust whether or not engine is being optimized (avoid things like overwriting SCALE_FACTOR to 1 if SCALE_PERFORMANCE is False)

_No response_

### Associated Bug Report

_No response_

jkirk5:
### Summary

Removes unnecessary variable Aircraft.Engine.SCALE_PERFORMANCE

### Related Issues

- Resolves #718

None

### New Dependencies

None


Kenneth-T-Moore (approved):

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -267,49 +260,17 @@ def _preprocess_inputs(self):
...
+            # extrapolation
             if idle_min > idle_max:
-                if self.get_val(Settings.VERBOSITY).value >= 1:
+                if self.get_val(Settings.VERBOSITY).value >= Verbosity.BRIEF:
                     warnings.warn(
-                        f'EngineDeck <{self.name}>: Minimum flight idle fraction '
-                        'exceeds maximum flight idle fraction. Values for min and max '
-                        'fraction will be flipped.'
+                        f'EngineDeck <{self.name}>: Minimum flight idle fraction exceeds maximum '
+                        'flight idle fraction. Values for min and max fraction will be flipped.'
```
cmbenne3: Personally I would be in favor of removing this 'helpfulness' and throwing an error if a user inputs min fraction > max fraction rather than guessing these are the correct values entered for the wrong variables.
If a user entered a fuselage width greater than its length would we throw a warning and try and flip them? What about for Min and Max Mach numbers for a cruise segment, I'm sure OpenMDAO / Dymos will error if Min > Max in that case rather than assuming we can guess what the user meant?

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1364,44 +1320,34 @@ def _set_reference_thrust(self):
...
+                    scaled_thrust,
+                    ref_thrust * scale_factor,
+                    abs_tol=1e-1,
+                    rel_tol=1e-4,
+                ):
+                    # user wants scaling but provided conflicting inputs, cannot be resolved
+                    raise AttributeError(
+                        f'EngineDeck <{self.name}>: Conflicting values provided for '
+                        'aircraft:engine:scale_factor and aircraft:engine:scaled_sls_thrust when '
+                        'compared against aircraft:engine:reference_sls_thrust'
```
cmbenne3:
```suggestion
                        f'EngineDeck <{self.name}>: Conflicting values provided for '
                        'aircraft:engine:scale_factor {scale_factor} and aircraft:engine:scaled_sls_thrust {scaled_thrust} when '
                        'compared against aircraft:engine:reference_sls_thrust {ref_thrust}'
```

Perhaps we can add in the values that break the check here so the user can easily find and change the discrepancy.

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1364,44 +1320,34 @@ def _set_reference_thrust(self):
...
-                    self.set_val(Aircraft.Engine.SCALE_FACTOR, 1.0)
+                    target_thrust = ref_thrust * scale_factor
+                    if self.get_val(Settings.VERBOSITY) >= Verbosity.VERBOSE:
+                        warnings.warn(
+                            f'EngineDeck <{self.name}>: aircraft:engine:scaled_sls_thrust and '
+                            'product of aircraft:engine_scale_factor and '
+                            'aircraft:engine:reference_sls_thrust are not an exact match but within '
+                            'tolerance. Setting scaled thrust target to calculated value of '
+                            f'{target_thrust} lbf.'
+                        )
```
cmbenne3, line 1355: I think we should simplify this and go with the following logic:
If both SCALED_SLS_THRUST and SCALE_FACTOR are provided warn the user, ignore one and use the other (maybe sscale_factor should take precedence)?
This eliminates needing to do a tolerance check, and Aviary's logic is clear: it uses scale_factor to scale the engine. If scale factor is not provided it will be calculate based on SCALED_SLS_THRUST and reference thrust.

I think this warning as written will always trigger if thrust_provided = True, even if the the scaling is an exact match? If we want this warning then we should add it under an additional 'if' checking for within tolerance but not exact? Personally I think this is unecessary - if the provided scaling matches to within 1e-4 that's 0.01% and more than accurate enough to be ignored.

aviary/subsystems/propulsion/engine_deck.py
```diff
@@ -1364,44 +1320,34 @@ def _set_reference_thrust(self):
             # both scale factor and target thrust provided:
```
cmbenne3, line 1317: As described below I think we should simplify this block of code for if we have both.
I would throw a warning stating both have been provided and so scaled_sls_thrust will be ignored.
Then I would move to the 'scale_factor has been provided, no scaled_thrust provided' logic block so Aviary is consistent and remove the complexity of the tolerance and warnings etc.

cmbenne3 (changes requested):

aviary/subsystems/propulsion/engine_deck.py
jkirk5: You shouldn't ignore one because SCALED_SLS_THRUST is later used for calculations, so if it doesn't match scale factor it gets overwritten to a "incorrect" value. It's either check for a match or always error when both are provided.

jkirk5: I made it a DEBUG message for now

aviary/subsystems/propulsion/engine_deck.py
jkirk5: As mentioned before, I am not sure if it is safe to ignore SLS thrust value provided. I think I will put a deprecation warning here that in the future we will not allow specifying both, with the plan to only accept scale factor in the future.

cmbenne3 (approved):

Issue #363, opened 2024-06-27
Unify legacy-code based constants


jkirk5:
### Desired capability or behavior.

Aviary should have a single set of constants to use. These constants should be set based on a logical (and documented) source, such as ANSI standard constants or limited to a specific number of significant figures.

jkirk5:
The constants file currently has takeoff/landing coefficients of friction. This is the wrong place for those variables - they are not universal physical constants. A user should be able to change them, so they should be variables in the hierarchy with default values that match what is currently in constants.

xjjiang:
Should we hard code `6.687 lbm/galUS` as aviation fuel density? This is GASP default. On FLOPS side, the default is `6.7`.

xjjiang:
> The constants file currently has takeoff/landing coefficients of friction. This is the wrong place for those variables - they are not universal physical constants. A user should be able to change them, so they should be variables in the hierarchy with default values that match what is currently in constants.

This is done in PR #1170.

jkirk5:
> Should we hard code `6.687 lbm/galUS` as aviation fuel density? This is GASP default. On FLOPS side, the default is `6.7`.

Good point! That should also be a constant. I am planning an update to the way Aviary handles fuel, and I can add fuel densities to our constants as part of that issue!

PR #1113, opened 2026-04-28
Aviary 1.0 API refresh

### LINKED ISSUE #985: NamedValues `get_item()` does not raise error on failure
### Description

When `NamedValues.get_item()` is called, if the requested key does not exist in the data it returns a default (None, None) value and units. Based on docstrings, this is very specifically and intentionally done, but I am not sure why that was the case. I find that it has caused significant confusion and incorrect code because it does not match the behavior of all other methods that take in `key` (like `get_val()`, `delete()`, etc.)

### Example

```
example = NamedValues()
try:
      example.get_val('test', 'unitless')
except KeyError:
      print('key not found')
try:
      example.get_item('test')
except KeyError:
      print('this never happens')
if example.get_item('test') == (None, None):
      print('but this happens instead')
```

### Aviary Version

0.10.0-dev

_No response_

jkirk5:
### Summary

Summary of PR.

### Related Issues

- Resolves #985

`get_items()`, `get_values()`, `get_keys()` functions of `NamedValues` are now methods whose names match Python dictionary methods (`items()`, `values()`, `keys()`). Deprecation warnings are in place for the removed functions.

Removed items from api.py:
- aviary.variable_info.options: **get_option_defaults** (imo this is a dangerous tool and shouldn't be a user-facing function)
- aviary.utils.named_values: **get_keys**, **get_values**, **get_items** (no longer exist, see above)
- aviary.constants: **MU_TAKEOFF**, **MU_LANDING** (#363)

### New Dependencies

None


aviary/api.py
```diff
@@ -11,51 +11,42 @@
...
-    setup_trajectory_params,
 )
 from aviary.utils.merge_hierarchies import merge_hierarchies
 from aviary.utils.merge_variable_metadata import merge_meta_data
-from aviary.utils.named_values import NamedValues, get_keys, get_items, get_values
+from aviary.utils.named_values import NamedValues
 from aviary.utils.aviary_values import AviaryValues
 from aviary.utils.csv_data_file import read_data_file, write_data_file
 from aviary.utils.data_interpolator_builder import build_data_interpolator
-from aviary.variable_info.enums import (
```
Kenneth-T-Moore, line 41: Enums might be useful in here because users import them for their phase_info.

jkirk5: They are still there! See the next added line (`import *`) - I assumed users will want ALL enums, and automatically add them as we create more, although if that is not true we can revert (if some are intended to be developer only)

Kenneth-T-Moore: Oh, they are! yeah, we probably want them all.

Kenneth-T-Moore (approved):

PR #1114, opened 2026-04-29
Models & data cleanup

### LINKED ISSUE #667: Store and access legacy files

The GASP/FLOPS engine files are stored in `utils/test/data` folder. They are not part of Aviary except that Aviary can convert them to Aviary data files. Note that legacy propeller map files are stored in `models/engines/propellers` folder. This may make it hard for a new user to know what is and is not an Aviary file.

1. We want to store all the legacy code files at one place. 
2. Any conversion utilities should have identical behavior on searching for files.

Example: Currently, we can run:

`aviary convert_engine utils/test/data/turboshaft_4465hp.eng turboshaft_4465hp.deck -f GASP_TS`

But the desired behavior is:

`aviary convert_engine turboshaft_4465hp.eng turboshaft_4465hp.deck -f GASP_TS`

_No response_

### Associated Bug Report

_No response_

### LINKED ISSUE #933: Move FLOPS data files to test folder

Right now there are 2 kinds of inputs for FLOPS aircraft: csv files (occasionally), and the data files that create AviaryValues objects for inputs and expected outputs. My proposal is that the models folder should contain only input files, so csvs and maybe a "scripted" input file as an example (but every model should also have a csv). The expected outputs, and the specific way they are stored (a giant dict containing both inputs and outputs) are not user-friendly or set up for easy use in the Aviary L2 api. They were specifically designed for modular testing, so that is the folder they should live in.

So, move all the data files to the `validation_cases/validation_data` folder and make sure that every model still has a csv representation in `models`. For example purposes, a "scripted" input file is still useful. The existing format of data file inputs are not well suited. I would prefer a new data file is created so there aren't 2 copies of a scripted input floating around in both model and test folders. That points to perhaps a GASP-originating file being a good candidate for this example. An actual new example should be created to demonstrate how to create an AviaryProblem by loading inputs from memory in this way.

_No response_

### Associated Bug Report

_No response_

### LINKED ISSUE #1011: Create docs page that shows all avaliable models
### Desired documentation content.

Create an automated docs page that links to all of the models available in Aviary that includes a description of what each file is.

This may require adding a file docstring to each model with the explanation of what they are - this is useful regardless of if it ends up getting used in by the docs page or not

### Is there any existing documentation on your requested topic? Please describe.

_No response_

jkirk5:
### Summary

Reorganized all legacy files, aircraft inputs for testing only, python data input files for testing only, and similar files into `validation_cases/validation_data`.
There are a few Aviary input files (a engine model and an aircraft model) that were used for tests only and I found buried in various test folders. Those have been moved to validation_data as well.

### Related Issues

- Resolves #667 
- Resolves #933 
- Resolves #1011 

None

### New Dependencies

None


aviary/models/missions/height_energy_default.py
jkirk5, line 1: Deleted, seems to be a holdover from the EOM name change

aviary/models/aircraft/multi_engine_single_aisle/multi_engine_single_aisle_data.csv
jkirk5, line 1: Deleted - it was one of the csv's generated from python data files. This one does not run as a standalone model because it needs two engines constructed (and this csv does not include the values for both engines). The python file is also just used for tests right now, so the csv is not necessary.

Kenneth-T-Moore (approved):

PR #1122, opened 2026-04-29
Useful load update

### LINKED ISSUE #1049: Update Calculation and name for Mission.USEFUL_LOAD

cmbenne3:
### Summary

This PR corrects the naming of Mission.USEFUL_LOAD to Mission.OPERATING_ITEMS_MASS, and adds a new component to calculate the correct value of Aircraft.Design.USEFUL_LOAD as Aircraft.Design.GROSS_MASS - Aircraft.Design.EMPTY_MASS.

A flops based test and gasp based test have been modified to ensure that this calculation is checked. It is a very simple calculation so is not added to every possible test.

Whilst looking through the mass calculations I also split the passenger payload mass calculations out of the flops based cargomass() component into their own component for clarity.
ZeroFuelMass is now calculated using the pre-computed TOTAL_PAYLOAD_MASS rather than calculating this again from its constituent parts. Neither of these reformulations change the functionality.

### Related Issues

- Resolves #1049 

None

### New Dependencies

None


aviary/subsystems/mass/flops_based/payload.py
```diff
@@ -33,54 +37,56 @@ def setup_partials(self):
...
+        self.declare_partials(Aircraft.CrewPayload.CARGO_MASS, '*', val=1.0)
+
+    def compute(self, inputs, outputs):
+        wing_cargo = inputs[Aircraft.CrewPayload.WING_CARGO]
+        misc_cargo = inputs[Aircraft.CrewPayload.MISC_CARGO]
+
+        outputs[Aircraft.CrewPayload.CARGO_MASS] = wing_cargo + misc_cargo
+
+
+class PaxPayloadMass(om.ExplicitComponent):
```
Kenneth-T-Moore, line 66: It looks like the PaxPayloadMass component is missing derivatives.

aviary/subsystems/mass/gasp_based/mass_summation.py
```diff
@@ -257,3 +264,21 @@ def compute(self, inputs, outputs):
         operating_mass = inputs[Mission.OPERATING_MASS]
 
         outputs[Mission.ZERO_FUEL_MASS] = operating_mass + payload_mass
+
+
+class UsefulLoadMass(om.ExplicitComponent):
```
jkirk5, line 269: I would prefer a mass build-up of the UsefulLoadMass components (summing masses up) rather than a diff of different masses. It makes it easier to identify what components actually make up the category, and we've been burned by doing these mass diff style computations before so they make me nervous it could get messed up in unexpected ways

aviary/variable_info/variable_meta_data.py
```diff
@@ -1784,6 +1784,16 @@
     default_value=False,
 )
 
+add_meta_data(
+    Aircraft.Design.USEFUL_LOAD_MASS,
+    meta_data=_MetaData,
+    historical_name={'GASP': None, 'FLOPS': None},
+    units='lbm',
+    desc='Useful load of the aircraft calculated as '
+    'Aircraft.Design.GROSS_MASS - Aircraft.Design.EMPTY_MASS',
```
jkirk5: I think this description would be more useful to explain what is in the mass category rather than the diff equation

jkirk5 (commented):
I'm trying to think of a better name than operating items - I feel like "items" is a little awkward but I haven't come up with a satisfactory name yet

aviary/subsystems/mass/flops_based/payload.py
cmbenne3: Do I need derivatives if the component only uses 'options' which can't change and it doesn't have any 'inputs'? Do I need to declare that all the derivatives are '0'? Not sure on the best practice here...

aviary/subsystems/mass/gasp_based/mass_summation.py
cmbenne3: Ok - most of the definitions I've found online actually define it like this - it's the difference in mass between your max gross and what it rolls off the production line as (empty mass). If we would rather build it up then it'll be:
operating_items + payload (for design mission) + fuel (for design mission). Assuming your design mission will be at max gross weight.

aviary/variable_info/variable_meta_data.py
cmbenne3: As a pilot you think of it less as a combination of things that add up, and more as an available 'capacity' that the aircraft has to carry mass. I guess it's a bit like having a fuel tank capacity separately from how much fuel you actually put in the tank. The useful_load could be any combination of operating_items, payload and fuel that results in the aircraft at max gross mass.

aviary/subsystems/mass/gasp_based/mass_summation.py
jkirk5: Oh interesting, that's an official definition! We could keep it as-is then, hopefully we don't encounter any problems with the formulation like we did for fuel mass calcs

aviary/subsystems/mass/flops_based/payload.py
Kenneth-T-Moore: I hadn't noticed that this has no inputs. Yeah, there isn't anything to define for derivatives, then.

Kenneth-T-Moore (approved):

aviary/subsystems/mass/gasp_based/mass_summation.py
cmbenne3: https://pilotinstitute.com/what-is-useful-load/
image

https://www.globeair.com/g/useful-load
image

https://www.faa.gov/sites/faa.gov/files/12_phak_ch10.pdf
image

aviary/variable_info/variable_meta_data.py
cmbenne3: Ok I attempted to calculate useful_load_mass as an addition of things, but then realized that it will incorrectly calculate for off design missions:

image

This will be fine for the design mission, but for off design it will be incorrectly except when the off design mission happens to be at max gross mass as it will use the payload, operating items and fuel from the off design mission and likely calculate a reduced useful mass...

I'm not sure it's worth copying and saving these variables for the design mission just to calculate this quantity...?

jkirk5: Oh I was talking about just updating the description in the metadata and not the actual code. A short version of one of the definitions you posted to my other question would probably be fine (basically just adding the short list of typical items after the formal definition)

jkirk5 (approved):

PR #1124, opened 2026-04-29
Computation of Aircraft.BWB.NUM_BAYS

### LINKED ISSUE #1084: Using the int() function on Aircraft.BWB.NUM_BAYS

For BWB aircraft using FLOPS based geometry, there is a need for the number of bays (`Aircraft.BWB.NUM_BAYS`). It has to be a float because it is needed by the computation of `Aircraft.Furnishings.MASS`. In the computation of `Aircraft.BWB.NUM_BAYS`, `int()` function is used. Here is a comment from Ken:

"For the int calls, I see that those are part of a while loop that solves a nonlinear equation by Gauss-Siedel until it converges. The interesting part is that it solves `int(x) = f(int(x))` instead of `x=f(x)`. if we smooth any of the ints, the algorithm will probably take many more iterations. I wonder if it would still converge. One solution might be to rewrite this using an OpenMDAO solver, solve for a real-valued `num_bays`, then use a smoothed int afterwards. FLOPS did something similar with the skin friction calculation, except there were no ints. I rewrote the equations in residual form and used a Newton solver on them."

I did a test that I removed `int()` in the while loop and made a call to `smooth_int_tanh()` after the loop. This would change a lot of our test results but all of them are minor. This testing modification is not merged into main.

See `BWBDetailedCabinLayout` in `subsystems/geometry/flops_based/fuselage.py` for details.

_No response_

### Associated Bug Report

_No response_

xjjiang:
### Summary

The computation of number of bays (`Aircraft.BWB.NUM_BAYS`) is improved.

### Related Issues

- Resolves #1084 

None

### New Dependencies

None

### DIFF
--- a/.github/actions/prepare_environment/action.yml
+++ b/.github/actions/prepare_environment/action.yml
@@ -4,25 +4,22 @@ inputs:
     NAME:
         required: true
         type: string
-    PY:
+    OPENMDAO_PIXI_ENVIRONMENT:
         required: true
         type: string
-    NUMPY:
-        required: true
-        type: string
-    SCIPY:
-        required: true
-        type: string
-    PYOPTSPARSE:
-        required: true
-        type: string
-    SNOPT:
+    OPENMDAO_INSTALL_FROM:
         required: true
         type: string
     OPENMDAO:
         required: true
         type: string
+    SNOPT:
+        required: true
+        type: string
     DYMOS:
+        required: true
+        type: string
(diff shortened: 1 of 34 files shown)


Kenneth-T-Moore:
After looking this over, I think we are probably better off keeping num-bays as the smoothified integer like it was previously.

jkirk5:
Since we decided to keep the int smoothing as-is, I will close this PR

PR #1130, opened 2026-05-06
Discontinuities in GASP based mass subsystem are removed

xjjiang:
### Summary

There are a few if-else code blocks in GASP based mass subsystems that result in discontinuities. They are removed. More unit tests are added.

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/utils/math.py
+++ b/aviary/utils/math.py
@@ -171,7 +171,7 @@ def d_smooth_max(x, b, mu=10.0):
     m = np.maximum(mu_x, mu_b)
     numerator = np.exp(mu_x - m)
     denominator = np.exp(mu_x - m) + np.exp(mu_b - m)
-    d_sum_log_exp = mu * numerator / denominator
+    d_sum_log_exp = numerator / denominator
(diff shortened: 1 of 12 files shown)


aviary/utils/math.py
```diff
@@ -171,7 +171,7 @@ def d_smooth_max(x, b, mu=10.0):
     m = np.maximum(mu_x, mu_b)
     numerator = np.exp(mu_x - m)
     denominator = np.exp(mu_x - m) + np.exp(mu_b - m)
-    d_sum_log_exp = mu * numerator / denominator
+    d_sum_log_exp = numerator / denominator
```
Kenneth-T-Moore, line 174: I wonder why we never noticed this mistake. I guess it must be because we were never operating near the transitions.

Kenneth-T-Moore (approved):

aviary/utils/math.py
xjjiang: We had set `mu = 1`.

jkirk5 (approved):

PR #1137, opened 2026-05-12
Fix the warning check for the cabin crew calculation.

Kenneth-T-Moore:
### Summary

The warning check for Aircraft.CrewPayload.NUM_CABIN_CREW didn't quite match the FLOPS definition. Note, this variable is not used for any calculation in Aviary or in FLOPS. It is just a computed output that is printed to the FLOPS output file. Aviary allows it to be specified, but we don't use it in our current aircraft concepts.

### Related Issues

- Resolves #

None

### New Dependencies

None


cmbenne3 (approved):

aviary/variable_info/variable_meta_data.py
```diff
@@ -785,7 +785,7 @@
 add_meta_data(
     Aircraft.CrewPayload.NUM_CABIN_CREW,
     meta_data=_MetaData,
-    historical_name={'GASP': None, 'FLOPS': None},
+    historical_name={'GASP': None, 'FLOPS': 'NCABCR'},
```
jkirk5: This isn't an input in FLOPS? It doesn't appear anywhere in the manual. Where does this variable name come from?

Kenneth-T-Moore: I got the name from the source code. It is an internal variable that is only used while printing to the FLOPS output file.

jkirk5: Ah ok - we don't need that in "historical name", since that is a mapping of input files -> Aviary variables for fortran_to_aviary. Right now we don't have an explicit place for these internal variables (some variables have these in comments next to them), we could do the same here

Kenneth-T-Moore: I've moved it to a comment like in previous lines.

cmbenne3 (approved):

PR #1154, opened 2026-05-22
Add CD and Cl to the timeseries.

### LINKED ISSUE #971: Add aerodynamics data to mission timeseries

Add calculated aerodynamics data (including CL, CD) to the mission timeseries (through aero builder's `add_timeseries()`)?

Add angle of attack to timeseries for 2DOF problems. Since 2DOF is getting slowly refactored I don't have a specific suggestion on the best place to implement it.

_No response_

### Associated Bug Report

_No response_

Kenneth-T-Moore:
### Summary

1. Add `Dynamic.Vehicle.LIFT_COEFFICIENT` and `Dynamic.Vehicle.DRAG_COEFFICIENT` to the aircraft data hierarchy.
- This involved some renames and alias promotes.
2. Make sure that lift, drag, and their coefficients are available at the top of the ode, and in the timeseries.
3. Make sure that the initial timeseries inputs from the phases are in alphabetical order.

### Related Issues

- Resolves #971

None

### New Dependencies

None


jkirk5:
My concern with this PR is that it adds a requirement that Aerodynamics outputs both Lift/Drag and CL/CD. If I wrapped an external aero tool and only captured the total forces previously that would have been acceptable, but now would that raise an error?

Kenneth-T-Moore:
> My concern with this PR is that it adds a requirement that Aerodynamics outputs both Lift/Drag and CL/CD. If I wrapped an external aero tool and only captured the total forces previously that would have been acceptable, but now would that raise an error?

We could probably move these into the aerodynamics builder's `get_timeseries` method and allow them to be added that way.

Note that all of our ODEs require Lift and Drag, so if you switch the aero over to calculating a force vector, and don't intend to offer lift and drag as outputs from that subsystem, then you also need to use a different phase_builder/ode/configurator.

jkirk5:
> Note that all of our ODEs require Lift and Drag, so if you switch the aero over to calculating a force vector, and don't intend to offer lift and drag as outputs from that subsystem, then you also need to use a different phase_builder/ode/configurator.

When we get around to generic force vectors I'm planning on adding a "conversion" component to the ODEs to compute the new generic forces so we hopefully don't have to change much.

jkirk5:
> We could probably move these into the aerodynamics builder's `get_timeseries` method and allow them to be added that way.

I like this implementation method better, let the aero comps choose what they have available to add to the timeseries

jkirk5 (approved):

PR #1155, opened 2026-05-26
New documentation on Aviary problem structure

jkirk5:
### Summary

New docs page that explains to beginning users how Aviary sets up analysis, and introduces them to the concept of Pre, Mission, Post groups.

Script that generates XDSM diagrams is included so the diagrams can be easily re-created or changed in the future.

Gen AI (model unknown) was used to help format the pyXDSM diagram to be more readable


aviary/docs/aviary_XDSMs.py
```diff
@@ -0,0 +1,91 @@
...
+full_xdsm.add_system('opt', 'Optimization', 'Optimizer')
+full_xdsm.add_system('pre', 'Function', 'Pre-Mission')
+full_xdsm.add_system('miss', 'Function', 'Mission~Analysis')
+full_xdsm.add_system('post', 'Function', 'Post-Mission')
+
+# 2. Process Execution Flow (This draws the thin routing line!)
+# Shows the loop: Optimizer -> Pre -> Miss -> Post -> back to Optimizer
+full_xdsm.add_process(['opt', 'pre', 'miss', 'post', 'opt'])
+
+# 3. Efull_xdsmternal Inputs (Routed from the left instead of taking up a matrix row)
```
jkirk5:
```suggestion
# 3. full_xdsm Inputs (Routed from the left instead of taking up a matrix row)
```

PR #1163, opened 2026-05-29
Mars + Venus Atmosphere Models

ehariton:
### Summary

Added Mars and Venus atmosphere models accessible just the same way as all the other atmospheres. This is part of the changes required to being simulation of fixed-wing aircraft in the martian / venutian atmospheres. 

### Related Issues

- Resolves #

None

### New Dependencies

None


aviary/subsystems/atmosphere/atmosphere.py
```diff
@@ -102,30 +110,79 @@ def setup(self):
...
+            gamma = 1.36  # Based on averaging values from Hellas_summar, Hellas_winter, Equatorial_summar,
+            # Equatorial_winter, North_Pole_summer, North_Pole_Winter output from Mars-GRAM
+            # Mars atmosphere is 95% Co2 so we use the southerland constant for Co2 https://descanso.jpl.nasa.gov/propagation/mars/MarsPub_sec4.pdf
+            self._S = 222  # (K) Southerlands constant for Mars atmosphere https://doc.comsol.com/5.6/doc/com.comsol.help.cfd/cfd_ug_fluidflow_high_mach.08.27.html
+            self._beta = 1.503e-6  # (s*m*K**(1/2)) viscosity scaling coefficient calculated from other constants listed for C02
+            # https://doc.comsol.com/5.6/doc/com.comsol.help.cfd/cfd_ug_fluidflow_high_mach.08.27.html
+            # self_beta = 1.370**10-5 * (273+self._S)/273**(3/2)
+        else:
+            warnings.warn('self.planet is not set.')
+            exit()
```
jkirk5: If you have to exit afterwards then you should raise an error instead.
Telling the user that a variable that is only used internally is invalid isn't helpful - I'd suggest raising the error after checking for atmosphere model and telling the user an invalid value was provided for Settings.ATMOSPHERE_MODEL

aviary/docs/user_guide_unreviewed/subsystems/atmosphere.ipynb
jkirk5, line 1: Change the top level title back to "Atmosphere Subsystem", and have 2 second-level headers for "Earth Atmosphere Models" and "Mars Atmosphere Models". Have all the generic documentation that applies to the subsystem in general (instead of specific models) under the main header.

aviary/subsystems/atmosphere/atmosphere.py
```diff
@@ -102,30 +110,79 @@ def setup(self):
...
+
+        # Mars altitude output is referenced to the MOLA constant potential surface (areoid) which is already equivalent to geopotential.
+        if self.planet is not 'Earth' and self._geometric is 'geometric':
```
jkirk5: Equations of motion need to use geometric to be technically correct (as 1 ft of physical space is not actually 1 ft in geopotential altitude systems). So this atmosphere will break with every EOM in Aviary right now!

jkirk5 (changes requested):
Need tests to make sure the "mars" path on the logic tree works too

aviary/subsystems/atmosphere/atmosphere.py
Kenneth-T-Moore: The values for this option should just be all of the enum values. Check out flightconditions.py for an example.

aviary/subsystems/atmosphere/atmosphere.py
```diff
@@ -102,30 +110,79 @@ def setup(self):
...
+            M_air = 43.34  # (kg/kmol), mean molar mass of Mars atmosphere https://descanso.jpl.nasa.gov/propagation/mars/MarsPub_sec4.pdf
+            gamma = 1.36  # Based on averaging values from Hellas_summar, Hellas_winter, Equatorial_summar,
+            # Equatorial_winter, North_Pole_summer, North_Pole_Winter output from Mars-GRAM
+            # Mars atmosphere is 95% Co2 so we use the southerland constant for Co2 https://descanso.jpl.nasa.gov/propagation/mars/MarsPub_sec4.pdf
+            self._S = 222  # (K) Southerlands constant for Mars atmosphere https://doc.comsol.com/5.6/doc/com.comsol.help.cfd/cfd_ug_fluidflow_high_mach.08.27.html
+            self._beta = 1.503e-6  # (s*m*K**(1/2)) viscosity scaling coefficient calculated from other constants listed for C02
+            # https://doc.comsol.com/5.6/doc/com.comsol.help.cfd/cfd_ug_fluidflow_high_mach.08.27.html
+            # self_beta = 1.370**10-5 * (273+self._S)/273**(3/2)
+        else:
+            warnings.warn('self.planet is not set.')
```
Kenneth-T-Moore: Should never be able to reach this line of code, so it isn't needed.

aviary/subsystems/atmosphere/atmosphere.py
ehariton: Ok I understand you're saying that EOM always outputs geometric so we must be able to ingest geometric for both Earth and Mars altitudes.

aviary/subsystems/atmosphere/atmosphere.py
ehariton: I ended up removing this warning because you cannot get there on the execution path anyways.

aviary/docs/user_guide_unreviewed/subsystems/atmosphere.ipynb
ehariton: done

ehariton:
> Need tests to make sure the "mars" path on the logic tree works too

Added

Kenneth-T-Moore (approved):

jkirk5 (approved):

PR #1170, opened 2026-06-04
Move MU_LANDING and MU_TAKEOFF out of constants.py

xjjiang:
### Summary

- Replace `MU_TAKEOFF` by `Mission.Takeoff.ROLLING_FRICTION_COEFFICIENT` for GASP based mission.
- Replace  `MU_LANDING` by `Mission.Landing.BRAKING_FRICTION_COEFFICIENT` for GASP based mission.

Note that the defaults (0.025, 0.3) for both variables are from FLOPS. So, in `fortran_to_aviary` run, we need to set their default values from GASP: 0.02 and 0.4.

Partially solved issue #363

### Related Issues

- Resolves #

None

### New Dependencies

None

### DIFF
--- a/aviary/mission/two_dof/ode/flight_path_eom.py
+++ b/aviary/mission/two_dof/ode/flight_path_eom.py
@@ -92,9 +92,10 @@ class FlightPathEOM(om.ExplicitComponent):
                 desc='angle of attack',
                 units='deg',
             )
-        self.add_input(
+        add_aviary_input(
+            self,
             Mission.Takeoff.ROLLING_FRICTION_COEFFICIENT,
-            units='lbm',
+            units='unitless',
             desc='braking friction coefficient',
         )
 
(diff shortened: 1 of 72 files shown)


aviary/mission/two_dof/ode/flight_path_eom.py
```diff
@@ -95,6 +92,11 @@ def setup(self):
                 desc='angle of attack',
                 units='deg',
             )
+        self.add_input(
```
Kenneth-T-Moore: I'd use add_aviary_input for these. This will take care of the units mismatch (shouldn't be lbm)

Another thought. You probably only need to add the input to all of these components if the plane is on the ground (so, groundroll for landing, and groundroll or rotation for takeoff).

xjjiang: Yeah, copy-paste errors occurred. Thanks!

Kenneth-T-Moore (approved):

Issue #1184, opened 2026-06-09
drag for excrescences per component for FLOPS based aero


xjjiang:
Note from D.J. in `SkinFrictionDrag` of `subsystems/aerodynamics/flops_based/skin_friction_drag.py`: Per component not completely implemented in aviary 1.0. (LEAPS) Variable `mission_skin_friction_drag_corrections_count` is a vector over components and is added to the drag. This may be "dead weight" from FLOPS.

D.J.: Moving from LEAPS1 to LEAPS2, we dropped support for the FLOPS skin friction drag corrections table. Esentially, we collapsed the input table to a single scalar input value. This may have been a temporary workaround, but we never came back to it. If people want the level of control offered by the old FLOPS way of doing it, then things need to be refactored a bit in Aviary.

It is likely that we will drop this comment in the source code. 

See note # 1184 in `SkinFrictionDrag` of `subsystems/aerodynamics/flops_based/skin_friction_drag.py`.

PR #1185, opened 2026-06-09
Apply Aircraft.Design.PERCENT_EXCRESCENCE_DRAG to FLOPS based aerodynamics

xjjiang:
### Summary

When `Aircraft.Design.PERCENT_EXCRESCENCE_DRAG` was added to GASP based aerodynamics, it was not applied to FLOPS based aerodynamics where it was named `excrescences_drag` as an option (value: `0.06`). 

This PR replaced `excrescences_drag` by `Aircraft.Design.PERCENT_EXCRESCENCE_DRAG`. It is changed to an aviary variable. The default value is kept as `0.0`. Users must set its value in their dataset (value: `0.06`).

FLOPS goes to detailed table for different components. See issue #1184 for details.

### Related Issues

- Resolves #

None

### New Dependencies

None


Kenneth-T-Moore (approved):

jkirk5 (changes requested):
Putting a quick block on merging - we can't loose the 0.06 FLOPS default because that will cause all future models using the method to produce incorrect answers

Kenneth-T-Moore:
> Putting a quick block on merging - we can't loose the 0.06 FLOPS default because that will cause all future models using the method to produce incorrect answers

Do we currently have any other vars or options that have different defaults between the 2 code-origins?

jkirk5:
> > Putting a quick block on merging - we can't loose the 0.06 FLOPS default because that will cause all future models using the method to produce incorrect answers
> 
> Do we currently have any other vars or options that have different defaults between the 2 code-origins?

I think we have some, but they are easily handled - typically the Aviary default is zero, and the different FLOPS/GASP defaults are added to the input file during file conversion. This is stuff where using the same default as the legacy code is important for matching results with that code, but aren't fundamental to the method itself.

This is a weird case, because in FLOPS the excrescence drag factor isn't actually an input, so that 0.06 seems like it may have been hardcoded into the method. So it's not just a matter of setting the variable during file conversion - it's something that needs to be happening for all Aviary models using EDET, whether they were converted from FLOPS or written from scratch so the results match what that method is expected to produce.

There is a unrelated potential issue that currently the excrescence drag factor has a default of 0, and none of our GASP models set that. So that drag component ends up always being zero in GASP aero which seems unintended.

jkirk5:
TLDR: This default is important because without it, the same set of inputs given to EDET vs. our Aviary implementation will produce different outputs unless the user knows to specifically add a certain variable with a specific value (which isn't normally an input to the method), which isn't helpful. This is different than most FLOPS/GASP disagreements because those are about which value an input has, and that is fine because we just care about having the correct outputs for a given set of inputs

aviary/utils/preprocessors.py
```diff
@@ -47,6 +47,13 @@ def preprocess_options(
             verbosity = meta_data[Settings.VERBOSITY]['default_value']
             aviary_options.set_val(Settings.VERBOSITY, verbosity)
 
+    if Settings.AERODYNAMICS_METHOD in aviary_options:
+        aero_method = aviary_options.get_val(Settings.AERODYNAMICS_METHOD)
+    else:
+        raise UserWarning(
+            'AERODYNAMICS_METHOD not specified. Cannot preprocess fuel aerodynamic inputs.'
+        )
+
```
jkirk5:
```suggestion
```

aviary/utils/preprocessors.py
```diff
@@ -82,6 +89,13 @@ def preprocess_options(
                         f'Aircraft.Wing.THICKNESS_TO_CHORD ({tc}).'
                     )
 
+    if Aircraft.Design.PERCENT_EXCRESCENCE_DRAG not in aviary_options:
+        # In FLOPS, excrescence drag percentage is not able to be set via the input file
+        # Therefore, it appears to have been hardcoded into the method
+        # Here we set the default value to that fixed value
+        if aviary_options.get_val(Settings.AERODYNAMICS_METHOD) is LegacyCode.FLOPS:
```
jkirk5:
```suggestion
        if (
                Settings.AERODYNAMICS_METHOD in aviary_options and 
                aviary_options.get_val(Settings.AERODYNAMICS_METHOD) is LegacyCode.FLOPS
        ):
```

jkirk5 (approved):

PR #1202, opened 2026-06-23
Standardize some aviary function call

xjjiang:
### Summary

- Replaced `get_path(phase_info)` by `get_path(phase_info, verbosity)`.
- Replaced `self.add_input` by `add_aviary_input` in mission.
- Replaced `self.add_output` by `add_aviary_output` in mission.

### Related Issues

- Resolves #

None

### New Dependencies

None


jkirk5 (changes requested):
I handled the `tag` argument differently in #1206, where I cover for all additional arguments we don't directly handle in `add_aviary_input`, `add_aviary_output` using `kwargs`. So we won't need the added `tag` argument once that is merged in.

xjjiang:
> I handled the `tag` argument differently in #1206, where I cover for all additional arguments we don't directly handle in `add_aviary_input`, `add_aviary_output` using `kwargs`. So we won't need the added `tag` argument once that is merged in.

Done.

Kenneth-T-Moore (approved):

jkirk5 (approved):

PR #1203, opened 2026-06-24
Phase linking overhaul.

Kenneth-T-Moore:
### Summary

Prior to this refactor, all of the phase linkages were added by custom code written independently for each problem origin (gasp, flops, solved twodof) and encapsulated in the configurators. This PR implements a general phase-linking that replaces all of these, and paves the way for a future capability of selecting the ODE individually in each phase.

1. General phase linking capability implemented in aviary_group.
2. The `get_linked_variables` method has been added to the phase builder base class. This method allows phases to define which variables should be linked.
3.  For analytic phases that need an initial parameter connected upstream, the phase linking supports this. For a variable like "mass"
- The input should be called "initial_mass"
- The `get_linked_variables` should contain "initial_mass" instead of "mass".
4. The phase_info has been updated to provide a new key for time/states/controls that gives the user control over whether the link is via a direct connection or a constraint. For a variable like mass, this key is called `mass_direct_link`.
- Default values for these are specified in the default keys for each phase. Some linkages were handled differently between the problem types, and that custom code has been replaced by these default values.
5. The former `link_phases` method on the problem configurators has been renamed to `link_trajectory`. This method now just performs any problem-specific linkage with components external to the trajectory.
6. When linkage constraints are created, the ref and ref0 now come from the ref/ref0 defined in the upstream or downstream phase. Formerly, they came from initial conditions or hardcoded values.
7. The new phase linking no longer creates a constraint between two ends that are already constrained by `initial_` and `final_`.

### Related Issues

- Resolves #

There is a possibility that your model may run a bit differently due to some scaling changes and some small differences in the resulting linkages (particularly downstream of a Breguet range phase). The new phase keys should restore some of this.

### New Dependencies

None


aviary/core/aviary_group.py
```diff
@@ -1147,67 +1146,207 @@ def link_phases(self, verbosity=None, comm=None):
...
+        # Assemble all linkable variables in the phases.
+        link_vars_dict = {}
+        for builder in self.phase_objects:
+            phase_name = builder.name
             phase_info = self.mission_info[phase_name]
             all_subsystem_options = phase_info.get('subsystem_options', {})
 
-            lists_to_link.append([])
+            link_vars = set()
             for subsys in self.external_subsystems:
```
jkirk5: Should we just iterate over all subsystems (I believe that should be `self.subsystems`), just in case in the future we have a core subsystem use linked variables? Just a future-proofing idea while we're here

aviary/subsystems/subsystem_builder.py
```diff
@@ -197,6 +197,10 @@ def get_linked_variables(self, aviary_inputs=None, user_options=None, subsystem_
         Return a list of variable names that will be linked when this phase is connected to another
         phase that shares the variable.
 
+        If you have an analytic phase, and you need to link an input parameter to the upstream
```
jkirk5: The subsystem designer will often have no knowledge of whether their subsystem will be used with an analytic phase or not, so I'm not sure how this can be used in practice

aviary/mission/energy_state_problem_configurator.py
```diff
@@ -228,74 +223,20 @@ def set_phase_options(self, aviary_group, phase_name, phase_idx, phase, user_opt
             **extra_options,
         )
 
-    def link_phases(self, aviary_group, phases, connect_directly=True):
+    def link_trajectory(self, aviary_group, phases):
```
jkirk5: The name change confuses me - `link_trajectory` makes me think there are trajectories being linked together? While `link_phases` made more sense to me, as it is the individual phases that are being linked together

aviary/subsystems/subsystem_builder.py
Kenneth-T-Moore: This part of the docstring should only be in the phase builder. It doesn't apply to external subsystems. I will remove it.

aviary/core/aviary_group.py
Kenneth-T-Moore: I don't think there's any reason not to do that.

aviary/mission/energy_state_problem_configurator.py
Kenneth-T-Moore: The configurator method doesn't actually create links between phases anymore. It just adds components and manually connects to things that are upstream or downstream of the trajectory -- stuff that isn't so generalizable. We could go with something like `configure_trajectory`.

jkirk5 (approved):

PR #1206, opened 2026-06-26
Simplified API for input/output helper functions

### LINKED ISSUE #1196: Any custom defaults set in components are overwritten by metadata defaults
### Description

The AddAviaryInput and AddAviaryOutput functions have a `val` option, to set what the default value for that input/output should be in the case it is not provided via connection (including from the IVC). However, in the AviaryGroup's configure step, we run set_input_defaults on _every_ variable in the metadata, using the default values and units provided there. That is done to resolve a handful of specific issues with ambiguous units or values for a small number of cases (about half a dozen).

This is a problem because it squashes any use of 'val' in AddAviary*(), making that part of the api nonfunctional plus it is not obvious to the user what happened.

There are legitimate reasons to want to use component-level defaults. This recently came up with EDET, which originally had a hardcoded PERCENT_EXCRESCENCE_DRAG of 0.06. This is a rare instance of a method-only default, we currently aren't making it a global Aviary default in metadata since that would affect the GASP aerodynamics calculations. The cleanest way to handle this would be to have the excrescence drag locally defaulted in the one component that uses it, to make it clear that the method only wants that specific default, and makes sure it gets applied whenever the method is used, including as a stand-alone component without using the preprocessors. That currently doesn't work because of the AviaryGroup default overwrite.

(I am not sure if this issue extends to AddAviaryOption as well)

### Example

_No response_

### Error Message

_No response_

### Aviary Version

1.0.1-dev

### Environment Information

_No response_

jkirk5:
### Summary

Removed the `val` argument from the helper functions `add_aviary_input` and `add_aviary_output`, which previously was always being reset to the metadata default regardless. 

Claude Haiku 4.5 was used to change the use of inputs.values() to individual calls to get the values of individual inputs in gaspaero for significantly improved readability/traceability/search-ability of the code.

### Related Issues

- Resolves #1196

I stumbled upon `opt2vals` in the 2dof taxi and it appears to be redundant, so I removed it. That is the only place it is being used and seems like it can just be directly set as an input (via the input csv) rather than an option that then becomes an input.

### New Dependencies

None


aviary/mission/two_dof/ode/v_rotate_comp.py
```diff
@@ -8,21 +8,20 @@
 class VRotateComp(om.ExplicitComponent):
     """
     Component that computes V_rotate based on vehicle properties and speed buffers.
-    Note: This component is not used.
+    NOTE: This component is not used.
     """
 
     def setup(self):
-        # Temporarily set this to shape (1, 1) to avoid OpenMDAO bug
```
Kenneth-T-Moore, line 15: This component appears to be unused outside of its unit test, which is why it has dynamic inputs that aren't shaped by num_nodes. I think we can delete it and its test.

EDIT: We actually replaced the component with an excecomp, so yeah, it is redundant now.

EDIT: WE might need to bring the v_rotation component back, because gravity is going to become an option instead of a bare input. This overlaps with Eliot's story, and with mine too, since I need to move that component into the takeoff ode. Probably leave it there for now.

aviary/mission/two_dof/ode/taxi_ode.py
```diff
@@ -19,16 +18,6 @@ def setup(self):
         subsystems = self.options['subsystems']
         user_options = self.options['user_options']
 
-        # NOTE calling add_opts2vals in this way makes Mission.Taxi.MACH a required variable.
-        # This is because ODE was relying on old behavior of NamedValues.get_item() which returns
-        # (None, None) in option_to_var.py to avoid errors.
-        # Pulling the default from the MetaData if it does not exist instead.
-        # TODO find a better way to handle this
-        if Mission.Taxi.MACH not in options:
-            default_val = CoreMetaData[Mission.Taxi.MACH]['default_value']
```
Kenneth-T-Moore, line 28: Mission.Taxi.MACH is still an option, and we don't do a set_val on options from the Aviary top, so I think you still need to get the value out of the metadata or `options` and set it as the default value for `taxi_mach` in the alias_comp definition below this. Otherwise, taxi mach is 0.0 as per the set_input_defaults at the bottom (which also should be removed.)

jkirk5: I made sure taxi_mach is no longer flagged as an option in metadata

Kenneth-T-Moore (approved):

Issue #1191, opened 2026-06-11
Remove RHO_SEA_LEVEL constant


jkirk5:
### Desired capability or behavior.

The constant `RHO_SEA_LEVEL` appears to be unnecessary and can be removed.

### Suggested implementation.

Currently the constant is being used in a couple different ways. First is as a hardcoded air density for some components that assume sea level altitude such as energy takeoff/landing - these components should be updated to use the full atmosphere component and accept any altitude. Second is in several places throughout the code to compute a density ratio of actual density / sea level density. This also shouldn't be hardcoded (since sea-level density changes depending on which atmosphere model is being used). I don't have a implementation idea for replacing the use of the constant in this area yet.

PR #1215, opened 2026-07-02
Remove Wing area from Simple Landing builder defaults

### LINKED ISSUE #980: Landing group should not require Aircraft.Wing.AREA as an input in a model
### Description

In `aviary/mission/height_energy_problem_configurator.py`, the line `self._add_landing_systems(aviary_group)` will fail if `Aircraft.Wing.AREA` is not given in a model. Per Ken, we only need the wing area in that group to set an input default on the landing group. We should just remove the value from that set_input_defaults call, and keep the unit. That way, we can remove the requirement for caching wing_area in the configurator/group.

### Example

Use `models/aircrafts/blended_wing_body/bwb_simple_FLOPS.csv` to run aviary and make sure you call function `self._add_landing_systems(aviary_group)`.

### Aviary Version

0.10.0.dev0

_No response_

### LINKED ISSUE #1221: Incorrect assignment of legacy approach velocities to Mission.Landing.INITIAL_VELOCITY
### Description

Mission.Landing.INITIAL_VELOCITY is incorrectly assigned to similar, but not identical historical variables in GASP & FLOPS:

Image

VGL in GASP is an internally computed local variable within the subroutine DLAND. It is not an input available to the user.

VAPPR is a constraint control variable in FLOPS, not an input or override, Aviary does not handle this constraint control and issue #1222 tracks this feature. Extract from FLOPS manual detailing VAPPR behavior:

Image

VAPPR is also input in kts, but gets converted to ft/s by Aviary.

Mission.Landing.INITIAL_VELOCITY could be mapped to the output variable VAPP, which appears in the Objectives/Variables/Constraints Summary section of FLOPS output file:

Image

Mapping to an output will have no affect on Aviary functionality, but would help translate variables for legacy code users.

### Example

_No response_

### Error Message

_No response_

### Aviary Version

1.0.2-dev

### Environment Information

_No response_

cmbenne3:
### Summary

This PR addresses issue 980 and removes the set input default calls for the energy state landing group, which were unnecessary.
This simplifies the configurator and landing builder.

The landing component is updated to remove hardcoded density/unit conversion for calculation of Landing.INITIAL_VELOCITY. OpenMDAO now handles units and the dynamic density input is used.

Incorrect historical names removed from meta data for Landing.INITIAL_VELOCITY variable and replaced with the closest equivalent local and output variables from GASP and FLOPS. These mappings have no impact on functionality since neither ever appear in input files, but are included for completeness and to assist legacy code users with understanding the Aviary variable hierarchy.

### Related Issues

- Resolves #980 
- Resolves #1221 

Any custom scripts using the Landing class builder will fail.

FLOPS made the assumption that altitude = sea level for calculation of Landing.INITAL_VELOCITY (VAPP). This has now been updated to correctly use Dynamic.Atmosphere.DENSITY so will now differ from FLOPS values where landing altitude is greater than sea level.

VAPPR will no longer be converted from FLOPS input files with the fortran_to_aviary() function (however Aviary wasn't correctly interpreting this conversion anyway)

### AI Usage

AI was used for assistance with changes to compute_partials() for the landing component.


aviary/mission/energy_state/phases/simplified_landing.py
```diff
@@ -20,7 +20,7 @@ def setup(self):
             desc='atmospheric density',
         )
 
-        add_aviary_input(self, Aircraft.Wing.AREA, val=700)
+        add_aviary_input(self, Aircraft.Wing.AREA, val=1370)
```
jkirk5: Use of `val` here doesn't actually do anything and will be removed soon (see #1196, #1206). You'll need to update the test to directly set that value to get matching results instead

jkirk5 (changes requested):

aviary/mission/energy_state/phases/simplified_landing.py
cmbenne3: Resolved - val removed from add_aviary_input call. We use set_input_defaults() in the tests instead

xjjiang:
Can you add Carl's comment on GASP overriding feature into `source_docs/variable_metadata.ipynb`?

aviary/mission/energy_state/phases/build_landing.py
```diff
@@ -56,15 +24,8 @@ def build_phase(self, use_detailed=False):
...
-        ##############
-        # Add Inputs #
-        ##############
-
-        landing = LandingGroup()
-        landing.set_input_defaults(Aircraft.Wing.AREA, val=self.ref_wing_area, units='ft**2')
-        landing.set_input_defaults(
-            Mission.Landing.LIFT_COEFFICIENT_MAX, val=self.Cl_max_ldg, units='unitless'
-        )
+        else:
```
Kenneth-T-Moore, line 27: The else isn't technically needed because the `if` branch raises an exception, and never proceeds to here.

aviary/mission/energy_state/phases/test/test_building_landing.py
```diff
@@ -35,8 +36,10 @@ def test_case1(self):
         assert_check_partials(partial_data, atol=1e-12, rtol=1e-12)
 
         tol = 1e-6
-        assert_near_equal(prob[Mission.Landing.GROUND_DISTANCE], 6332.4878059, tol)
-        assert_near_equal(prob[Mission.Landing.INITIAL_VELOCITY], 134.9752, tol)
+        assert_near_equal(prob[Mission.Landing.GROUND_DISTANCE], 6332.13214907, tol)
```
Kenneth-T-Moore, line 39: What causes the small differences in the asserts in these couple of tests?

aviary/mission/energy_state/phases/build_landing.py
cmbenne3: I was preparing for when we do add something for use_detailed = True the landing group would then be overwritten with the simple landing group, unless the simple landing group is moved into an else...?
Happy to remove the else for now and add it at that future point though - what do you think is best?

Kenneth-T-Moore: In that case, don't worry about it for now.

aviary/mission/energy_state/phases/test/test_building_landing.py
cmbenne3: [list_vars_for_landing_group_test.txt](https://github.com/user-attachments/files/29866369/list_vars_for_landing_group_test.txt)

There are a couple of very small differences:
1. The original test had a default altitude of 1m for some reason. The test now uses an altitude of 0. This impacts the landing distance.
2. I switched from using the RHO_SEA_LEVEL_METRIC to RHO_SEA_LEVEL_ENGLISH. The metric value is saved as 1.225, but the English version = 0.0023769. When you query the atmosphere model for density at 0 altitude you obtain a very slightly different ratio for density/constant when using english vs metric units which impacts the landing and initial_velocities.

Here is a table detailing the old and new methods, differences and why the test values needed to change - cells in green show the old and new test values.

Image

I am putting the remaining discrepancies beyond the 4th decimal down to some additional rounding/precision somewhere.

For completeness see attached for 3 sets of prob.model.list_vars() for this test showing original, new, and a modified new with altitude set = 1m to prove the component is working as expected.

cmbenne3: These discrepancies would be reduced with a solution to #1191 (although the test would still change since the altitude has been updated)

Kenneth-T-Moore: Ok, this looks reasonable. I'm working on 1191 presently.

Kenneth-T-Moore (approved):

cmbenne3:
> Can you add Carl's comment on GASP overriding feature into `source_docs/variable_metadata.ipynb`?

There is already some more information on this in develop_metadata.py:
image

If I understand Carl's comment correctly it is actually possible to overwrite any internal GASP variable by adding it to the input file outside of the namelists. If this is the case and Aviary wants to support this functionality then either fortran_to_aviary needs updating so it understands the full convention described here, or we need to simplify the conventions for GASP variables and only have 2 categories: <name_list>.<var_name> for variables in input namelists, and <var_name> for all other variables that could be listed outside of namelists in the input file and used to overwrite internal variables.

I have not made any additional changes to this PR, but have written a new issue #1229 to investigate overwriting of GASP variables outside input namelists.

jkirk5 (approved):

Issue #1192, opened 2026-06-11
Update values of constants


jkirk5:
### Desired capability or behavior.

The `constants.py` file should be updated to use the following values for constants. Similar constants (that differ by code or with metric/english versions) will be replaced by a single value. The preference for constants is to use SI base units. We should use the openmdao unit conversion utilities to change to English units where needed. The source for the constants should be included as a comment for each entry. All other constants not mentioned in the list below should be removed.

## Constants
GRAV_EARTH
Val: 9.80665
Units: m/s^2
Source: NIST (https://physics.nist.gov/cgi-bin/cuu/Value?gn|search_for=gravity)
Notes: A variable for gravitational acceleration needs to be added to the Aviary hierarchy and set in preprocessing based on which atmospheric model is being used.

GRAV_MARS
Val: 3.712
Units: m/s^2
Source: GRAM constants

GRAV_VENUS
Val: 8.87
Units: m/s^2
Source: GRAM constants

RADIUS_EARTH
Val: 6,371,009
Units: m
Source: GRS80, mean earth radius (rounded to nearest meter)
Notes: A variable for planetary radius needs to be added to the Aviary hierarchy and set in preprocessing based on which atmospheric model is being used.

RADIUS_MARS
Val: 3,386,200
Units: m
Source: GRAM constants, avg of equatorial and polar radii

RADIUS_VENUS
Val: 6,051,800
Units: m
Source: GRAM constants, avg of equatorial and polar radii

RHO_SEA_LEVEL
Notes: This will be addressed as part of issue #1191. For this issue, it can be kept as-is.

jkirk5:
@ehariton can you copy the name of the source for planetary data GRAM mentions here? We should attribute the source to that rather than to GRAM

ehariton:
GRAM data references NASA websites that no longer host the information :(

MARS:
Planetary constants (radius, gravity, etc.) are from the NASA Space Science Data Coordinated Archive Planetary Fact Sheet for Mars Web page:
https://nssdc.gsfc.nasa.gov/planetary/factsheet/marsfact.html


VENUS:
Planetary constants (radius, gravity, etc.) are from the NASA Space Science Data Coordinated Archive Planetary Fact Sheet for Venus Web page: https://nssdc.gsfc.nasa.gov/planetary/factsheet/venusfact.html.

jkirk5:
> GRAM data references NASA websites that no longer host the information :(

[...]

PR #1219, opened 2026-07-07
Gravity update foundation

ehariton:
### Summary

Added Mission.GRAVITY. Did not start using this in any components yet. That will be a follow-on PR.

### Related Issues

Partial fix for #1192, but more work needs to be done to complete replace constants.

None

### AI Usage


aviary/subsystems/atmosphere/atmosphere.py
```diff
@@ -117,60 +104,26 @@ def setup(self):
...
+        M_air = 0  #
+        gamma = 0  #
 
         # The constants below are used as a simplification to enable calculation of properties not given by by source data tables
         if self.planet == 'Earth':
             M_air = 28.97  # (kg/kmol), molar mass of dry air on Earth
             gamma = 1.4  # Ratio of specific heats
             self._S = 110.4  # (K) Southerlands constant for Earth air
             self._beta = 1.458e-6  # (s*m*K**(1/2)) viscosity scaling coefficient
-            self._R0 = 6_356_766  # (meters) The effective Earth Radius
```
Kenneth-T-Moore, line 173: I think the tests failures come from the change in values here, from 6356766 to 6371009.  Old value was polar radius, and new value is mean.

Kenneth-T-Moore:
This seems like a good approach that we could also use for sea level density in the components where it is used.

Kenneth-T-Moore (approved):

aviary/constants.py
```diff
@@ -2,9 +2,36 @@
...
+#     3386200,
+#     'm',
+# )  # Mars Global Reference Atmospheric Model (Mars-GRAM) 2024: User Guide, NASA/TM-20240012934, avg of equatorial and polar radius
+RADIUS_MARS = (3396200, 'm')  # TODO remove and replace with above
+RADIUS_VENUS = (
+    6051800,
+    'm',
+)  # Venus Global Reference Atmospheric Model (Venus-GRAM): User Guide, NASA/TM-20210022168, avg of equatorial and polar radius
+
+GNS = 9.8236930  # grav_accel_at_surface_earth # remove this asap
```
jkirk5: Why was this added in the first place? It's not used anywhere, can we remove it right now?

ehariton: I commented this line out but left it as a TODO.

GNS is used in a utils.py file. It was hard coded and didn't reference the constants file so it skipped under the radar. 

I put it here in constants so we would remember to change it.

PR #1228, opened 2026-07-09
GASP mass fixed.py and fuel.py updates

xjjiang:
### Summary

1. Added non-smooth computation of `c_eng_pos` in `MassParameters()` in `mass/gasp_based/fixed.py`.
2. Added non-smooth computation of `max_wingfuel_wt` in `FuelComponents()` in `mass/gasp_based/fuel.py`.
3. Fixed a bug where there are mixed wing engines and fuselage engines.
4. Added unit tests for the new implementation.
5. Use ~~`Aircraft.Engine.NUM_WING_ENGINES`~~ `Aircraft.Propulsion.TOTAL_NUM_WING_ENGINES` directly in `MassParameters()` class.
6. Do not set `Aircraft.Engine.POSITION_FACTOR` as multi-valued. In fact, it is renamed to `Aircraft.Propulsion.ENGINE_POSITION_FACTOR`.

### Related Issues

- Resolves #

None

### AI Usage

None


aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -90,11 +91,22 @@ def setup_partials(self):
...
         sweep_c4 = inputs[Aircraft.Wing.SWEEP]
         taper_ratio = inputs[Aircraft.Wing.TAPER_RATIO]
         AR = inputs[Aircraft.Wing.ASPECT_RATIO]
         wingspan = inputs[Aircraft.Wing.SPAN]
         num_engines = self.options[Aircraft.Propulsion.TOTAL_NUM_ENGINES]
+        num_fuselage_engines = self.options[Aircraft.Engine.NUM_FUSELAGE_ENGINES]
+        num_wing_engines = num_engines - num_fuselage_engines
+        if num_wing_engines > 4 or num_wing_engines < 0:
+            if verbosity > Verbosity.BRIEF:
+                print(
```
jkirk5: If this case isn't supported then should we raise an error instead?

aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -90,11 +91,22 @@ def setup_partials(self):
...
+        smooth = self.options[Aircraft.Design.SMOOTH_MASS_DISCONTINUITIES]
+        verbosity = self.options[Settings.VERBOSITY]
+
         sweep_c4 = inputs[Aircraft.Wing.SWEEP]
         taper_ratio = inputs[Aircraft.Wing.TAPER_RATIO]
         AR = inputs[Aircraft.Wing.ASPECT_RATIO]
         wingspan = inputs[Aircraft.Wing.SPAN]
         num_engines = self.options[Aircraft.Propulsion.TOTAL_NUM_ENGINES]
+        num_fuselage_engines = self.options[Aircraft.Engine.NUM_FUSELAGE_ENGINES]
+        num_wing_engines = num_engines - num_fuselage_engines
```
jkirk5: `Aircraft.Engine.NUM_WING_ENGINES` is directly available so we should use that instead of re-calculating

aviary/subsystems/mass/gasp_based/fixed.py
xjjiang: If `num_wing_engines > 4`, issue a warning and continue. If `num_wing_engines < 0`, raise an error.

aviary/subsystems/mass/gasp_based/fixed.py
xjjiang: Done.

aviary/subsystems/mass/gasp_based/wing.py
```diff
@@ -418,7 +418,7 @@ def setup(self):
         add_aviary_input(self, Aircraft.Wing.ULTIMATE_LOAD_FACTOR, units='unitless')
         add_aviary_input(self, Aircraft.Wing.MASS_COEFFICIENT, units='unitless')
         add_aviary_input(self, Aircraft.Wing.MATERIAL_FACTOR, units='unitless')
-        add_aviary_input(self, Aircraft.Engine.POSITION_FACTOR, shape=num_engine_type)
```
jkirk5, line 421: This breaks this component when multiple engines are used - this variable has to remain engine-specific. Instead we need to look for ways to update the component to support a vector of engine positions. In this case maybe we average the wing engine positions together? It is not clear to me what "position factor" actually represents in GASP, so I am not sure what the correct answer is yet. How did GASP handle 4-engine aircraft like the 747 with multiple engine positions on the wing?

aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -57,7 +55,6 @@ def setup(self):
         add_aviary_output(
             self,
             Aircraft.Engine.POSITION_FACTOR,
```
jkirk5: I already commented on `wing.py` about this variable change, my comment there applies here too! We need to update the component to work with a vector of engine positions instead of assuming engine position can never be a vector

aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -122,18 +126,38 @@ def compute(self, inputs, outputs):
...
-            c_eng_pos = 0.98 * sigmoidX(max_mach, 0.75, -1.0 / 320.0) + 0.95 * sigmoidX(
-                max_mach, 0.75, 1.0 / 320.0
-            )
-        if not_fuselage_mounted and num_engines == 4:
-            c_eng_pos = 0.95 * sigmoidX(max_mach, 0.75, -1.0 / 320.0) + 0.9 * sigmoidX(
+        if smooth:
+            c_eng_pos = 1.0 * sigmoidX(max_mach, 0.75, -1.0 / 320.0) + 1.05 * sigmoidX(
                 max_mach, 0.75, 1.0 / 320.0
             )
+            # GASP code has an error. We changed to the following. If num_wing_engines = 3, then
```
jkirk5, line 134: Can we extend this method to allow for more engines? Some ideas I have are having it handle odd or even engines instead of hardcoding 2 vs 3 vs 4 engines (which is what FLOPS does), or simply have the last `c_eng_pos` calculation apply for any number of engines >= 4?

aviary/variable_info/variable_meta_data.py
```diff
@@ -2311,7 +2311,6 @@
     units='unitless',
     desc='engine position factor',
     default_value=0,
-    multivalue=True,
```
jkirk5, line 2314: We can't remove this flag - all engine variables need to use the multivalue flag so Aviary allows them to be vectors. I have some comments on the places in the code where this variable gets used with suggestions on how to update those components

jkirk5 (changes requested):
We can't remove the `multiengine` flag from any `Aircraft.Engine` variables or it could break aircraft with more than one engine type, I have some suggestions to fix

aviary/subsystems/mass/gasp_based/wing.py
xjjiang: In GASP, the corresponding variable is `SKEPOS`. Here is an answer from Jeff:

> When the engines are mounted on the wings, they provide inertial relief to the wing root bending moment (primary driver of the wing weight).  The more the engine number/weight or the farther out-board, the more the reduction in the bending moment.  `SKEPOS` is a factor in the wing weight estimation equation to account for this effect.  It depends only on the aircraft type (thru the cruise Mach No.), number of engines and if the engines are mounted on the wing (`YP>0`).

My interpretation is that it depends on the number of wing engines (regardless the types of wing engines and locations of wing engines). For the case `max Mach < 0.75`, it holds `SKEPOS = 0.98` if there are two wing engines, and `SKEPOS = 0.95` if there are four wing engines. It is used to compute the first guess of wing weight. It is just a little adjustment to an empirical equation. Do you still think we need to consider multiple engine types? Maybe `Aircraft.Engine.POSITION_FACTOR` is not a good name.

jkirk5: Ok, my proposal is that we rename `Aircraft.Engine.POSITION_FACTOR` to `Aircraft.Propulsion.ENGINE_POSITION_FACTOR`, so it is a single value and no longer implied to be per engine type. `MassParameters` (fixed.py) will need to be updated to account for multiple engines - use `Aircraft.Propulsion.TOTAL_NUM_WING_ENGINES` (instead of `Aircraft.Engine.NUM_WING_ENGINES` which is a vector)

xjjiang: Done.

aviary/subsystems/mass/gasp_based/fixed.py
```diff
@@ -90,11 +87,20 @@ def setup_partials(self):
...
         taper_ratio = inputs[Aircraft.Wing.TAPER_RATIO]
         AR = inputs[Aircraft.Wing.ASPECT_RATIO]
         wingspan = inputs[Aircraft.Wing.SPAN]
-        num_engines = self.options[Aircraft.Propulsion.TOTAL_NUM_ENGINES]
+        num_wing_engines = self.options[Aircraft.Propulsion.TOTAL_NUM_WING_ENGINES]
+        if num_wing_engines > 4:
+            if verbosity > Verbosity.BRIEF:
+                warnings.warn(
+                    f'The case num_wing_engines = {num_wing_engines} is not currently '
+                    'supported in Aviary. Will set Aircraft.Propulsion.ENGINE_POSITION_FACTOR using default.'
```
jkirk5:
```suggestion
                    f'GASP mass equations do not support more than 4 total engines on the wings. '
                    'Aircraft.Propulsion.ENGINE_POSITION_FACTOR will use the default equation, '
                    'which is not scaled by number of wing-mounted engines.'
```

jkirk5 (approved):

PR #1233, opened 2026-07-14
Sea Level Density overhaul -> Requires OpenMDAO bugfix

### LINKED ISSUE #1191: Remove RHO_SEA_LEVEL constant

Kenneth-T-Moore:
### Summary

Remove the RHO_SEA_LEVEL_ENGLISH and RHO_SEA_LEVEL_SI constants that were used in various places and had different values.
- In places where this was used in the denominator of a density ratio, it was replaced with an option `Mission.SEA_LEVEL_DENSITY`. This option will be set automatically by the preprocessor if it isn't specified in the csv file. Like gravity, the preprocessor will compute it from the atmosphere model that is chosen.
- In places where Density is meant to be the current state, it is now pulled from an existing atmosphere component instead.

In 2DOF, thee VRot execcomp, which computes the speed at which rotation starts, has been moved into the groundroll phase along with its constraint.
- The execcomp has been replaced with an `ExplicitComponent` in preparation for the gravity overhaul part 4.
- The density input now uses the local atmosphere instead of a hard-coded value.
- The mass input now uses the mass at rotation instead of at taxi end.

This work has exposed a bug in OpenMDAO when connecting an input to another input. This requires a bugfix that is currently a PR, but will soon be merged, and will eventually become part of a release.  The CI has been configured to load OpenMDAO from github in all workflows for now.

### Related Issues

- Resolves #1191

None

### AI Usage

None


aviary/mission/two_dof/ode/takeoff_ode.py
```diff
@@ -70,7 +71,8 @@ def setup(self):
             Dynamic.Mission.VELOCITY,
             Dynamic.Mission.FLIGHT_PATH_ANGLE,
         ] + ['aircraft:*']
-        if not self.options['ground_roll']:
+
+        if not ground_roll:
```
ehariton, line 75: did you mean to make this change? It doesn't seem like `ground_roll` is defined.

Kenneth-T-Moore: It is defined above `ground_roll = self.options['ground_roll']`, we just weren't using it.

ehariton: ok great I just didn't see it for some reason!

aviary/subsystems/propulsion/propeller/test/test_hamilton_standard.py
```diff
@@ -162,7 +161,7 @@ def test_postHS(self):
         )
         prob.set_val(
             Dynamic.Atmosphere.DENSITY,
-            np.array([1.0001, 1.0001, 0.4482]) * RHO_SEA_LEVEL_ENGLISH,
+            np.array([1.0001, 1.0001, 0.4482]) * 0.0023769,
```
ehariton: could use a note here saying 0.002 is RHO Sea Level Density in slugs / etc

aviary/variable_info/variable_meta_data.py
```diff
@@ -6988,6 +6988,18 @@
     default_value=0,
 )
 
+add_meta_data(
```
ehariton, line 6991: Do you need to set Mission.SEA_LEVEL_DENSITY in preprocessor.py if the user hasn't called it out in the CSV? I know with Mission.GRAVITY we did have to do this.

ehariton (approved):

aviary/variable_info/variable_meta_data.py
Kenneth-T-Moore: Discussed with Eliot, but yeah, this is there already.

PR #1235, opened 2026-07-15
Gravity v4 

### LINKED ISSUE #1192: Update values of constants

ehariton:
### Summary

All gravity constants are now uniform and encapsulated in constants.py . The differences between FLOPS and GASP gravities have been eliminated.

Follow on to #1219, #1227, #1232.
Removes GRAV_ENGLISH_GASP and replaces with Mission.GRAVITY. 

One reference to GNS as gravity in a standard python function was not replaced because this function seems to be trying to account for the rotation rate of the earth as well, so we just left it alone and put some warnings around it indicating it won't work for non-earth planets.

### Related Issues

Resolves #1192 

None

### AI Usage

None.


aviary/mission/two_dof/ode/test/test_flight_path_ode.py
```diff
@@ -63,7 +65,7 @@ def test_case1(self):
             'fuselage_pitch': [0.0, 0.0],
             'load_factor': [0.25086942, 0.25086942],
             Dynamic.Mission.ALTITUDE_RATE: [0.0, 0.0],
-            Dynamic.Mission.ALTITUDE_RATE_MAX: [-0.01815763, -0.01815763],
+            Dynamic.Mission.ALTITUDE_RATE_MAX: [-0.018143, -0.018143],
```
cmbenne3, line 68: Why do these values change?

aviary/subsystems/aerodynamics/gasp_based/interference.py
```diff
@@ -378,11 +379,12 @@ def setup_partials(self):
         )
 
     def compute(self, inputs, outputs):
+        grav_english = self.options[Mission.GRAVITY][0]
         CKW, CBARW, EM, T0, XKV, AREASHIELDWF, FEINTWF = inputs.values()
 
         # from gaspmain.f
         # reli = reynolds number per foot
-        RELI = np.sqrt(1.4 * GRAV_ENGLISH_GASP * 53.32) * EM * np.sqrt(T0) / XKV  # dynamic
+        RELI = np.sqrt(1.4 * grav_english * 53.32) * EM * np.sqrt(T0) / XKV  # dynamic
```
cmbenne3, line 403: This might be outside of the scope of this PR, but I think this equation looks like it is assuming air, and calculating the speed of sound and multiplying by mach number to obtain the V in reynolds number.
1.4: The ratio of specific heats for air.
53.32: The specific gas constant for air in English units.
Perhaps these values should be obtained from the atmopshere component so that it correctly calculates for gasses other than air.

aviary/subsystems/propulsion/utils.py
```diff
@@ -98,15 +99,19 @@ def convert_geopotential_altitude(altitude):
...
+            'convert_geopotential_altitude() only functions properly when constants.GRAV_EARTH is specified in m/s**2'
+        )
+    radius_earth = constants.RADIUS_EARTH[0]
     if constants.RADIUS_EARTH[1] != 'm':
         warnings.warn(
             'convert_geopotential_altitude() only functions properly when constants.RADIUS_EARTH is specified in meters!'
         )
+
     CM1 = 0.99850  # Center of mass (Earth)? Unknown
     OC2 = 26.76566e-10  # Unknown
```
cmbenne3:
```suggestion
CM1 = 0.99850  # Oblateness/Gravity exponent (accounts for earth not being perfect sphere)
OC2 = 26.76566e-10  # Accounts for centrifugal acceleration due to Earth's rotation
```
From conversation with ChatGSFC:
Constants Derivation for Geopotential to Geometric Altitude Conversion

CM1 (Oblateness/Gravity Gradient Term): 0.99850
- Represents the modified inverse-square law exponent for Earth's gravity.
- A perfect point-mass sphere yields an exponent of 2.0.
- Accounting for Earth's J2 harmonic (equatorial bulge/oblateness), the effective gravity
  gradient exponent at standard latitude drops to ~1.9985.
- Since the equation uses (R_earth / R)**(CM1 + 1.0), CM1 is defined as 1.9985 - 1.0 = 0.9985.

OC2 (Centrifugal Relief Term): 26.76566e-10 (approx. 2.677e-9)
- Represents the centrifugal acceleration due to Earth's rotation, which counteracts gravity.
- Derived from Earth's angular velocity (ω ≈ 7.2921e-5 rad/s) at the standard atmospheric 
  reference latitude (φ = 45° 32' 33").
- It is proportional to the centrifugal potential: ω² * cos²(φ).
  (7.2921e-5)² * cos²(45.5425°) ≈ 2.676e-9 .

aviary/subsystems/propulsion/utils.py
```diff
@@ -98,15 +99,19 @@ def convert_geopotential_altitude(altitude):
...
-    GNS = 9.8236930  # grav_accel_at_surface_earth?
```
cmbenne3, line 109: Since this function used a different constant before (we have now updated it to use the single gravity constant = 9.80665) then this function will evaluate differently. I think we are now accounting for the centrifugal relief twice. GNS is the old "newtonian gravity", which would be the gravity we feel if the earth isn't spinning. Our new grav_metric value already includes the effect of spinning and so we are now double counting it.

From discussion with ChatGSFC:

Why the Old Algorithm Works
In the old algorithm:
GN = GNS * (radius_earth / R) ** (CM1 + 1.0)
H = (R * GN * ... - Z * (R - Z / 2.0) * OC2) / g

It starts with GNS (9.8236930), which is the pure Newtonian gravity at the Earth's surface, ignoring rotation.
It decays that pure gravity with altitude to find the local pure gravity (GN).
It calculates the gravitational potential energy.
It subtracts the centrifugal potential energy explicitly with the OC2 term.
Finally, it divides by g (which is standard effective gravity, 9.80665) to convert the total potential energy into Geopotential Altitude ($H$).

Note: This exactly mirrors the official equations from the U.S. Standard Atmosphere 1976.

Why the New Algorithm is Bugged
In the new algorithm:
GN = grav_metric * (radius_earth / R) ** (CM1 + 1.0)
H = (R * GN * ... - Z * (R - Z / 2.0) * OC2) / grav_metric

It starts with grav_metric (9.80665). This number already has the surface centrifugal force subtracted from it.
It calculates the potential energy based on this artificially lowered gravity.
Then, it subtracts the OC2 centrifugal term AGAIN.

By doing this, the new algorithm penalizes the aircraft for Earth's rotation twice. As you go higher in altitude, this error will compound, resulting in an incorrect Geopotential Altitude.

cmbenne3 (changes requested):

aviary/subsystems/propulsion/utils.py
cmbenne3: Maybe we should add a test for this utility and compare to a data table from somewhere to make sure this analysis is correct?

aviary/subsystems/propulsion/utils.py
```diff
@@ -117,8 +122,10 @@ def convert_geopotential_altitude(altitude):
 
         while abs(DH) > 1.0:
             R = radius_earth + Z
-            GN = GNS * (radius_earth / R) ** (CM1 + 1.0)
-            H = (R * GN * ((R / radius_earth) ** CM1 - 1.0) / CM1 - Z * (R - Z / 2.0) * OC2) / g
+            GN = grav_metric * (radius_earth / R) ** (CM1 + 1.0)
```
cmbenne3:
```suggestion
            GN = (grav_metric + (OC2 * radius_earth)) * (radius_earth / R) ** (CM1 + 1.0)
```

I believe this would correct the calculation, we might prefer to calculate the constant outside the loop for very marginal performance gains:
GNS = grav_metric + (OC2 * radius_earth)

aviary/subsystems/aerodynamics/gasp_based/interference.py
ehariton: I agree that this whole component will fail for non-earth air cases. The extent of what I've done in other scenarios is put a warning.

aviary/subsystems/propulsion/utils.py
ehariton: Resolution. I'm not going to change this at all and I added some warnings to the file so folks know it won't be applicable to non-earth planets.

aviary/subsystems/aerodynamics/gasp_based/interference.py
ehariton: I added a warning here:
warnings.warn(
                'WingFuselageInterferenceMission() is not valid for planets other than earth.'
            )

aviary/subsystems/propulsion/utils.py
ehariton: I added in your comments.

aviary/mission/two_dof/ode/test/test_flight_path_ode.py
ehariton: Because we switched from from using GRAV_ENGLISH_GASP (32.2  # ft/s^2) to GRAV_EARTH = (9.80665, 'm/s**2',) which converts to 𝟑𝟐.𝟏𝟕𝟒𝟎𝟓 ft/s^2. The differences in the decimal place drive the difference you're seeing there.

aviary/subsystems/aerodynamics/gasp_based/interference.py
jkirk5: Since this component is unused I am fine with doing this rather than reworking it. Can you add a quick comment after the warning explaining why (the hardcoded constants assuming Earth-atmosphere)?

aviary/subsystems/propulsion/utils.py
```diff
@@ -98,15 +98,17 @@ def convert_geopotential_altitude(altitude):
     except TypeError:
         altitude = [altitude]
 
-    g = constants.GRAV_EARTH
+    g = constants.GRAV_EARTH1[0]
```
jkirk5: What is `GRAV_EARTH1`? Is this just a typo?

aviary/subsystems/propulsion/utils.py
```diff
@@ -98,15 +98,17 @@ def convert_geopotential_altitude(altitude):
...
     radius_earth = constants.RADIUS_EARTH[0]  # meters
-    if constants.RADIUS_EARTH[1] != 'm':
+    if constants.RADIUS_EARTH[1] != 'm' or constants.GRAV_EARTH1[1] != 'm/s**2':
```
jkirk5: We can just convert units if they aren't what we expected
```
from openmdao.utils.units import convert_units

convert_units(val, old_units, new_units)
```

I'm not sure why we need to check that our hardcoded constants have their expected units to begin with though

aviary/constants.py
```diff
@@ -5,31 +5,40 @@
...
+    6051800.0,
     'm',
-)  # Venus Global Reference Atmospheric Model (Venus-GRAM): User Guide, NASA/TM-20210022168, avg of equatorial and polar radius
+)
+# Venus Global Reference Atmospheric Model (Venus-GRAM): User Guide, NASA/TM-20210022168, avg of equatorial and polar radius
 
-# GNS = 9.8236930  # grav_accel_at_surface_earth # TODO: Remove this from other parts of Aviary
-GRAV_METRIC_GASP = 9.81  # m/s^2
-GRAV_ENGLISH_GASP = 32.2  # ft/s^2
 GRAV_ENGLISH_LBM = 1.0  # lbf/lbm
```
jkirk5: This constant will eventually also need to go for Aviary to fully support different gravity, as the ratio of lbm/lbf = 1 only applies at earth sea level

aviary/constants.py
```diff
@@ -5,31 +5,40 @@
 GRAV_EARTH = (
     9.80665,
     'm/s**2',
-)  # NIST https://physics.nist.gov/cgi-bin/cuu/Value?gn|search_for=gravity
+)
+# NIST https://physics.nist.gov/cgi-bin/cuu/Value?gn|search_for=gravity
+# convert_geopotential_altitude() is a python utility function for engine decs that require GRAV_EARTH to be specified in m/s**2!
```
jkirk5: I don't think this comment is necessary because these are hardcoded values that won't ever change - the constants are such fundamental values that a user should fully expect to break things if they change anything here (plus we can easily add handling for unexpected units in the util)

aviary/subsystems/propulsion/utils.py
ehariton: That IS a typo!

aviary/mission/two_dof/ode/test/test_flight_path_ode.py
cmbenne3: But we do an aviary_options.set_val for this test which should overwrite the value to 32.2, which is the same as what this test was using before?
This suggests the override isn't working?

ehariton: I see that we are using the set-val, but at the same time this particular output is extremely sensitive, if I change the set val from 32.2 to 32.20001 that is enough to cause these sorts of changes in `ALTITUDE_RATE_MAX`

aviary/subsystems/aerodynamics/gasp_based/interference.py
ehariton: done.

aviary/constants.py
ehariton: I removed the second comment.

aviary/constants.py
ehariton: I added a note for this.

aviary/subsystems/propulsion/utils.py
ehariton: I added the convert_units function as requested. the equations inside of this helper function require particular units on input and can't convert on their own.

aviary/subsystems/propulsion/utils.py
ehariton: I'm hesitant to put this correction in because I'm not sure how these specific equations work. Additionally, it is a tiny correction factor.

cmbenne3: yeah let's leave as is - I think we decided not to touch this!

aviary/mission/two_dof/ode/test/test_flight_path_ode.py
cmbenne3: ok - that's interesting that it's that sensitive since we don't have an optimizer in here it's just running inputs through the ODE? If we're happy that the sensitivity isn't indicative of any errors then we should be good to go!

cmbenne3 (approved):

jkirk5 (approved):

Issue #598, opened 2024-11-04
Convert Options to Inputs


ehariton:
### Desired capability or behavior.

There are a number of options that users will want to manipulate during the optimization. 

For example: `Aircraft.Wing.LOAD_PATH_SWEEP_DIST` are actually functions of `Aircraft.Wing.SWEEP` and should change as sweep changes. However, `Aircraft.Wing.LOAD_PATH_SWEEP_DIST` is an option and thus is not allowed to change during the optimization. This value contributes as a secondary effect to wing weight calculations.

ehariton:
> motor_mission and motor_premission add the whole of aviary_values as an option to the mdao component. I don't think it gets used in motor_mission so perhaps we can remove it...

It looks like we are using aviary_values for `Aircraft.Engine.RPM_DESIGN` in `motor_premission.py`. But I agree it doesn't look like we use aviary_values at all in `motor_mission.py`

cmbenne3:
> > motor_mission and motor_premission add the whole of aviary_values as an option to the mdao component. I don't think it gets used in motor_mission so perhaps we can remove it...
> 
> It looks like we are using aviary_values for `Aircraft.Engine.RPM_DESIGN` in `motor_premission.py`. But I agree it doesn't look like we use aviary_values at all in `motor_mission.py`

To loop back round on this Issue - @ehariton do you think I should just add the single 'Aircraft.Engine.RPM_DESIGN' as an option to motor_premission.py, remove aviary_values from motor_mission.py and motor_premission.py and then close this issue entirely? @jkirk5 what do you think?

xjjiang:
Summary of options of float type

FLOPS based geometry:

> ~~characteristic_lengths.py~~
>>    ~~NacelleCharacteristicLength~~
>>>  &nbsp; &nbsp; &nbsp;    ~~Aircraft.Engine.REFERENCE_SLS_THRUST, lbf~~

>  ~~fuselage.py~~
>>    ~~DetailedCabinLayout~~
>>>  &nbsp; &nbsp; &nbsp;     ~~Aircraft.CrewPayload.Design.SEAT_PITCH_FIRST, inch~~
>>>  &nbsp; &nbsp; &nbsp;     ~~Aircraft.CrewPayload.Design.SEAT_PITCH_ECONOMY, inch~~

>>    ~~BWBDetailedCabinLayout~~
[...]

PR #1240, opened 2026-07-17
Add a unit test for non-zero Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER

xjjiang:
### Summary

This PR added a unit test for non-zero `Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER`.

### Related Issues

- Resolves #

None

### AI Usage

None.


Kenneth-T-Moore:
I wonder if this is a case where we might want more granular control over which mass discontinuities to smooth during a run. When smoothing here, looks like the pre-mission test has an operating mass that is 75 lbm different from the original non-smoothed result.

aviary/subsystems/mass/gasp_based/test/test_mass_premission.py
```diff
@@ -2554,14 +2554,14 @@ def test_case1(self):
 
         # fuel values:
         assert_near_equal(
-            self.prob['OEM_wingfuel_mass'], 59367.49187498, tol
+            self.prob['OEM_wingfuel_mass'], 59444.34537075, tol
```
cmbenne3: Why do these values update? Is the difference completely due to SMOOTH_MASS_DISCONTINUTIES? Seems like large differences given how close your linear approximation is to the step function?

I guess the maximum difference in cargo_handling_weight is about 82.5lbs so maybe that's enough to cause these differences, can we test this to make sure there's not something else going on?

I think in the test case the values are as follows:

if smooth:
    cargo_handling_wt = (PAX * uld_per_pax + 0.5) * unit_weight_cargo_handling
    = (154 * 0 + 0.5 ) * 165 = 82.5
else:
    uld_per_pax = uld_per_pax.real
    cargo_handling_wt = (int(PAX * uld_per_pax) + 1) * unit_weight_cargo_handling
    = (int(154*0)+1)*165 = 165

I think GASP is basically saying a ULD is 165lbs and you have to round up to a whole number of them.

cmbenne3 (changes requested):
Please can we check nothing else is changing and all the test value difference are driven by the addition of the smoothed version of cargo_container_mass calculation. (Smooth mass discontinuities was set = True for some of these tests).

cmbenne3:
> I wonder if this is a case where we might want more granular control over which mass discontinuities to smooth during a run. When smoothing here, looks like the pre-mission test has an operating mass that is 75 lbm different from the original non-smoothed result.

Perhaps we should copy the FLOPS cargo container component which uses a sin_int4() function to approximate the int() whilst being differentiable:
image

Both these components essentially do the same thing in different ways.
GASP asks the user to specify number of containers per passenger (uld_per_pax), rounds up to next whole integer and assumes each container is 165lbs.
FLOPS adds up the amount of cargo and loads up 950lbs into each container, rounds up the number of containers and assumes each container is 175lbs.

Here are the integer smoothing functions we have in utils:
image

Looks like we also have a tanh smoother as well as a sin4 smoother. Not sure if one is better than the other - do we need both and should we consolidate all instances of using these functions to one or the other?

aviary/subsystems/mass/gasp_based/test/test_mass_premission.py
xjjiang: The Fortran code in GASP is `FLOAT((IFIX(PAX*CW(14))+1))*165`. I am not sure why Jeff does it this way. I am trying to ask Jeff. I may smooth this function is that is the right way.

xjjiang:
> I wonder if this is a case where we might want more granular control over which mass discontinuities to smooth during a run. When smoothing here, looks like the pre-mission test has an operating mass that is 75 lbm different from the original non-smoothed result.

I am trying to ask Jeff about it.

Kenneth-T-Moore:
This might be a question for @cmbenne3 , but is there a compelling case for ULD_MASS_PER_PASSENGER  to be turned into a variable?

cmbenne3:
> This might be a question for @cmbenne3 , but is there a compelling case for ULD_MASS_PER_PASSENGER to be turned into a variable?

If definition of variable is 'something a user wants to give to the optimizer as a design variable' then I think the answer is 'no', Jason had a different view that we discussed on in the #598 issue that everything that could be a variable (non integers, booleans) should be a variable, and Aviary shouldn't guess what a user might or might not want to do. I can get on board with this logic and so I'm happy to make this change.

cmbenne3:
> > I wonder if this is a case where we might want more granular control over which mass discontinuities to smooth during a run. When smoothing here, looks like the pre-mission test has an operating mass that is 75 lbm different from the original non-smoothed result.
> 
> I am trying to ask Jeff about it.

I don't think we should add additional variables to control 'smoothing' of this. I think i would prefer to approximate the int() function with a differentiable version of it, like we do for FLOPS and accept we will have small differences to GASP with the benefit that we can use a gradient based optimizer.

xjjiang:
@cmbenne3 @Kenneth-T-Moore The step function in this case is very special case. Here the equation is:

`cargo_handling_wt = (int(PAX * uld_per_pax) + 1) * unit_weight_cargo_handling`

For PAX = 180, if `Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER` has a change of 1/180, `cargo_handling_wt` has a jump of 165 lbm. It is too sensitive to variation. Therefore, I don't think we should smooth the `int()` function. That was why I tried to use a linear function. Jeff think the `int()` function is necessary to model ULD. So, now I think we should not give our users `Aircraft.Design.SMOOTH_MASS_DISCONTINUITIES` as an option. I will remove this part.

xjjiang:
@cmbenne3 @Kenneth-T-Moore The next question is: should we change `Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER` from option to variable? According to Jeff, this parameter is computed "based on data for several aircraft/airlines". And he said: "Some A/C don’t use ULD’s, like the Boeing Max-8". Will users perturb this parameter for optimization? If not for sure, I will close this PR.

cmbenne3:
> @cmbenne3 @Kenneth-T-Moore The step function in this case is very special case. Here the equation is:
> 
> `cargo_handling_wt = (int(PAX * uld_per_pax) + 1) * unit_weight_cargo_handling`
> 
> For PAX = 180, if `Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER` has a change of 1/180, `cargo_handling_wt` has a jump of 165 lbm. It is too sensitive to variation. Therefore, I don't think we should smooth the `int()` function. That was why I tried to use a linear function. Jeff think the `int()` function is necessary to model ULD. So, now I think we should not give our users `Aircraft.Design.SMOOTH_MASS_DISCONTINUITIES` as an option. I will remove this part.

This is all behaving like i would expect - the ULD is a big metal box weighing 165 lbs and the function is trying to work out how many boxes you're loading onto the aircraft. You can't have a fraction of the box in real life hence the int(). My preference would be to copy what we have done on the flops side, approximate the int with a smooth function and accept that on the edges of the int() we'll calculate a fractional container that wouldn't be there in GASP or in real life. This allows us to differentiate the int(). If we don't need to differentiate it then leaving as is and probably converting the flops component back is the better thing to do.

cmbenne3:
> @cmbenne3 @Kenneth-T-Moore The next question is: should we change `Aircraft.CrewPayload.ULD_MASS_PER_PASSENGER` from option to variable? According to Jeff, this parameter is computed "based on data for several aircraft/airlines". And he said: "Some A/C don’t use ULD’s, like the Boeing Max-8". Will users perturb this parameter for optimization? If not for sure, I will close this PR.

This is a question for @jkirk5 .
In a very complex model it's possible that the size and shape of the ULD containers would change if they were being redesigned for a novel aircraft. This could then be used to estimate the ULD_MASS_PER_PASSENGER, and aviary's core calculation used to sum them up. In this case though it would probably just be easier for the user to calculate and override cargo_container_mass directly.

xjjiang:
@cmbenne3 After discussing with Ken, I decided to switch it back to option from variable. Now, it contains some new comments and one new unit test.

Kenneth-T-Moore (approved):

cmbenne3 (approved):
ok - happy to leave this as an option if that's what we've decided.

PR #1244, opened 2026-07-21
Ia376 option usage vis

### LINKED ISSUE #376: Option Usage Visualization Tool

From @jdgratz10:
Right now, when a user selects what options to set, it is difficult for them to know if they should set the option from an old FLOPS case, an old GASP case, or both. We need to develop a tool to help them visualize where their option will be used.

Perhaps a tool similar to the N2 or piggybacking off of the N2 would be helpful here. If we could show the user the path through the subsystems which their option will take, it would help them understand what the option will affect and how they should set it.

_Note: This is closely related to ongoing discussions on option tracking inside OpenMDAO and actual implementation may differ from this initial suggestion, however the creation of a tool to add to the dashboard showing option use is still highly desired_

_No response_

### Associated Bug Report

_No response_

hschilling:
### Summary

Added tab to the dashboard that shows the output of `Problem.list_options()`

Eventually there will be a more feature-rich version of this, but this at least gives users some information about the options. 

### Related Issues

- Resolves #376 

None

### AI Usage

No AI was used in the writing of this code


aviary/interface/reports.py
```diff
@@ -87,6 +87,22 @@ def register_custom_reports():
...
     )
+    register_report(
+        name='list_options',
+        func=_list_options_report,
+        desc='Generates a report on the Problem options',
+        class_name='AviaryProblem',
+        method='run_driver',
+        pre_or_post='post',
+    )
+    register_report(
```
Kenneth-T-Moore, line 98: Is this the same report twice?

Ah, run_driver and run_model.

Kenneth-T-Moore (approved):

PR #1246, opened 2026-07-21
updates to fix engine scaling index bug

### LINKED ISSUE #1245: Index Error results in incorrectly scaled engines
### Description

There is a bug in the engine scaling logic for EngineMass subsystem.

We use np.where() inside an array already sliced by np.where() to find an index, and then apply that index to the original unsliced array, which results in incorrect scaling of engines.

Our test for the engine subsystem just happens to pass because of the order that we setup the engines in, if you swap the order of engines 1 and 2 in the test, the test fails.

### Example

I added debug print statements into EngineMass component:

Image

Swap line 93 in test_engine.py from:
preprocess_propulsion(options, [engine, engine2, engine3])
to
preprocess_propulsion(options, [engine2, engine, engine3])
and update the expected results from:
mass_expected = np.array([5779.16494294, 6000.0, 5814.38])
total_mass_expected = np.array([35187.08988589])
additional_mass_expected = np.array([5201.24844865, 5400.0, 5232.942])
to:
mass_expected = np.array([6000.0, 5779.16494294, 5814.38])
total_mass_expected = np.array([35187.08988589])
additional_mass_expected = np.array([5400.0, 5201.24844865, 5232.942])

Here is the printout from my debug statements for the original and swapped engines from test_engine:

Original:
obtaining scale and param indices:
scale_mass = [True, False, True]
scale_idx = np.where(scale_mass) = (array([0, 2]),)
scaling_parameter = [1.15 1.15 0.2 ]
param_idx = np.where(scaling_parameter[scale_idx] >= 0.3) = (array([0]),)
initialized calc_mass = [6000. 6000. 6000.]
after linear scaling calc_mass = [4932.685 6000.   5814.38 ]
after power law scaling final calc_mass = [5779.16494294 6000.  5814.38]
You can see that this matches the mass_expected list

Swapped engines 1 and 2:
obtaining scale and param indices:
scale_mass = [False, True, True]
scale_idx = np.where(scale_mass) = (array([1, 2]),)
scaling_parameter = [1.15 1.15 0.2 ]
param_idx = np.where(scaling_parameter[scale_idx] >= 0.3) = (array([0]),)
initialized calc_mass = [6000. 6000. 6000.]
after linear scaling calc_mass = [6000.    4932.685 5814.38 ]
after power law scaling final calc_mass = [5779.16494294 4932.685  5814.38]

All engines have been scaled here even though scale_mass was false for the first engine due to the bug where scale index = array([1, 2])
and then param_idx returns array([0]) when it should return array([1]) when applied to the calc_mass list.

### Error Message

There is no error - Aviary incorrectly scales engines with no warnings to the user

### Aviary Version

1.0.2-dev

### Environment Information

_No response_

cmbenne3:
### Summary

This PR corrects the bug with engine scaling indexing and adds an extra test to check that it works correctly (it matches test_case_2 with the engines simply listed in a different order - this new test used to fail with the old code)

### Related Issues

- Resolves #1245 

None

### AI Usage

### DIFF
--- a/aviary/subsystems/mass/flops_based/engine.py
+++ b/aviary/subsystems/mass/flops_based/engine.py
@@ -43,26 +43,38 @@ class EngineMass(om.ExplicitComponent):
         scaled_sls_thrust = np.array(inputs[Aircraft.Engine.SCALED_SLS_THRUST])
         scaling_parameter = np.array(inputs[Aircraft.Engine.MASS_SCALER])
 
+        print('obtaining scale and param indices:')
         scale_idx = np.where(scale_mass)
+        print(f'scale_mass = {scale_mass}')
+        print(f'scale_idx = np.where(scale_mass) = {scale_idx}')
         # indices where scaling is applied and scaling equation is used
         param_idx = np.where(scaling_parameter[scale_idx] >= 0.3)
-
+        print(f'scaling_parameter = {scaling_parameter}')
+        print(f'param_idx = np.where(scaling_parameter[scale_idx] >= 0.3) = {param_idx}')
+        is_scaled = scale_mass
+        is_power = is_scaled & (scaling_parameter >= 0.3)
+        is_linear = is_scaled & (scaling_parameter < 0.3)
+        print(f'is_power = {is_power}')
+        print(f'is_linear = {is_linear}')
         # default mass is reference mass
         # use dtype to make complex safe
         calc_mass = np.array(ref_engine_mass, dtype=scaled_sls_thrust.dtype)
+        print(f'initialized calc_mass = {calc_mass}')
 
         # scale engine mass using equation chosen by value of user-provided mass scaler
         thrust_ratio = scaled_sls_thrust / ref_sls_thrust
 
-        calc_mass[scale_idx] = (
-            ref_engine_mass[scale_idx]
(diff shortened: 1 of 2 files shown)


aviary/subsystems/mass/flops_based/engine.py
```diff
@@ -90,34 +88,30 @@ def compute_partials(self, inputs, J):
...
 
-        scale_deriv[param_idx] = (
-            ref_engine_mass[param_idx] * thrust_ratio[param_idx] ** scaling_parameter[param_idx]
+        scale_deriv[is_power] = (
+            ref_engine_mass[is_power] * thrust_ratio[is_power] ** scaling_parameter[is_power]
         )
-        if len(param_idx) > 0:
-            scale_deriv[param_idx] = scaled_mass * np.log(thrust_ratio[param_idx])
+
+        scale_deriv[is_power] = scaled_mass * np.log(thrust_ratio[is_power])
```
Kenneth-T-Moore, line 110: This computation overwrites the one above. Maybe youy meant += or *=.

cmbenne3: good spot - I will remove it since the correct derivative is the one I was overwriting it with!

Kenneth-T-Moore (approved):

PR #1259, opened 2026-08-11
BWB Updates

Kenneth-T-Moore:
### Summary

1. Redefined `Aircraft.Wing.ROOT_CHORD` to go from leading edge to trailing edge instead of ending at the rear spar.
2. Added hierarchy variables for `Aircraft.BWB.REAR_SPAR_PERCENT_CHORD_ROOT` and `Aircraft.BWB.REAR_SPAR_PERCENT_CHORD_CENTERLINE`.
3. Cleaned up BWB detailed wing so that they don't require a placeholder value at the centerline.
4. Added support for specifying detailed wings that start at the centerline. This was supported in FLOPS through the declaration of a higher index in "NESOB" that defines the root index of the wing. The NESOB namelist input is now implemented in Aviary as `Aircraft.BWB.WING_ROOT_INDEX`.
5. Moved the CD and CL timeseries out of the phase builders and into the aero builder, since the components that compute them are added by the aero subsystem.
6. Restored a couple of derivative checks that were commented out in the BWB components. These components mix complex step with some analytic sparse declarations.
7. Fixed a small bug in the number of bays calculation that effectively rounded down instead of up.
8. Fixed some of the centerline/root replacement values for the NESOB=0 case.

### Related Issues

- Resolves #

None

### AI Usage


aviary/variable_info/variable_meta_data.py
```diff
@@ -5958,7 +5989,7 @@
     meta_data=_MetaData,
     historical_name={
         'GASP': 'INGASP.CROOTW',
-        'FLOPS': 'WTIN.XLW',
+        'FLOPS': 'WTIN.XLOUT',
```
jkirk5: XLOUT isn't a FLOPS input, but instead an intermediate calculated value (not part of WTIN namelist). Put "XLOUT" as a comment here for now

aviary/subsystems/geometry/flops_based/fuselage.py
```diff
@@ -100,20 +91,14 @@ def compute(self, inputs, outputs):
...
             )
-        if rear_spar_percent_chord <= 0.0:
-            raise ValueError(
-                'Rear_spar_percent_chord must be positive, '
-                f'however {rear_spar_percent_chord} is provided.'
-            )
 
         ref_diameter = 0.5 * (max_height + max_width)
-        planform_area = max_width * (length + root_chord / rear_spar_percent_chord) / 2.0
+        planform_area = max_width * (length + root_chord) / 2.0
```
jkirk5, line 101: This change means that Aviary doesn't match the FLOPS equations directly anymore, and messes up converting FLOPS files. `root_chord`, previously XLW (and a FLOPS input) has been redefined to match XLOUT (an intermediate FLOPS calculation) which makes replicating a FLOPS case require hand-calculating this new value and adding it to the CSV. Why?

aviary/subsystems/geometry/flops_based/fuselage.py
```diff
@@ -67,14 +67,6 @@ def setup(self):
         add_aviary_input(self, Aircraft.Fuselage.MAX_WIDTH, units='ft')
         add_aviary_input(self, Aircraft.Fuselage.MAX_HEIGHT, units='ft')
         add_aviary_input(self, Aircraft.Wing.ROOT_CHORD, units='ft')
-        self.add_input(
```
jkirk5, line 70: Why add a variable for RSPSOB to the hierarchy if we are simplifying the equations to no longer use it?

jkirk5 (changes requested):
This change diverges from FLOPS in a pretty significant way by redefining what wing_root refers to. It's not clear to me why this was necessary?

Kenneth-T-Moore:
> This change diverges from FLOPS in a pretty significant way by redefining what wing_root refers to. It's not clear to me why this was necessary?

This was motivated by a confusion about Aircraft.Wing.ROOT_CHORD as defined for the BWB, which was not the root chord because it stopped at the rear spar instead of the trailing edge, meaning it is not the chord at that location.

I just took another look at these variables in the Fortran code, and I think the correct way to address this is to add WTIN.XLW as a new separate variable in the hierarchy. This is defined in FLOPS as the "Side Wall Length".  It is computed by both simple and detailed cabin design. I should restore the output there to XLW, and then compute root chord in a separate simple component (XLOUT, not a FLOPS input). This will restore the ability to specify XLW (which would become `Aircraft.Fuselage.SIDEWALLLENGTH`) in the csv file.

aviary/variable_info/variable_meta_data.py
```diff
@@ -3222,6 +3222,34 @@
...
+    Aircraft.Fuselage.CABIN_SIDEWALL_LENGTH,
+    meta_data=_MetaData,
+    historical_name={
+        'GASP': None,
+        'FLOPS': 'WTIN.XLW',
+    },
+    units='ft',
+    desc='Outboard wall of the passenger cabin in a BWB.',
+    default_value=0.0,
+)
```
jkirk5: It doesn't look like this variable gets used anywhere? Having a variable in the hierarchy that does nothing is worse than not having it at all in my opinion

Kenneth-T-Moore: > It doesn't look like this variable gets used anywhere? Having a variable in the hierarchy that does nothing is worse than not having it at all in my opinion

Done. I forgot to remove that when we decided to postpone supporting this way f specifying the model.

jkirk5 (approved):

PR #1262, opened 2026-08-12
Add a minimum FLOPS aircraft example

### LINKED ISSUE #1141: Creation of a "minimum" aircraft case for examples
### Missing documentation or idea for new documentation.

Our examples currently borrow one of the existing aircraft models, which are based on FLOPS/GASP models that themselves are calibrated to match real-world aircraft or technology forecasts. This isn't required and is overly complicated for new users. We should create an aircraft that only has the bare minimum number of inputs needed to run a "realistic" aircraft example and no calibration factors or any similar variables that just adjust the calculation.

### Issue with existing documentation.

This is present in all examples in documentation, as well as snippets of aircraft input files shown in https://openmdao.github.io/Aviary/user_guide/fundamentals/aircraft_definition.html

suhaslord:
## Summary

Adds an intentionally generic single-aisle aircraft input deck that uses FLOPS mass and aerodynamics methods with the energy-state equations of motion. The example contains the fundamental payload, propulsion, geometry, and mission inputs without the calibration factors and derived overrides in the advanced single-aisle reference deck.

Updates the aircraft-definition documentation to use the minimum deck as its beginner-facing example and adds a focused end-to-end mission regression test.

Fixes #1141.

## Reduction method

The existing `advanced_single_aisle_FLOPS.csv` deck was reduced in grouped and isolated SLSQP experiments. Each candidate had to load, preprocess, optimize, satisfy the target range, and retain finite, positive aircraft masses, geometry, lift, drag, thrust, and aerodynamic coefficients.

The base deck was measured at 194 converted input rows plus 282 unconverted legacy rows. The new deck has 42 input rows:

- removed calibration mass and drag scalers;
- removed unconverted legacy rows and derived/output overrides;
- used safe Aviary defaults for standard options and conventional technology assumptions;
- used the simple wing-mass method instead of detailed-wing distributions;
- retained inputs whose zero defaults caused invalid physics, including wing fuel fraction and hydraulic pressure;
- retained total engine count for topology preprocessing and wing location because omitting location materially changed landing-gear mass.

## Validation

The repository's canonical energy-state/SLSQP configuration converged for the untouched aircraft with 13,234.4 lbm mission fuel. Changing only takeoff fuel did not control convergence: the default energy-state phase definition converged both with zero and deck takeoff fuel, while `advanced_single_aisle/phase_info.py` failed on SLSQP iteration 1 in both cases. This isolates the earlier singularity to the advanced trajectory configuration rather than the aircraft deck or takeoff-fuel value.

The minimum aircraft converged with SLSQP in 12 iterations for a 1,906 nmi mission. Representative outputs are:

- design gross mass: 135,723.3 lbm;
- operating mass: 86,251.1 lbm;
- mission fuel: 18,672.2 lbm;
- wing mass: 14,245.4 lbm;
- fuselage mass: 16,790.1 lbm;
- wing area/span/aspect ratio: 1,220 ft² / 118.75 ft / 11.56;
- climb lift coefficient range: 0.50–1.88;
- climb drag coefficient range: 0.030–0.208.

## Tests

- `testflo . --nompi`: 1,697 passed, 0 failed, 46 skipped
- focused mission/input/preprocess tests: 9 passed
- FLOPS mass and aerodynamics tests: 564 passed
- `aviary check`: passed with SLSQP
- `pre-commit run --all-files`: passed
- changed documentation page: executed and built successfully with Jupyter Book

## Limitations

This is a generic starting point, not a calibration to a particular aircraft or legacy study. Users seeking higher fidelity should add appropriate technology factors, detailed geometry, or component overrides for their application.


aviary/models/aircraft/minimum_single_aisle/phase_info.py
jkirk5, line 1: Since we're re-using the default phase info here we don't need an entirely new file, so this can be removed

aviary/mission/test/test_minimum_aircraft.py
jkirk5, line 1: Can you move this file to `aviary/validation_cases/benchmark_cases`?

jkirk5 (changes requested):
Great work! Thanks for the contribution - I have a few minor requests but otherwise this is good to go!

jkirk5 (approved):

PR #1268, opened 2026-08-18
Convert Aircraft.CrewPayload.Design.SEAT_PITCH_* from options to variables

xjjiang:
### Summary

The following options are converted to variables from options:

- `Aircraft.CrewPayload.Design.SEAT_PITCH_BUSINESS`
- `Aircraft.CrewPayload.Design.SEAT_PITCH_FIRST`
- `Aircraft.CrewPayload.Design.SEAT_PITCH_ECONOMY`

Moved hard-coded initialization of variables in fuselage layout to `preprocessors.py`.

This is part of issue #598.

### Related Issues

- Resolves #

None

### AI Usage

None


aviary/subsystems/geometry/flops_based/fuselage.py
```diff
@@ -214,48 +214,28 @@ def compute(self, inputs, outputs):
...
         if num_economy_class_pax > 200:
             if num_seat_abreast_economy <= 0:
-                num_seat_abreast_economy = 8
-                if verbosity > Verbosity.BRIEF:
-                    print(
-                        'Set Aircraft.CrewPayload.Design.NUM_SEATS_ABREAST_ECONOMY ='
-                        f'{num_seat_abreast_economy}'
-                    )
+                # num_seat_abreast_economy = 8
+                raise UserWarning(
```
jkirk5: _(See other comment on fuselage.py)_ Once the preprocessor sets defaults then we no longer need these warnings

aviary/subsystems/geometry/flops_based/fuselage.py
```diff
@@ -214,48 +214,28 @@ def compute(self, inputs, outputs):
         # The 200 was derived from B757 - the largest single aisle western desig.
         if num_economy_class_pax > 200:
             if num_seat_abreast_economy <= 0:
-                num_seat_abreast_economy = 8
```
jkirk5, line 217: For the removed hardcoded values, Aviary should still be able to help users set the "correct" values that will replicate FLOPS/GASP results without extra work on their part (manually adding the right values to their csv).

In the preprocessors, we already do this for several other variables, can you add a section setting all these variables (seat pitch and number abreast) to the correct default based on legacy code if not provided by the user?

aviary/subsystems/geometry/flops_based/fuselage.py
```diff
@@ -401,7 +376,7 @@ def compute(self, inputs, outputs):
 
         # Calculate the passenger compartment length
 
-        num_engines = self.options[Aircraft.Engine.NUM_ENGINES][0]
+        num_engines = self.options[Aircraft.Propulsion.TOTAL_NUM_ENGINES]
         eng_flag = num_engines - 2 * int(num_engines / 2)  # a center mounted engine if 1.
```
jkirk5: If there are two sets of engines that both have odd numbers then this flag will not capture that there are center-mounted engines. The correct workaround may be to check each engine set and if any of them have odd number of engines then set the flag to 1. This can probably be done in a single line if we get clever with list comprehension but I'd have to think about it for a bit to figure out how to do that.

jkirk5 (changes requested):

aviary/subsystems/geometry/flops_based/fuselage.py
xjjiang: > can you add a section setting all these variables (seat pitch and number abreast) to the correct default based on legacy code if not provided by the user?

I've added a function `preprocess_fuselage_layout()` in `preprocessors.py` for this purpose. Do I misunderstand what you are asking?

aviary/subsystems/geometry/flops_based/fuselage.py
xjjiang: If there are two separate central engines, can we assume they are both aligned along the same line relative to the fuselage? GASP assumes they are. It is used to add 25 feet to fuselage length. Otherwise, we have to add 2*25 feet to fuselage length.

jkirk5: We should probably only apply the extension once, no matter how many centerline engines are assumed to be on the fuselage

xjjiang: This is what I did for now.

aviary/utils/preprocessors.py
```diff
@@ -101,13 +101,58 @@ def preprocess_options(
...
+    mass_method = aviary_options.get_val(Settings.MASS_METHOD)
+    design_type = aviary_options.get_val(Aircraft.Design.TYPE)
+
+    # check aircraft type
     if Aircraft.Fuselage.SIMPLE_LAYOUT not in aviary_options:
         if mass_method == LegacyCode.FLOPS:
             aviary_options.set_val(Aircraft.Fuselage.SIMPLE_LAYOUT, True, 'unitless')
             simple_layout = True
+            if verbosity >= Verbosity.BRIEF:
+                warnings.warn('Aircraft.Fuselage.SIMPLE_LAYOUT is not set. Assuming True.')
```
jkirk5: The default is already True so we don't need to set it here again

aviary/utils/preprocessors.py
```diff
@@ -120,28 +165,270 @@ def preprocess_options(
...
+                num_first_class_pax = aviary_options.get_val(
+                    Aircraft.CrewPayload.Design.NUM_FIRST_CLASS, 'unitless'
+                )
+            else:
+                num_first_class_pax = 0
+                aviary_options.get_val(Aircraft.CrewPayload.Design.NUM_FIRST_CLASS, 0)
+                if verbosity >= Verbosity.BRIEF:
+                    warnings.warn(
+                        'Aircraft.CrewPayload.Design.NUM_FIRST_CLASS not set, assuming default 0.'
+                    )
```
jkirk5: We already have code that handles num_passengers in `preprocess_crewpayload` so this shouldn't be needed

aviary/subsystems/geometry/flops_based/fuselage.py
xjjiang: Done.

aviary/utils/preprocessors.py
xjjiang: removed.

aviary/utils/preprocessors.py
xjjiang: Got you. Now removed.

jkirk5 (approved):

aviary/utils/preprocessors.py
```diff
@@ -120,28 +163,234 @@ def preprocess_options(
...
                 if Aircraft.Fuselage.SEAT_WIDTH_FIRST not in aviary_options:
                     aviary_options.set_val(Aircraft.Fuselage.SEAT_WIDTH_FIRST, 25.0, 'inch')
                     if verbosity >= Verbosity.BRIEF:
                         warnings.warn(
                             'Aircraft.Fuselage.SEAT_WIDTH_FIRST not set, '
                             'assuming default 25.0 inches.'
                         )
+
+                # set number of seat abreast
+                if num_economy_class_pax > 200:
```
cmbenne3, line 228: This is never set in this function - we need to do a get val first:
num_economy_class_pax = self.options[Aircraft.CrewPayload.Design.NUM_ECONOMY_CLASS

Perhaps we should add a test to for the preprocessors to catch this sort of thing in future!

PR #1274, opened 2026-08-21
Provide a better report to help identify overrides that prevent design variables from correctly impacting the problem.

### LINKED ISSUE #722: Input files overridding variables that should be calculated.
### Description

I ran into the issue when testing out external store tank model on N3CC Concept Library Model. I replicated the model using the B737max8 Concept Library Model. 

The issues was found while running the model in AVIARY Mode. I set all of the enumerated subsystem models to Aviary mode or off.  I added wing area as a design variable. The aviary model would then tend to maximize this variable instead of moving toward the generally accepted value. I noticed in the screen output that while Wing Area was being maximized and Aspect Ratio remained fixed. Since span was also fixed, the aspect ratio should have decreased. Instead the value remained fixed. This appears to be due to the value being specified in the aviary_inputs.csv file located in the mission subfolder of the model. Removing aspect ratio from the inputs csv file allowed for the value to be calculated correctly.

This issue may crop up for other variables that are dependent on design variables, but may be overridden as constant value from the input csv file. I have attached a zip file with my test case data of inputs and outputs.

[IssueCaseData.zip](https://github.com/user-attachments/files/19592687/IssueCaseData.zip)

### Example

The issue occurs when a variable that is dependent on a design variable is also listed in an inputs csv file that is loaded via the prob.load_inputs(inputsfile.csv) method call when setting up aviary problem.  I found running that when running a basic aviary model from the Concepts Library

##### Excerpt from Modified B737max8 Framework.py file#####
#lines 390

    # Adding in wing area as a design variable

    prob.model.add_design_var(
         Aircraft.Wing.AREA,
         ref=1340.,
         lower=1200,
         upper=1600,
         units="ft**2")

    prob.add_design_variables()

#line 443 Modifying USER SET FLAGS in "____main____" 
# USER-SET FLAGS
    geom_type = GeomTypes.AVIARY
    # compute_aero_polars if False while enable_openvsp is
    # True, then the polars will be loaded from the table
    # instead of computed and the VSP geometry will be used.
    aero_type = AeroTypes.AVIARY  # Select the method/tool to calculate aerodynamics
    # "AVIARY" will use Aviary empirical methods
    # "TABULAR" will use tabular aero data generated for this concept
    # "VSPAERO" will use VSPAERO to calculate the drag polar
    propulsion_type = PropTypes.AVIARY
    # AVIARY, PYCYCLE, PYCYCLE_DECK, PYCYCLE_PRECOMPUTED, NPSS_DECK.
    # Presently, propulsion does not include electrification.
    structures_type = StructTypes.AVIARY  # Select the method/tool to calculate wing structural weight
    # "AVIARY" will use Aviary empirical methods
    # "SURROGATES" will use GT TBW wing weight surrogates
    # "SUITCASE" will use SUITCASE to estimate truss-braced wing assembly weight
    enable_cg = None  # Enable center of gravity and moments of inertia calculations
    sandc_type = [SandCTypes.NONE]
#line

#################End############################

This produces output like the following:

###### Excerpt from core_geometry.md  output ############
# Geometry: FLOPS method
## Wing
| Variable Name | Value | Units |
| :- | :- | :- |
| aircraft:wing:area | 1539.91 | ft**2 |
| aircraft:wing:span | 118.75 | ft |
**| aircraft:wing:aspect_ratio | 11.43 | - |      #<- should be  8.81**
#################################################

Removing the variable aircraft:wing:aspect_ratio from aviary_input.csv file appears to fix the issue, but the problem maybe other variables in the input list that may be dependent on the design variable (Tail volume ratios, and associated tail areas and other values).  This could be a problem with aviary users building new models off these established vehicles.

### Aviary Version

0.9.9

_No response_

Kenneth-T-Moore:
### Summary

This is in a response to an issue where the wing aspect ratio, which is calculated in Aviary's height energy geometry from the wing area and span, was overridden with a value from the csv file. When wing area is selected as a design variable, the computed aspect ratio is overwritten by the .csv file and is no longer correct, and this impacts the drag computation.

Aviary's override system is flexible, but it allows you to create a situation like described above.  To help the user understand that this might be happening, this PR adds a new section to the override report:

z

The report uses the OpenMDAO variable relevancy graph to detect when an override interferes with dataflow from the design variables to the constraints and objective. The 'depends on" section shows which design variables are being blocked, and the "impacts" section shows which responses this effects.  The "feeds" section shows the components that are using the overriden value instead of the computed value. In this example, the wing area affects the computation of the fuselage wetted area, and this has potential impacts on the mass during flight (which is important for the final objective.)

### Related Issues

- Resolves #722

None

### AI Usage


cmbenne3 (approved):
Really looking forward to having this change in I think it's going to be super valuable!

PR #1279, opened 2026-08-24
add FLOPS tail sizing from tail volume coefficients

### LINKED ISSUE #1265: Add FLOPS Tail Sizing Tail Volume Coefficients Subsystem

For FLOPS based analysis Aviary does not calculate HT or VT areas, and these must be specified by the user.
This is ok if you're just copying the design of an existing aircraft, but if redesigning an aircraft or creating a optimizing a new concept this results in incorrect (static) tail sizes.

### Suggested implementation.

We have GASP based tail volume coefficients already:
aviary/subsystems/geometry/gaspbased/empennage.py
We should check if the implementation of this is the same for GASP and FLOPS. 
If not we should add a FLOPS component.
If the implementation is the same then we should add the GASP subsystem to FLOPS geometry when setting up the Aviary problem for FLOPS based geometry.

https://developer.nasa.gov/AATT-SAI/SandC/blob/main/SandC/FLOPS_Tail_Sizing.py

vyom-aggarwal:
### Summary

FLOPS-based geometry takes horizontal and vertical tail areas (`WTIN.SHT` and `WTIN.SVT`)
as user inputs, so the tails hold a fixed size while the rest of the aircraft is resized.
This adds an optional FLOPS empennage group that derives either tail area from its tail
volume coefficient, so the tails scale with the wing.

**Answering the question the issue raises first ("check if the implementation of this is
the same for GASP and FLOPS"):** the sizing relation itself is the same — it is the
standard definition, and GASP's `TailSize` implements exactly it. But the GASP subsystem
cannot be dropped into FLOPS geometry as-is, for a concrete reason:

- GASP's `TailSize` reads `Aircraft.Wing.AVERAGE_CHORD`. That variable is output only by
  `aviary/subsystems/geometry/gasp_based/wing.py`, is not computed anywhere in
  `flops_based`, and has no FLOPS historical name. In a FLOPS model it would sit at its
  `0.0` default and silently size the horizontal tail to zero.
- GASP's `TailVolCoef` (the empirical estimator for the coefficient itself) uses
  GASP-specific constants and cabin width.
- GASP parameterizes the moment arm through `MOMENT_RATIO`, a GASP idiom
  (`INGASP.COELTH` / `INGASP.BOELTV`).

So this PR takes the issue's second branch and adds a FLOPS component.

#### What was added

- `aviary/subsystems/geometry/flops_based/empennage.py` — `TailSize` (orientation
  `horizontal` or `vertical`) and an `EmpennageSize` group, mirroring the GASP module
  layout.
- Two options, both defaulting to `False`: `Aircraft.Design.COMPUTE_HTAIL_AREA` and
  `Aircraft.Design.COMPUTE_VTAIL_AREA`.
- The group is added to `PrepGeom` ahead of `wetted_area`, which consumes the tail areas.
- Documentation in the core geometry theory guide.

#### On the reference chord

A horizontal tail volume coefficient is referenced to the wing **mean aerodynamic chord**.
FLOPS-based geometry does have `Aircraft.Wing.CHARACTERISTIC_LENGTH`, but that is the mean
*geometric* chord, `S/b`, used for skin friction. The two are not interchangeable — on the
reference wing used in the tests they differ by 7.7% (12.52 ft vs 11.63 ft), which would
propagate directly into tail area.

The component therefore computes the trapezoidal-wing MAC from `Aircraft.Wing.AREA`,
`Aircraft.Wing.SPAN`, and `Aircraft.Wing.TAPER_RATIO` (`CONFIN.TR`), all of which FLOPS
provides. This matches what GASP's `AVERAGE_CHORD` represents.

#### Behavior and compatibility

Both options default to `False`, so existing FLOPS models continue to take tail areas as
inputs and are bit-for-bit unaffected. Sizing a tail makes its area an output of the
geometry group; if a user also supplies that area, the existing
`AIRCRAFT_DATA_OVERRIDE` mechanism in `override_aviary_vars` applies and the user's value
takes precedence.

Both moment arms default to `0.0`, which would divide by zero. Rather than emit `inf`, the
component raises a `ValueError` naming the variable that needs to be set.

#### Testing

`aviary/subsystems/geometry/flops_based/test/test_empennage.py` adds 16 tests:

- Horizontal and vertical areas against a hand calculation written from the root chord, a
  different arrangement of the expression than the implementation uses.
- Round-trip checks that feeding the computed area back through the definition of a volume
  coefficient recovers the requested coefficient.
- Analytic partials vs complex step for both orientations, plus taper ratio = 1, which
  sits at the root of the `(tr - 1)` factor in the taper derivative.
- Both error paths.
- Option combinations on `EmpennageSize`, and on `PrepGeom` that the tail areas are inputs
  by default and become outputs only when requested.

Full suite before this change: 1702 passed / 2 failed / 46 skipped.
After: 1718 passed / 2 failed / 46 skipped. The 16 additional passes are exactly the new
tests; the 2 failures are pre-existing and unrelated (`test_bench_solved2dof` requires
pyOptSparse, which is missing the `@require_pyoptsparse` decorator — filed separately).

#### Not included

The empirical estimator for the volume coefficient itself (the FLOPS analogue of GASP's
`TailVolCoef`) is not part of this PR. GASP's version depends on GASP-specific empirical
constants, and I did not want to invent a FLOPS equivalent. Happy to add one in a
follow-up if you can point me at the intended relationship.

I was also unable to open the `developer.nasa.gov/AATT-SAI/SandC` link in the issue, so
this implements the standard volume-coefficient definition rather than that specific
script. Worth a check against it during review in case it does something different.

### Related Issues

- Resolves #1265

None. Both new options default to `False`, which preserves the current behavior of taking
tail areas as user inputs. No existing component or variable was modified; the change is
purely additive (555 insertions, 0 deletions).

### AI Usage

Claude (Opus 5), via Claude Code, was used to explore the codebase and help me understand what changes were necessary. It was also used to fix some parts of this PR.


jkirk5:
The PR doesn't fully address issue #1265 as it is missing the FLOPS regressions for the volume coefficients

PR #1282, opened 2026-08-25
Simplify GASP mass solver loop

jkirk5:
### Summary

Broke up the GASP mass solver loop for fuel calculations by separating out the calculations strictly required for the feedback loop from other inputs/outputs that are not involved. Necessary for future multi-fuel work through isolating the tank sizing calculations which did not require the solver. I will update the tests to separately check the split components instead of together as a group once tank capacity is moved out of the mass group.

The tank capacity equations and the wing_fuel_min calculation (which is part of the feedback loop) share intermediate calculations - these were moved into a function so their duplicated code can be kept identical. I didn't want to make this a new component because the values being calculated are not useful on their own and shouldn't get promoted up.

### AI Usage

Claude Opus 4.8 was used to analyze the solver group and isolate which calculations could be separated out into separate components.


Kenneth-T-Moore:
It looks like one of the .png files is sitting in the top aviary directory. Was this intentional?

jkirk5:
> It looks like one of the .png files is sitting in the top aviary directory. Was this intentional?

No, it looks like the image got copied instead of moved after I generated it in the top directory, I've removed it now!

Kenneth-T-Moore (approved):

Issue #1251, opened 2026-08-04
Remove `get_option_defaults`


jkirk5:
### Desired capability or behavior.

The utility `get_option_defaults()` (in `aviary.variable_info.options.py`) which sets up an AviaryValues object with every variable set to their default units. This is often used to set up tests. This util makes the tests extremely annoying to debug because it makes it very opaque which inputs are being set that the component actually uses, and what value they are set to. The defaults are mostly zero anyway, and also add an unnecessary dependency for tests on default values in metadata. If a default is change, it shouldn't break every single test using that variable.

TLDR: `get_option_defaults()` is a shortcut that obscures 1. what variables are actually necessary for a test, 2. their values, and 3. creates fragile dependencies for the tests on the current metadata defaults

### Suggested implementation.

Update tests to not use `get_option_defaults()` and instead directly set input values for the OM problem. If there are no special cases where `get_option_defaults()` is absolutely essential, it should be completely removed. I can't think of any good reasons why this should exist, as it is a dangerous tool that looks like a handy shortcut but can be a little bit of a trap for the aforementioned reasons.

jkirk5:
This issue can probably be addressed in chunks, updating a set of components at a time

PR #1283, opened 2026-08-26
Ia1251 remove get option defaults

hschilling:
### Summary

Replace `get_option_defaults()` in unit tests with minimal, explicit `AviaryValues()` seeded only with the options each component under test actually declares via `add_aviary_option`. This makes each test's option dependencies visible at the call site and removes hidden coupling to the global metadata defaults.

Not all tests were migrated to the new approach. This is just a large number of them. More will be done in a future PR.

**Why**
`get_option_defaults()` returns every option in `_MetaData`. Tests using it fail silently when defaults drift, and it's impossible to tell from the test which options actually matter.

**Approach**
For each migrated test:

- Identify the subsystem tree under test (via graphify).
- Enumerate `add_aviary_option` declarations across that tree.
- Build a minimal `AviaryValues()` with only options the test needs to override.

### Related Issues

- #1251 

None

### AI Usage

AI was used in part for all the changes in this PR.


aviary/subsystems/aerodynamics/gasp_based/test/test_gaspaero.py
```diff
@@ -724,8 +723,7 @@ def test_case1(self):
         assert_near_equal(prob['SA7'], [0.03978045, 0.03978045], tol)
 
     def test_case_multiengine(self):
-        # 3-engine test case. 2nd and 3rd engine's properties are arbitrary
```
Kenneth-T-Moore, line 727: might want to keep this comment

aviary/subsystems/geometry/flops_based/test/test_prep_geom.py
```diff
@@ -799,7 +799,14 @@ class BWBSimplePrepGeomTest(unittest.TestCase):
     """
 
     def setUp(self):
-        options = self.options = get_option_defaults()
+        # PrepGeom under BWB + simple layout + detailed-wing-computed branch instantiates:
+        # BWBSimpleCabinLayout, BWBComputeDetailedWingDist, BWBFuselagePrelim, BWBWingPrelim,
```
Kenneth-T-Moore: full component list might be a little verbose for these

aviary/subsystems/geometry/gasp_based/test/test_wing.py
```diff
@@ -432,61 +453,11 @@ def test_case1(self):
         assert_check_partials(partial_data, atol=2e-12, rtol=1e-12)
 
 
-class WingFoldVolumeTestCase1(unittest.TestCase):
```
jkirk5, line 435: Why was this test removed?

Kenneth-T-Moore: > Why was this test removed?

I looked into this, and this test class appears twice, probably a cut and paste error.

hschilling: @Kenneth-T-Moore  Correct!

aviary/subsystems/aerodynamics/gasp_based/test/test_gaspaero.py
hschilling: done

aviary/subsystems/geometry/flops_based/test/test_prep_geom.py
hschilling: Removed

jkirk5 (approved):

PR #1284, opened 2026-08-27
Unify subsystem addition and solver control across all of our ODEs.

### LINKED ISSUE #1198: 2DOF ODE's do not use solver group flag
### Description

Various ODE's in the 2DOF mission have solver groups for things such as thrust/drag balance, but they do not use the "solver group" flag to determine if a subsystem should belong in the group. This means that external subsystems that ask to be added to the group might not be.

### Example

Depending on how the existing solvers are set up in the ODEs we may want to either have a second solver group for those subsystems, or we add them to the existing required solver groups.

It is ok for the 2DOF ODEs to force certain subsystem types (such as PropulsionBuilders) to be added to a specific solver group, even if that subsystem ends up being a user-provided subsystem. In general, it is better to add a system to the solver group when that wasn't strictly necessary then not adding a system to the solver group that needed to be there.

### Error Message

_No response_

### Aviary Version

1.0.1-dev

### Environment Information

_No response_

Kenneth-T-Moore:
### Summary

This PR unifies subsystem addition and solver control across all of Aviary's ODEs. Previously, only the Energy method allowed the subsystem to specify that it needed to be in a solver loop.

1. Renamed the ODE base class method `add_subsystems` to `add_subsystems_and_solver`. This method is now also responsible for adding all subsystems to either the top ODE or a solver sub group.
2. Each ODE is aware of the kind of coupling is implemented, so add_subsystems_and_solver now has arguments to control whether the propulsion (throttle) or aerodynamics (angle of attack) are coupled to the EOM. 
3. In the subsystem base class, `needs_mission_solver` now returns False. This should be set to True by the user if any additional subsystem coupling is introduced.
4. The solved-2DOF ODE has two separate solver subsystems, so `add_subsystems_and_solver` supports passing in two premade subsystems, and it assigns them to the appropriate solver group.
5. The setting of predetermined aero subsystem_options in the 2dof phases has been removed. Those options have been moved into the phase_info files instead.
6. Tightened up some solver tolerances, which seems to help some cases that had trouble hitting optimality.
7. Similarly, fixed a few spots where we were performing complex steps with a large stepsize (1e-8 vs 1e-40). This should also help with optimality.
8. Removed the `flight_path_ode`, which seemed to be an early unfinished experiment in unifying the 2dof ODEs.

Note: the solver setup is still not perfect. OpenMDAO 4 promises a complete decoupling of solver specification from the group, and this will streamline most of the difficulties.

### Related Issues

- Resolves #1198 

None

### AI Usage


aviary/mission/base_ode.py
```diff
@@ -52,31 +52,40 @@ def add_atmosphere(self, **kwargs):
...
         ----------
-        solver_group : om.Group
-            If not None, subsystems that require a solver (subsystem.needs_mission_solver() == True)
-            are placed inside solver_group.
-
-            If None, all subsystems are added to BaseODE regardless of if they request a solver.
-            TODO add solver compatibility to all ODEs
-
+        solver_sub: None or om.Group
+            Pre-created subsystem for the solver.
```
jkirk5: The word subsystem is confusing in this context, because I know you are talking about OM Systems but it sounds like it might be for Aviary subsystems

aviary/mission/base_ode.py
```diff
@@ -52,31 +52,40 @@ def add_atmosphere(self, **kwargs):
...
+        couple_propulsion : bool
+            When True, the ODE couples with any propulsion subsystems via a throttle to commanded
+            thrust balance.
+        couple_aero : bool
+            When True, the ODE couples with any aerodynamics subsystems via a force balance.
+        aero_solver_sub : None or om.Group
```
jkirk5, line 73: This argument seems hyper-specialized for a specific ODE, which I think disqualifies it as useful to include in `BaseODE`. Is a better approach to have these ODEs have their own definitions for the method? I don't want to keep having to add to `BaseODE` as we add additional equations of motion that also have special rules for how they want things set up.

aviary/mission/base_ode.py
```diff
@@ -121,4 +157,4 @@ def add_subsystems(self, solver_group=None):
                     promotes_outputs=mission_out,
                 )
 
-        return use_mission_solver
+        return solver_sub if solver_sub else self
```
jkirk5, line 158: Is returning the group useful? I feel like it will cause confusion, because an easy mistaken interpretation of this function is that it returns the correctly configured group of subsystems and solvers they then need to add to the ODE.

If users need the group object for some reason it already got added to the ODE so they should probably use OM interface to access that info in their ODE.

jkirk5: My suggestion is simply no return statement at all here

aviary/mission/base_ode.py
```diff
@@ -52,31 +52,40 @@ def add_atmosphere(self, **kwargs):
...
         nn = self.options['num_nodes']
         aviary_options = self.options['aviary_options']
         all_subsystems = self.options['subsystems']
         all_subsystem_options = self.options['subsystem_options']
         user_options = self.options['user_options']
-        use_mission_solver = False
+
+        # Prevent circular import
+        from aviary.subsystems.propulsion.propulsion_builder import PropulsionBuilder
+        from aviary.subsystems.aerodynamics.aerodynamics_builder import AerodynamicsBuilder
```
jkirk5: I wonder what causes the circular import?? If it is due to imports required for type hints in some other file, I found a way around that (see `subsystem_builder.py` imports)

jkirk5 (commented):
This update is a good demonstration of how OM4 could be valuable for Aviary and a point in favor of updating

aviary/mission/base_ode.py
Kenneth-T-Moore: Looks like I'm not seeing any interaction now, so I think I can safely move these back to the top.

aviary/mission/base_ode.py
Kenneth-T-Moore: After we return it, some of the ODEs will add the ode or a balance comp to the solver group.  I could modify it back so that each ODE is responsible for creating the Group and passing it into the add_subsystems, which is what it was doing originally. It is just some duplicated code.

aviary/mission/base_ode.py
Kenneth-T-Moore: The solved 2dof is the only ODE that creates two different solver groups.

aviary/mission/base_ode.py
Kenneth-T-Moore: should be Group

aviary/mission/base_ode.py
jkirk5: Oh ok that makes sense! I guess it is better how you have it here

jkirk5 (approved):

PR #1287, opened 2026-08-28
Restore the connection between the first reserve phase and the final regular phase. Fix a logic hole in the initial conditions setting for controls in the energy method.

Kenneth-T-Moore:
### Summary

This fixes a couple of bugs that @cmbenne3  identified.

1. The first reserve phase was completely unlinked from the previous regular phase. While there is probably a usecase for this, most of the time we are going to want at least the mass to be linked. This fix restores connection of all variables, except those that are given a value on both sides.
2. There was a hole in the logic for figuring out the initial controls in height energy, which allowed it to try to set a None value. This hole has been covered, and a nominal value is used even if there are no guesses, bounds, or constraints on either end.

NOTE: This PR also includes #1286 , so that should be reviewed first.

### Related Issues

- Resolves #

None

### AI Usage

### DIFF
--- a/aviary/core/aviary_group.py
+++ b/aviary/core/aviary_group.py
@@ -1021,10 +1021,6 @@ class AviaryGroup(om.Group):
             phase_info2 = self.mission_info[phase2]['user_options']
             vars2 = link_vars_dict[phase2]
 
-            if self.reserve_phases and phase2 == self.reserve_phases[0]:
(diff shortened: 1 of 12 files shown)


cmbenne3 (approved):

jkirk5:
Waiting for #1286 which currently has merge queue issues

PR #1291, opened 2026-08-31
Fix preprocessors for detailed fuselage

cmbenne3:
### Summary

This PR fixes a couple of errors with detailed fuselage component.
The preprocessor logic for fuselages requires passenger numbers and these get_vals were omitted from previous PR.
Reorganize logic for setting of defaults in preprocessors.
Update the fuselage.py detailed component to accept user specifying zero seats abreast when there are zero passengers of that class.

### Related Issues

- Resolves #

None

### AI Usage


aviary/utils/preprocessors.py
```diff
@@ -161,235 +161,241 @@ def preprocess_fuselage_layout(aviary_options: AviaryValues, verbosity=None):
...
-                    num_seat_abreast_business = 5
-                    aviary_options.set_val(
-                        Aircraft.CrewPayload.Design.NUM_SEATS_ABREAST_BUSINESS, 5
+                # set defaults for seat width
+                seat_width_economy = 20
+                seat_width_business = 22
+                seat_width_first = 25
+
+                # set defaults for seat pitch
+                seat_pitch_economy = 34
```
Kenneth-T-Moore, line 297: I wonder if we should remove the defaults from the variable metadata for these variables that have AircraftType specific defaults.

aviary/utils/preprocessors.py
```diff
@@ -161,235 +161,241 @@ def preprocess_fuselage_layout(aviary_options: AviaryValues, verbosity=None):
...
                 )
-            else:
-                num_seat_abreast_economy = 0
-                aviary_options.set_val(Aircraft.CrewPayload.Design.NUM_SEATS_ABREAST_ECONOMY, 0)
-                if verbosity >= Verbosity.BRIEF:
-                    warnings.warn(
-                        'Aircraft.CrewPayload.Design.NUM_SEATS_ABREAST_ECONOMY not '
-                        'set, assuming default 0.'
+                if num_economy_class_pax > 0 and num_seat_abreast_economy <= 0:
+                    raise ValueError(
```
Kenneth-T-Moore, line 195: There is a software engineering practice where we want to have test coverage for every line of code. To cover these, we would need a test that intentionally loads a negative value for each of these, and then verifies that the error is raised with the correct expected text. We haven't been adhering to this so far in early Aviary development, but it would be a good practice to adopt now that we are more mature.

cmbenne3: In general I don't think we have a test for preprocessors at all, so maybe we need to create one, unfortunately it's going to be MASSIVE!!!! What's the best practice? Should there be a separate unit test for every logic branch in preprocessors? Perhaps we can get an AI to help write this!

Kenneth-T-Moore: This does sound like an ideal low-impact task for an LLM. When I make these, I tend to put a bunch of checks in a single test. I suspect an LLM will make a separate test for each message, and that might be closer to "textbook" unit-testing.

aviary/utils/preprocessors.py
jkirk5: I think that is the right play. We want the defaults to actually get used, otherwise they are misleading and can confuse users, like "why is my seat pitch 24 inches when the default is 20, etc."

The metadata description should probably then mention the default is set based on aircraft type and differs between mass methods

aviary/utils/preprocessors.py
jkirk5: I will add this to the collection of issues around refactoring the preprocessors!
(Done, #1297)

jkirk5:
I am making some tweaks to this PR and double checking with FLOPS & GASP source code

jkirk5 (approved):

PR #1295, opened 2026-09-03
Make Aircraft.Nacelle.TOTAL_WETTED_AREA a scalar

### LINKED ISSUE #1208: Aircraft.Nacelle.TOTAL_WETTED_AREA has multivalue=True in the metadata.
### Description

Aircraft.Nacelle.TOTAL_WETTED_AREA shouldn't be multivalue because it is the sum of all nacelles. So far this bug hasn't affected us because it usually isn't in an input file. Setting it to False breaks the propulsion preprocessor in many of our tests, so this needs some careful work.

### Example

_No response_

### Error Message

_No response_

### Aviary Version

1.0.1-dev

### Environment Information

_No response_

xjjiang:
### Summary

This PR fixed the bug reported in #1208. In `preprocess_propulsion()`, all the `Aircraft.Nacelle` variables were assumed to allow multi-value. But `Aircraft.Nacelle.TOTAL_WETTED_AREA` is an exception. See `_get_engine_variables()` for details. The fix is to exclude it.

Two other scalar nacelle variables are `Aircraft.Nacelle.DRAG_FACTOR` and `Aircraft.Nacelle.PYLON_DRAG_FACTOR`. They are not excluded. In fact, they should be multi-valued. This will be taken care in a separate PR.

### Related Issues

- Resolves #1208 

None

### AI Usage

None.


aviary/utils/preprocessors.py
```diff
@@ -1199,5 +1199,11 @@ def _get_engine_variables():
     for item in get_names_from_hierarchy(Aircraft.Engine):
         yield item
 
+    excluded_names = [
+        Aircraft.Nacelle.TOTAL_WETTED_AREA,
+        Aircraft.Nacelle.DRAG_FACTOR,
+        Aircraft.Nacelle.PYLON_DRAG_FACTOR,
```
jkirk5: These variables shouldn't get excluded, it is reasonable for each set of engines to have different drag calibration factors

jkirk5 (changes requested):

aviary/utils/preprocessors.py
xjjiang: I understand but the current implementation assume that they are not vectors. Additional work must be done for them to be vectors.

xjjiang: I removed them from `excluded_names`. I will deal with them in a separate PR.

Kenneth-T-Moore (approved):

jkirk5 (approved):

PR #1296, opened 2026-09-04
Fix high linear fuel flow scaling in models and tests

jkirk5:
### Summary
There is an edge case where if fuel flow linear term is 1, and engine scale factor is 2, then fuel flow scaling as a whole becomes 0. That term was really only intended for very small values, well below 1. When the linear term is large, the intent of the scaling equation breaks down, and the optimizer will abuse this by making the engine as large as possible, as it sees an absolutely massive fuel burn reduction by doing so.

- Existing models were modified to use 0.0 instead of 1.0 for `Aircraft.Engine.FUEL_FLOW_SCALER_LINEAR_TERM` (I think it was mistakenly thought that the default for the linear term was 1, but is actually 0)
- Existing tests were modified to use a small value for linear scaler instead of large ones (for example 0.05 instead of 1.05)

Anyone using the affected models and also changing engine scale factor will see slight changes in results.


aviary/validation_cases/benchmark_tests/test_bench_throttle_allocation.py
jkirk5, line 1: This test sees large changes because the fuel flow was _significantly_ affected by engine scale factor due to the high linear scaling term. With a more reasonable, small, linear factor, the balance between engines changes.

aviary/validation_cases/benchmark_tests/test_bench_throttle_allocation.py
```diff
@@ -166,17 +168,19 @@ def test_multiengine_dynamic(self):
...
         with self.subTest('cruise_allocation'):
             # Cruise is pretty constant, check exact value.
-            assert_near_equal(alloc_cruise[0], 0.6452, tolerance=1e-2)
+            assert_near_equal(alloc_cruise[0], 0.33626162, tolerance=1e-2)
 
         with self.subTest('climb_allocation'):
-            # Check general trend: favors engine 1.
-            self.assertGreater(alloc_climb[2], 0.55)
+            # Check general trend: favors engine 2.
+            self.assertGreater(alloc_climb[2], 0.21)
```
jkirk5:
```suggestion
            self.assertLesser(alloc_climb[2], 0.22)
```

cmbenne3 (commented):
I think we should add validation to prevent this linear_fuel_term from being >=1 or <0 as it really doesn't make any sense.

We should probably also warn the user if they attempt to set something bigger than a certain threshold (e.g. 0.2).
In general a warning if they try and scale their engine more than 50% would probably also be a good idea - you probably need a new engine deck at that point, but that's unrelated to this PR.

At the very least we should explain somewhere (perhaps the metadata) what it represents. This is my explanation:
The idea of this linear term is it gives you an efficiency benefit for a larger engine and a penalty for a smaller engine. The size of the term represents the sensitivity of fuel burn change to the size of the engine.
If you scale the engine up or down 10%, a value of 0.1 would represent a 1% change in fuel burn (1% reduction for larger engine 1% penalty for smaller engine. Larger engines are more efficient than smaller ones due to blade-tip clearances, surface area to volume ratios etc).
A value of 1.0 means that for every 1% you scale the engine down or up, the fuel flow increases or decreases by 1%. The math breaks when you scale the engine up by 100% (SCALE_FACTOR=2), because the equation calculates a 100% fuel discount: 1.0 * (1 - 2.0) = -1.0. The engine would produce double the thrust but burn zero fuel. (This is what @jkirk5 explains in the PR description).
If you scale even more than 2 then the model could actually generate fuel!
If you set a linear term less than zero you reverse the physical effect this term is trying to model.
I don't think we should let users specify values that could lead to fuel generation, so I think we should add validation 0 <=linear term <1 somewhere to prevent this!

My suggestion for the metadata description:
 desc='Linear term in fuel flow scaling equation. Models efficiency changes due to engine sizing. Value of 0.1 increases/decreases fuel flow by 1% for every 10% engine is scaled down/up. (Note that the slope of fuel flow with respect to this variable changes from positive to negative as Aircraft.Engine.SCALE_FACTOR crosses from below to above 1).',
 
 Personally I find the 'Note' on the slope difficult to understand, but it is the mathematical explanation of what's happening.
 
 If we don't want to update the metadata then we could add this to the docs somewhere (not sure where the best place for that would be).

jkirk5:
> I think we should add validation to prevent this linear_fuel_term from being >=1 or <0 as it really doesn't make any sense.
See issue #1300 for a potential solution that is generalizable for any variable

PR #1125, opened 2026-04-29
Added a migration guide for 1.0.

Kenneth-T-Moore:
### Summary

Summary of PR.

### Related Issues

- Resolves #

### Backwards incompatibilities

None

### New Dependencies

None

PR #1299, opened 2026-09-11
Deprecation warnings for all variables since v.1.0

### LINKED ISSUE #1289: Add deprecation warnings for all variable hierarchy name changes since 1.0.
### Description

Name changes can catch people by surprise. We should enforce some kind of warning behavior to help them migrate their csv file or components to address the changes.

There are different ways we can handle this:
- raise an error?
- raise a warning and automatically replace the old names with the correct ones?
- do we keep the old names in the hiearchy?
- how long until we remove the deprecation?

### Example

_No response_

### Error Message

_No response_

### Aviary Version

1.0.3-dev

### Environment Information

_No response_

xjjiang:
### Summary

This PR targeted Issue #1289. I am sure this is not the final version but maybe a good point to start.

### Related Issues

- Resolves #1289

None

### AI Usage

None


jkirk5:
I had some ideas on how to use the check on the original issue (#1289) - it is kind of similar to your idea but a little different

aviary/utils/process_input_decks.py
```diff
@@ -157,6 +159,17 @@ def parse_inputs(
 
     guess_names = list(initialization_guesses.keys())
 
+    deprecated_vars = []
+    renamed_vars = []
+    deprecated_variables_file = get_path('utils/deprecated_variables.csv')
+    with open(deprecated_variables_file, newline='') as f:
```
Kenneth-T-Moore, line 165: I am glad that you did this!

Might be easier if you just stored the deprecated var information in a dictionary keyed by the strings in the first column (old varnames). 

`
{ "aircraft:design:thrust_takeoff_per_eng" : (removed, '#1142','1.0.0'), ...
`

Then you don't have to read a file, and the logic for checking it becomes 

```
if var_name in deprecated_vars:
    raise error and give new name and info
else:
    raise warning that the name wasn't found.
```

Kenneth-T-Moore:
Here is a little migration guide I made that might help catch some more variables. I closed my PR since it is stale, and your solution is a better approach.

https://github.com/OpenMDAO/Aviary/pull/1125

PR #1302, opened 2026-09-15
Make the FLOPS BWB work with Tabular Cruise aero.

Kenneth-T-Moore:
### Summary

The FLOPS BWB wouldn't make it through setup due to a couple of bugs.
1. Fixed the aero builder to support BWB + tabular.
2. Fixed `check_and_preprocess_inputs` so that we don't create any GASP geometry if the core method is FLOPS and we are only using GASP because of tabular_cruise.

NOTE: We should ultimately eliminate the need to set the aero_method to GASP just to use tabular cruise.

### Related Issues

- Resolves #

None

### AI Usage


aviary/core/aviary_group.py
```diff
@@ -490,18 +490,22 @@ def check_and_preprocess_inputs(self, verbosity=None):
...
         # If all phases ask for tabular aero, we can skip pre-mission. Check phase_info
-        tabular = False
+        tabular = []
         for phase in self.mission_info:
             if phase not in ('pre_mission', 'post_mission'):
                 try:
                     if (
-                        'tabular'
+                        'tabular_cruise'
                         in self.mission_info[phase]['subsystem_options']['aerodynamics']['method']
```
jkirk5:
```suggestion
                        any(val in (`tabular`, `tabular_cruise`, `tabluar_low_speed`) for val in 
                        self.mission_info[phase]['subsystem_options']['aerodynamics']['method'])
```
This just shifted the bug to not work for GASP low-speed tabular or FLOPS tabular!

Need to check if any of (`tabular`, `tabular_cruise`, `tabluar_low_speed`) in the phase

aviary/core/aviary_group.py
```diff
@@ -490,18 +490,22 @@ def check_and_preprocess_inputs(self, verbosity=None):
         energy = CoreEnergyBuilder('energy', meta_data=self.meta_data)
 
         # If all phases ask for tabular aero, we can skip pre-mission. Check phase_info
-        tabular = False
+        tabular = []
```
jkirk5: Maybe call it `all_tabular` for clarity? I'm also not sure what the advantage of making it a list is, but it doesn't hurt anything as far as I know

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -671,9 +671,9 @@ def get_parameters(self, aviary_inputs=None, user_options=None, subsystem_option
             elif method == 'cruise':
                 all_vars = set(AERO_2DOF_INPUTS + AERO_CLEAN_2DOF_INPUTS)
             elif method == 'tabular_low_speed':
-                all_vars = AERO_2DOF_TABULAR_LS_INPUTS
+                all_vars = set(AERO_2DOF_TABULAR_LS_INPUTS)
```
jkirk5, line 674: We could probably make all of the input lists sets to begin with so we never forget to convert them in places like here

aviary/subsystems/aerodynamics/aerodynamics_builder.py
```diff
@@ -685,7 +685,7 @@ def get_parameters(self, aviary_inputs=None, user_options=None, subsystem_option
             except KeyError:
                 design_type = AircraftTypes.TRANSPORT
 
-            if design_type is AircraftTypes.BLENDED_WING_BODY:
+            if design_type is AircraftTypes.BLENDED_WING_BODY and method != 'tabular_cruise':
```
jkirk5: Should `tabular_low_speed` also be excluded from this logic path?

Kenneth-T-Moore: Thinking more about this, external should also be excluded, so maybe the right way is to add these vars for "cruise" and "low_speed".

aviary/core/aviary_group.py
Kenneth-T-Moore: I thought we just renamed "tabular", but you're right, that is the older FLOPS tables.

Kenneth-T-Moore: I've reverted this change, because it already catches all 3 cases.

aviary/core/aviary_group.py
Kenneth-T-Moore: The previous code didn't stop iterating when it hit False, so it really was returning whether the final phase had tabular. I can rework this to make it a bool instead of a list though.

Kenneth-T-Moore:
Did a rewrite that addresses all except for the use of sets.

jkirk5 (approved):

PR #1304, opened 2026-09-17
Add multiengine capability to 2DOF missions

jkirk5:
### Summary

Added multiengine capability to 2dof mission analysis. The throttle allocator was added for phases that actually used throttle as a control (basically just cruise), while the other phases had throttle shape set to be able to accept multiple throttles per node. Nothing was changed about how throttle was used for each phase, so they are still mostly user-set static values per-phase.

I think there is potential to consolidate the existing `add_throttle_control` in 2dof ODEs with the updated `add_subsystems_and_solver` present in all ODES (from BaseODE class).

### Related Issues

- Resolves #

None

### AI Usage

AI was used to:
- debug some issues with why throttle indices were mismatched, helping to trace where their values ultimately get set from for the different 2dof phases.
- Re-work function on line 105 in mission/utils.py to support 2d throttle vectors
- Re-work epsilon function in compute of `PercentNotInFuselage` component of geometry/gasp_based/engine.py to support multiengine


aviary/mission/two_dof/ode/two_dof_ode.py
```diff
@@ -168,85 +184,106 @@ def add_alpha_control(
...
+
+            self.set_input_defaults(
+                Dynamic.Vehicle.Propulsion.THROTTLE, val=np.ones(nn), units='unitless'
+            )
+
+        if use_mission_solver or throttle_enforcement != 'control':
+            # TODO tie to verbosity
+            print_level = 2
+
+            propulsion_group.linear_solver = om.DirectSolver()
```
Kenneth-T-Moore, line 278: I think these 2 lines are superseded by the line at the bottom.

jkirk5:
The 2DOF throttle allocation test always returns a 50/50 split between the engines which definitely seems wrong, and I will need to dig in a little more to verify it is actually working

Kenneth-T-Moore (approved):

